PR #74, opened 2011-08-16
Added Support for JRuby-HTTPClient

spiegela:
I've added an adapter for the JRuby-HTTPClient.  This client uses the Apache HTTPClient Java lib.

The only test it was failing was for the HTTP PATCH command which isn't supported in the Java library (I've disabled it for this adapter.)

JRuby-HTTPClient doesn't yet support http-streaming or multi-part requests, but both are capable if someone wants to do it.

Thanks!

-Aaron


iMacTia:
Closing as @technoweenie said. We can always reopen it :)

PR #102, opened 2011-12-29
Automatic parallel manager detection

jimeh:
This enables the use of `#in_parallel` without manually passing in a parallel manager instance.

If the adapter you're using has a parallel manager, it will work as expected. If it doesn't, it'll work as normal sequential requests.

This is a bit of a dirty trick to figure out which middleware being loaded is an adapter by check if they respond to `#setup_parallel_manager`.

**Update:** The dirty hack in Builder is gone, and a custom parallel manager can either be passed in through the `:parallel_manager` option when creating the connection object, or directly to `#in_parallel`. If neither of those are done, a parallel manager will be created by calling `#setup_parallel_manager` without any arguments on the Adapter that's being used.

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -30,7 +30,7 @@ module Faraday
       @params                   = Utils::ParamsHash.new
       @options                  = options[:request] || {}
       @ssl                      = options[:ssl]     || {}
-      @default_parallel_manager = options[:parallel]
+      @default_parallel_manager = options[:parallel_manager]
 
       proxy(options.fetch(:proxy) { ENV['http_proxy'] })
 


bogdanRada:
I am not sure where to ask this, but i can't seem to use faraday with em_http adapter. i keep getting errors like <NoMethodError: undefined method `add' for true:TrueClass>  in gems/faraday-0.9.2/lib/faraday/adapter/em_http.rb:97:in `perform_request'  and another error  NoMethodError: undefined method `finish' for nil:NilClass  gems/faraday-0.9.2/lib/faraday/adapter/em_http.rb:118:in `block in perform_request' because it seems option **parallel_manager** is either sometimes NIL or TRUE , instead of being a instance of Faraday::Adapter::EMHttp::Manager. 


I am currently using this:

```ruby
Asana::Client.new do |c|
        c.authentication :access_token, @default_options.fetch(:asana_api_key, nil)
        c.debug_mode
        c.faraday_adapter :em_http
        c.configure_faraday do |conn|
          conn.use Faraday::Adapter::EMHttp
          conn.request  :url_encoded         
          conn.response :logger
          conn.default_parallel_manager = Faraday::Adapter::EMHttp.setup_parallel_manager
        end
      end
```

Maybe i am doing something wrong? Even if i remove the last line from the **configure_faraday** i get same errors. Could you help me please?

PR #170, opened 2012-06-19
Add ability to query builder/connection to see if they contain an adapter. 

avdi:
I want this in Leadlight so I can optionally add a default adapter if there isn't one already.


iMacTia:
Hello everyone,

I can see this PR have been open for quite a long time.
Code would need changes to be compatible with latest version, but now that we're approaching v1.0 (#620) we already decided to address this issue properly.

I'm sorry to close this after having it open for so long, all this discussion will be useful to decide how to address this issue the right way in v1.0

PR #200, opened 2012-10-15
insert/insert_after optionally takes an array for index

nearapogee:
This modifies builder's insert and insert_after to allow you to pass an array as the index. This allows you to insert before or after an array of other middlewares. See the tests in test/middleware_stack_test.rb for its exact behavior.

I made this commit awhile ago, but it should merge fine into master without any conflicts. Even still I can update it if necessary.

Please let me know what you think and I am certainly open to feedback. Thanks!

### DIFF
--- a/lib/faraday/builder.rb
+++ b/lib/faraday/builder.rb
@@ -112,7 +112,7 @@ module Faraday
 
     def insert(index, *args, &block)
       raise_if_locked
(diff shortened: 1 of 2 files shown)


mislav:
Sorry for the late reply, but we've decided to not add new methods to the middleware builder since we don't feel that they're needed right now.

PR #245, opened 2013-03-06
Escape sequence broken on Ruby 1.8.7 when operating in unicode mode

JoePym:
In Ruby 1.8, when you are operating with unicode encoding, \w includes UTF-8 characters such as ☃. Faraday therefore does not escape these correctly. This patch takes from the regular expressions used by @sferik in https://github.com/laserlemon/simple_oauth/blob/master/lib/simple_oauth/header.rb when deciding what to should be escaped.

I found this bug when I was attempting to post a tweet (using the Twitter gem) containing unicode in Ruby 1.8, and it was returning an "Authentication failed" message. Upon digging through the gems, simple_oauth was (correctly) escaping the special characters before signing the message, but Faraday was not escaping them when sending, causing the signing to mismatch.

I'm open to a better reg-exp way of addressing it. Manually spelling out the characters feels dirty to me, but I couldn't see a better way on a quick search.

### DIFF
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -1,6 +1,6 @@
 module Faraday
   module NestedParamsEncoder
-    ESCAPE_RE = /[^a-z0-9 \-\.\_\~\#\@\!\$\'\(\)\*\,\;\=]/i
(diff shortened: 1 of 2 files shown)


cyzgbw:
try https://github.com/qhwa/string_utf8

PR #247, opened 2013-03-13
Add support for no_proxy environment variable when using http_proxy

pixelblend:
Faraday currently supports the `http_proxy` environment variable and sets it as a default proxy if another is not explicitly set when a connection is created.

The `no_proxy` environment variable contains a comma-separated list of domains for which the proxy is not required[1]. If the requested URL matches an entry in this list, the default proxy should not be used.

This patch adds support for the `no_proxy` behaviour.

[1] This is already supported in a large number of existing ruby projects, e.g. [RubyGems](http://docs.rubygems.org/read/chapter/13#page51)

### DIFF
--- a/lib/faraday/autoload.rb
+++ b/lib/faraday/autoload.rb
@@ -73,7 +73,8 @@ module Faraday
       :Authorization => 'authorization',
       :BasicAuthentication => 'basic_authentication',
       :TokenAuthentication => 'token_authentication',
-      :Instrumentation => 'instrumentation'
+      :Instrumentation => 'instrumentation',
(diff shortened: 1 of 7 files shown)


brjennin:
Just wondering if there are any updates to this, seems like there hasn't been much chatter in the last year but we are in a scenario where we would like to use HTTP_PROXY for some requests and skip the proxy for others, preferably through the use of an ENV var

st33v:
Yep... I could use this right about now.

iMacTia:
Closing in favour of #361

PR #304, opened 2013-09-23
Raise Unauthorized error for 401 and Dont raise error for 422

prcongithub:
Raises Unauthorized error when HTTP 401 is received.
Don't raise error for 422. 422 can be easily handled by the parent application.


mislav:
Sorry, we have decided that testing for specific error codes is easy enough to do in code that we don't need special cases for 401 and 422.

PR #307, opened 2013-10-07
Document HTTPS usage

michaelklishin:
Shamelessly stoled from [this gist by Mislav](https://gist.github.com/mislav/938183) but this
stuff really should be documented.


aselder:
When documenting this, you may want to mention that different adapters need different values for some of the optios

:ssl[:client_cert] needs to be an OpenSSL::X509::Certificate object is using :net_http, but needs to be a file path is using :excon

mislav:
Closing as outdated. We still plan to add this information to the wiki, though!

Issue #322, opened 2013-11-21
Errno::ETIMEDOUT - Connection timed out - connect(2): faraday issue


AnanthSrinivasan:
I'm using Farady and it results in the below error any help would be appreciated. Couldn't find a reason why time out happens.

I am using the same code and faraday to connect to the service from my mac machine it works but on deploying to this linux ubuntu server and trying to get the response it fails with timeout. No clue whats going wrong here. 

```
D, [2013-11-21T08:53:06.220751 #7744] DEBUG -- request: User-Agent: "Faraday v0.8.8"
Content-Type: "text/xml"
Content-Encoding: "XML"
Errno::ETIMEDOUT - Connection timed out - connect(2):
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:878:in `initialize'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:878:in `open'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:878:in `block in connect'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/timeout.rb:66:in `timeout'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:877:in `connect'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:862:in `do_start'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:851:in `start'
    /home/amanoharan/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/net/http.rb:1367:in `request'
    /home/amanoharan/.rvm/gems/ruby-2.0.0-p247/gems/faraday-0.8.8/lib/faraday/adapter/net_http.rb:75:in `perform_request'
    /home/amanoharan/.rvm/gems/ruby-2.0.0-p247/gems/faraday-0.8.8/lib/faraday/adapter/net_http.rb:38:in `call'
    /home/amanoharan/.rvm/gems/ruby-2.0.0-p247/gems/faraday-0.8.8/lib/faraday/response.rb:8:in `call'
[...]

PR #324, opened 2013-12-04
wrapping Errno::ETIMEDOUT in a Faraday::Error::TimeoutError exception (#322)

ricardovaleriano:
Hey guys, thank you for you AWESOME work.

I don't know if the tests are shaped in the best form, if you want this patch, but want it with a different test suite, just let me know and I will try to put it together.

This addresses the issue #322 in respect to wrapping the `Errno::ETIMEDOUT` in a `Faraday::Error::TimeoutError` exception.


codesnik:
This is a really sad thread. No patch should be testes so rigorously that obvious very real problem isn't patched for a second year. If I'd have more time, I'd try to test it, but for now, I, as many others probably, going to write just

rescue Faraday::Error, Errno::ETIMEDOUT

test/adapters/integration.rb
```diff
@@ -178,6 +178,17 @@ def test_timeout
         end
       end
 
+      def test_timeout_before_open
+        valid_but_inexistent_host = 'http://10.255.255.254'
```
mislav: We can't affort to maintain external dependencies in our test suite.

The proper way would be to spin up our own test server that's programmed to misbehave, i.e. delay accepting a connection. I'm not sure how to pull that off in Ruby.

mislav:
@ricardovaleriano Thanks for your work on this. Sorry that it's so hard to make tests work consistently. It's one of the common problems in Faraday test suite because of so many adapters and Ruby versions that we need to support.

test/adapters/integration.rb
kainosnoema: Why not just stub `http.request` to raise `Errno::ETIMEDOUT`?

ricardovaleriano:
@mislav I totally understand your concerns man. And thank YOU for so much hard work on this. 
I'm not giving up yet, but I just don't know (yet) how to go further from here. =)

test/adapters/integration.rb
mislav: I guess we are forced to revert to stubbing for now. @ricardovaleriano can you try that approach?

ricardovaleriano: > I guess we are forced to revert to stubbing for now. @ricardovaleriano can you try that approach?

Yep, give a couple of days and I'll take a look.

PR #339, opened 2014-02-06
Adding support for newer SSL options in the net/http adapter

aetherknight:
In recent Rubies, net/http supports setting the list of ciphers, as well as an SSL timeout and a callback for additional server certificate verification. This PR adds support for those attributes to the net/http adapter.

Let me know if there are any changes you would like made. Thanks!

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -100,6 +100,9 @@ module Faraday
         http.ca_path      = ssl[:ca_path]      if ssl[:ca_path]
         http.verify_depth = ssl[:verify_depth] if ssl[:verify_depth]
         http.ssl_version  = ssl[:version]      if ssl[:version]
+        http.ciphers      = ssl[:ciphers]      if ssl[:ciphers]
+        http.verify_callback = ssl[:verify_callback] if ssl[:verify_callback]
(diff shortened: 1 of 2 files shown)


mislav:
Sorry for the late review. PR looks good, but can we provide the same options to other adapters as well? Generally, request options should work the same across all adapters.

aetherknight:
I don't think it is possible for every adapter to provide all of these parameters without significantly patching the underlying libraries. `net::http::persistent` is straightforward, but the others are not.

Furthermore, the existing adapters do not consistently implement the existing SSL options. For example, the net::http and net::http::persistent expect OpenSSL::X509::Certificate and OpenSSL::PKey::* objects for `:client_cert` and `:client_key`, but libcurl-based clients can only accept filenames for the certificate and private key (and the [file type required varies based on implementation of libcurl](http://curl.haxx.se/libcurl/c/CURLOPT_SSLCERT.html)), and their adapters treat `:client_cert` and `:client_key` as filenames.

Another example is that the excon adapter confusingly uses both `:client_cert` and `:client_key`, but also `:certificate` and `:private_key`.

aetherknight:
Closing this in favor of #544, which comes from my own fork of faraday instead of an old organization fork.

Issue #343, opened 2014-02-11
Connection#run_request with full URL ignores basic auth parameters


temochka:
Given that [#build_exclusive_url](https://github.com/lostisland/faraday/blob/master/test/connection_test.rb#L230) allows passing full URLs to override connection settings, it seems logical to assume that URLs with basic auth should be processed consistently. 

Following example demonstrates how two GET requests to the same URL result in different responses, depending on the way I set up the connection.

``` ruby
# GET http://httpbin.org/get => 200 OK
conn = Faraday::Connection.new
conn.get('http://httpbin.org/get')
#  => #<Faraday::Response:0x007fa069bcaa28 @on_complete_callbacks=[], @env=#<Faraday::Env @method=:get @body="{\n  \"args\": {},\n  \"url\": \"http://httpbin.org/get\",\n  \"origin\": \"68.82.9.238\",\n  \"headers\": {\n    \"User-Agent\": \"Faraday v0.9.0\",\n    \"Host\": \"httpbin.org\",\n    \"Connection\": \"close\",\n    \"Accept-Encoding\": \"gzip;q=1.0,deflate;q=0.6,identity;q=0.3\",\n    \"X-Request-Id\": \"9639fb50-b827-44bc-bfd9-566395b14488\",\n    \"Accept\": \"*/*\"\n  }\n}" @url=#<URI::HTTP:0x007fa069bc3840 URL:http://httpbin.org/get> @request=#<Faraday::RequestOptions (empty)> @request_headers={"User-Agent"=>"Faraday v0.9.0"} @ssl=#<Faraday::SSLOptions (empty)> @response_headers={"access-control-allow-origin"=>"*", "content-type"=>"application/json", "date"=>"Tue, 11 Feb 2014 18:19:51 GMT", "server"=>"gunicorn/0.17.4", "content-length"=>"334", "connection"=>"Close"} @status=200>> 

# GET http://user:password@httpbin.org/basic-auth/user/password => 401 Unauthorized
conn = Faraday::Connection.new
conn.get('http://user:password@httpbin.org/basic-auth/user/password')
[...]

PR #344, opened 2014-02-12
Do not reuse connection instance between requests with basic authentication parameters in URL.

temochka:
I think this should be the simplest solution for #343, considering this issue is more of an edge case than architectural problem. I wrote unit tests and added missing integration tests for described scenario.


mislav:
Thanks for suggesting the PR, and sorry it went ignored for so long. I'm closing this as stale since we decided we're not going to handle this case. See https://github.com/lostisland/faraday/issues/343#issuecomment-254279759

PR #347, opened 2014-02-17
EOF in case when we retry muiltipart request

isakovp:
IO rewind required if we do retry for multipart request


mislav:
Closing this as stale. Sorry we took such a long time to reach a decision. We're thankful for the contribution, but I think this is an edge case that is best left unhandled until there is a more elegant workaround.

PR #349, opened 2014-03-03
Make Faraday easily swappable with Net::HTTP

romanbsd:
There're several occasions that I needed to use Faraday instead of Net::HTTP, and while being mostly compatible, I had to monkey patch it for providing the "code" method.


mislav:
No, sorry. Your idea isn't bad, but this isn't something I want to start doing because someone else would ask to alias another method, and so on. For now, these Response types are simply incompatible.

PR #369, opened 2014-04-24
Add an optional "origin" option to authorization middleware.

knu:
This allows user to specify an "origin" URL only to which site the
authentication credentials given are passed via request header.

```
conn = Faraday.new { |faraday|
  faraday.request :url_encoded
  faraday.request :basic_auth, 'user', 'password1', 'http://example.org/private/'
  faraday.request :basic_auth, 'user', 'password2', 'http://example.net/'
}
```


mislav:
> I don't think it's desirable behavior to send out credentials for one site to other sites that require basic auth.

You're absolutely right, but I don't think this implementation is a good workaround. A better approach would be defining your own custom middleware (perhaps subclassing the Request::BasicAuthentication one) where you implement your own rules for sending passwords to various domains.

PR #376, opened 2014-05-05
Faraday::Utils::Headers#replace has side-effects

ketan:
The headers hash contains some local state inside the @names instance variable as a means of doing case-insensitive hashes. This has the side-effect that one has to remember to perform mirror operations on the @names hash in addition to the object itself(Headers).

A case in point is the replace method – which did not clear out the @names hash which was causing side-effects by not performing a mirror operation (clear) on the @names hash.

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -81,6 +81,7 @@ module Faraday
 
       def replace(other)
         clear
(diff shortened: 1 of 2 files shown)


mislav:
Thanks! Good catch!

PR #382, opened 2014-06-25
Add Response#redirect? to check for HTTP 3xx

nathanl:
Added only the sort of test that currently exists for `success?` in an effort to conform to the project's current approach. I could add a test where `redirect?` is true if you like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Faraday Changelog
 
+## Unreleased
+
+* Add `Response#redirect?` to check for HTTP status codes in 3xx (@nathanl)
+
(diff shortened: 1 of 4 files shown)


nathanl:
Rebased on current master

jirutka:
+1

mislav:
Hey, thanks for the solid contribution! The code looks great but we decided not to pull this in since we don't believe that `redirect?` method is needed right now. Checking for redirects is not as easy as just checking for a 3xx status; if you wanted to follow a redirect you would have to check for a specific 3xx status, the "location" header, the request method and so much more.

nathanl:
K, no worries. :)

PR #389, opened 2014-07-16
Add test coverage for POST with Retry middleware

Nitrodist:
There was a fix put in bc82276 (#296) that preserved the body of a
POST request that was being retried in the Retry middleware.
Unfortunately, there was no test added that ensures that this
functionality continues to work.

This commit adds the test to ensure that it continues to work in the
future.

### DIFF
--- a/test/middleware/retry_test.rb
+++ b/test/middleware/retry_test.rb
@@ -11,53 +11,53 @@ module Middleware
         b.request :retry, *retry_args
         b.adapter :test do |stub|
           ['get', 'post'].each do |method|
-            stub.send(method, '/unstable') {
+            stub.send(method, '/unstable') do |env|
               @times_called += 1
-              @explode.call @times_called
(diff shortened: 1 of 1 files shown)


kennethkalmer:
I had a quick look to help triage this, and it seems that rebasing this should not be a problem. I just wonder about changing `@explode` to `@response_handler` everywhere, it seems to only be a matter of taste. Since this PR I think one or two more tests got added, and they all stuck with `@explode`.

/cc @mislav @technoweenie @mike-bourgeous

mislav:
Yeah, let's avoid the renames to keep the diff minimal.

Nitrodist:
1 failure now for a failing test in master, it seems.

mislav:
Thanks! I've cleaned this up further.

Nitrodist:
:+1:

PR #390, opened 2014-07-16
Preserve and separate request and response bodies

### LINKED ISSUE #297: POST body is lost with Retry middleware
When retrying an HTTP POST the `env[:body]` of the request is overwritten by the failed response body.

This causes the retried operation to send a blank body.

This is one [workaround](https://github.com/librato/librato-metrics/blob/f18abf1f6e8a6a98f0908890308070e20d626431/lib/librato/metrics/middleware/retry.rb#L12-L18) for the issue.

Nitrodist:
The body attribute would be set with the body from the response once a
request was completed. The other attributes of the request (url,
headers, etc.) are not clobbered because they don't share the same keys.

In the case of the headers, the work had already been done to separate
them out into separate attributes (response_headers vs.
request_headers). This does similar work.

Fixes #297

Depends on #389 since this breaks untested functionality in the Retry middleware. Once that has been merged, then this pull request needs to be modified so that the small amount of work from https://github.com/Nitrodist/faraday/commit/3348e99f8c0e94881848879f3b7b0a3794bea14a goes into it as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ end
 ## GET ##
 
 response = conn.get '/nigiri/sake.json'     # GET http://sushi.com/nigiri/sake.json
-response.body
+response.response_body
 
 conn.get '/nigiri', { :name => 'Maguro' }   # GET http://sushi.com/nigiri?name=Maguro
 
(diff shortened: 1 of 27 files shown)


mislav:
I agree that this changes too much.

I get how a single `body` property can be ambiguous, especially as `env[:body]` where it holds request body during request phase and response body during respose. However, I don't think we need to go as far to redefine every `body` accessor in the codebase. I'm fine with API such as `req.body = 'REQUEST'` and `response.body #=> 'RESPONSE'`.

I would be more in favor of saving original request body in env so it isn't trumped over during response. Something like this was attempted in #517.

Right now there are no official plans for what goes in 1.0, but whenever there is a major release next, I don't want to change any more APIs needlessly because people already had a rought transition from 0.8 to 0.9 and I don't want to get them on a similar rollercoaster again.

PR #391, opened 2014-07-17
Update Utils::ParamsHash#update

DragosIorgulescu:
Allowing for nil when setting params

There are situations where it makes sense to set request params with an object which may be nil or a hash. This may be just a preference, since you can always make sure you are not passing nil, but this would require an extra check sometimes which can be skipped.

If you deem this as a good idea, let me know.

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -126,7 +126,7 @@ module Faraday
       alias_method :key?, :include?
 
       def update(params)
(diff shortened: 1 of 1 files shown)


mislav:
Sorry, in this case we want an exception to be thrown when passing an object that doesn't respond to `each`. You should type check outside of calling `update()`, thanks!

Issue #397, opened 2014-08-15
Instance without URL prefix


kenn:
Is there a way to create an instance with common settings without defining URL prefix? The documentation wasn't immediately clear on this matter.

Apple's iOS in-app purchase verification service [requires to try production server first, then fallback to sandbox if necessary](https://developer.apple.com/library/ios/technotes/tn2259/_index.html#//apple_ref/doc/uid/DTS40009578-CH1-FREQUENTLY_ASKED_QUESTIONS).

> How do I verify my auto-renewable subscriptions receipt (iOS)?
> Always verify your receipt for auto-renewable subscriptions first with the production URL; proceed to verify with the sandbox URL if you receive a 21007 status code. Following this approach ensures that you do not have to switch between URLs while your application is being tested or reviewed in the sandbox or is live in the App Store.

The following is a working example using Faraday:

```ruby
# Hit production first
response = Faraday.new(url: 'https://buy.itunes.apple.com/verifyReceipt').post do |req|
  req.headers['Content-Type'] = 'application/json'
  req.body = postdata
  req.options.timeout = 5
  req.options.open_timeout = 2
end
parsed_receipt = parse_json(response.body)

if parsed_receipt.status == 21007
  # Not a production receipt - Fallback to sandbox
  response = Faraday.new(url: 'https://sandbox.itunes.apple.com/verifyReceipt').post do |req|
    req.headers['Content-Type'] = 'application/json'
    req.body = postdata
    req.options.timeout = 5
    req.options.open_timeout = 2
  end
end
```

But there's a huge overlap for the exact same config.

It would be nice if I could do instead:

```ruby
faraday = Faraday.new do |req|
  req.headers['Content-Type'] = 'application/json'
  req.options.timeout = 5
  req.options.open_timeout = 2
end

response = faraday.post 'https://buy.itunes.apple.com/verifyReceipt', postdata
parsed_receipt = parse_json(response.body)
if parsed_receipt.status == 21007
  response = faraday.post 'https://sandbox.itunes.apple.com/verifyReceipt', postdata
end
```

Is this kind of use case supported with Faraday, or is there a better way?

mislav:
@kenn You forgot to specify the adapter in your `Faraday.new do ... end` block. That's why the request fails. Whenever you supply a block, you need to specify the adapter. Sorry for the late reply.

PR #401, opened 2014-08-19
 Instance without URL prefix #397 - solved

davidraj:
I have solved this by adding some methods

The options should be set as follows before hand

Faraday.set_options(:headers => {'Content-Type'=> 'application/json'},:request => {:timeout => 5,:open_timeout => 2})


mislav:
Sorry, I don't understand what this solves. It's entirely possible to create a Faraday instance without a specific host prefix. Please see https://github.com/lostisland/faraday/issues/397#issuecomment-56461029

PR #403, opened 2014-08-20
Added authentication to readme

Austio:
Added examples for adding token and basic authentication.  Added as new section because it is independent of type of request (POST, GET, etc)


mislav:
Hi thanks for the contribution, but we believe that this information is not well suited for the README. We'll try to add such usage documentation our wiki instead.

PR #405, opened 2014-09-01
Fix interval_randomness data type

dtaniwaki:
I found `interval_randomness` was decimal number but it's converted to integer in the getter method.

[Review on Reviewable](https://reviewable.io/reviews/lostisland/faraday/405)


mislav:
Good catch!

PR #409, opened 2014-09-07
Remove remnant `require_libs` call from `Faraday::Utils`

loopj:
Removes a call to `Faraday.require_libs` from `Faraday::Utils` which looks like it should have been removed in 1e5284c5adbe07561a44d2119e6415eb3390701f.

The `parameters.rb` file is already required in `lib/faraday.rb`.


mislav:
Good catch. Eventually I'd like to get rid of `require_libs` stuff in favor of plain requires.

PR #416, opened 2014-09-22
failing specs for nested encoding and decoding of arrays of hashes

danhunter:
Found that the NestedParamsEncoder wasn't decoding a non-indexed array of hashes the way I expected, and indeed the way that the NestedParamsEncoder encodes the same array. 

Is this expected behavior? Should an array index always be included? 

Pull request **only** includes **failing** specs.


mislav:
This is now handled by #531. Closing in favor of that

PR #425, opened 2014-10-21
Clarify block definition of connections

mltsy:
Just adding or removing an empty block from the call to `Faraday.new` changes the behavior.  This should be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,16 @@ Rack::Test, and a Test adapter for stubbing requests by hand.
 ## Usage
 
 ```ruby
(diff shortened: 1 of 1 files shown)


dasch:
:+1: that one surprised me.

iMacTia:
Hi @mltsy thank you for your PR and sorry for the late reply.
Indeed we noticed this behaviour is not clear and documentation is not helping at all.
However, we have also other things that we would like to add into the documentation, plus a general revamping of the Wiki in the next future.
Appreciating your input, I'm going to close this as it will be part of that revision.

PR #429, opened 2014-10-29
fix bug with request_headers being rewritten

olek:
At least for clients using OAuth, retrying fails with 400 or 401 errors because request headers are modified by 'failure response'.
This change fixes the problem and allows retries to actually work in those scenarios.

Bug  was first noticed about 2-3 years ago, but reason (and solution) for it surfaced only now.

### DIFF
--- a/lib/faraday/request/retry.rb
+++ b/lib/faraday/request/retry.rb
@@ -97,13 +97,15 @@ module Faraday
     def call(env)
       retries = @options.max
       request_body = env[:body]
+      request_headers = env[:request_headers]
       begin
-        env[:body] = request_body # after failure env[:body] is set to the response body
         @app.call(env)
       rescue @errmatch => exception
         if retries > 0 && retry_request?(env, exception)
           retries -= 1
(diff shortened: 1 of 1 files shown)


mislav:
Sorry for the late reply. I don't get this bug. The response process does overwrite `env[:body]`, but it never overwrites `env[:request_headers]`. Instead, it writes to `env[:response_headers]`.

Can you submit a test that demonstrates this bug? Or, does it happen in combination with some specific middleware? It might be that some middleware is behaving with regard to `env[:request_headers]`.

olek:
Sorry, but too much time have passed, my memory does not last that long :)
It might have been a weird interaction with middleware that allows for painless OAuth 1.0 - I think I have verified at that time that same test without OAuth middleware was working just fine, but as soon as OAuth middleware was introduced, retries would never succeed (until this patch).

mislav:
Is that middleware available somewhere for you to link us to? Maybe the bug is in the middleware itself.

mislav:
Closing as stale.

PR #435, opened 2014-11-12
Replace empty array indexes for Faraday::NestedParamsEncoder

rusikf:
@mislav , @sporkmonger , @danhunter, @sferik ,@technoweenie
More detail see [here](https://github.com/lostisland/faraday/pull/416)

### DIFF
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -63,6 +63,7 @@ module Faraday
 
     def self.decode(query)
       return nil if query == nil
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for the improvement! I went for fuller Rack::Utils (used by Rails) compatibility by rewriting that whole decoder method in https://github.com/lostisland/faraday/pull/531, please take a look.

It handles the case when there are multiple array items generated by nested hash keys.

PR #437, opened 2014-11-14
Allow specifying methods that bypass :retry_if on Request::Retry

mike-bourgeous:
This expands the `IDEMPOTENT_METHODS` functionality to allow speifying the
methods that will be retried without calling `:retry_if`.  The new `:methods` option
to `Request::Retry` allows `:retry_if` to be called for every HTTP method, a user-
specified set of HTTP methods, or never.

It's likely that `:retry_if`, `IDEMPOTENT_METHODS`, and `:methods` could be
combined into a simpler API (such as by changing the `:retry_if` default value to
check for an idempotent method), but this separate approach preserves the
existing behavior if `:methods` is not specified.

Tests and documentation are included in the pull request; I ran the tests against
Ruby 2.1.3 and Ruby 1.8.7.

### DIFF
--- a/lib/faraday/request/retry.rb
+++ b/lib/faraday/request/retry.rb
@@ -22,7 +22,8 @@ module Faraday
 
     IDEMPOTENT_METHODS = [:delete, :get, :head, :options, :put]
 
-    class Options < Faraday::Options.new(:max, :interval, :interval_randomness, :backoff_factor, :exceptions, :retry_if)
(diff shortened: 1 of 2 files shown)


technoweenie:
Thanks for the patch!  Just FYI, I'm replacing Faraday with a new lib: Hurley.  https://github.com/lostisland/faraday/issues/454

Issue #444, opened 2014-12-18
Issue with a string keys in Faraday::Connection options


aliaksandrb:
Hi!
Recently faced a bug which reproduces by the gem that uses Faraday as dependency ([elasticsearch-ruby](https://github.com/elasticsearch/elasticsearch-ruby) to be more precise).
Simplified situation and steps to reproduce:
According to docs `Faraday::Connection` [allows](https://github.com/lostisland/faraday/blob/master/lib/faraday/connection.rb#L46) passing the `options` argument that should be "Hash or `Faraday::ConnectionOptions`"
Lets check this:

```
conn = Faraday::Connection.new 'http://sushi.com', { :request => { :timeout => 5 } }
# => Faraday::Connection
conn.get 'nigiri'
# => Faraday::Response
conn.options
# => <Faraday::RequestOptions timeout=5>
```

Everything fine!
But what if the keys in options hash are strings?

```
conn2 = Faraday::Connection.new 'http://sushi.com', { "request" => { "timeout" => 5} }
# => Faraday::Connection
conn2.get 'nigiri'
# => NoMethodError: undefined method `params_encoder' for {"timeout"=>5, :proxy=>nil}:Hash
    from ~/.rvm/gems/ruby-2.1.5@yocto/gems/faraday-0.9.0/lib/faraday/request.rb:58:in `url'
    from ~/.rvm/gems/ruby-2.1.5@yocto/gems/faraday-0.9.0/lib/faraday/connection.rb:371:in `block in run_request'
    from ~/.rvm/gems/ruby-2.1.5@yocto/gems/faraday-0.9.0/lib/faraday/connection.rb:388:in `block in build_request'
....
conn.options
# => {"timeout"=>5}
```

Fail! And Elasticsearch just can't connect to it's server and nothing is working :(

I believe that in [faraday/options.rb](https://github.com/lostisland/faraday/blob/d8b9eda2d2d156c8e088337900636aa1b319e8bc/lib/faraday/options.rb#L19) suboptions defined as symbols:

```
...
options :request => RequestOptions, :ssl => SSLOptions
...
```

so `#update` method just could not find it:

``` ruby
    def update(obj)
      obj.each do |key, value|
    ******if sub_options = self.class.options_for(key)******
          value = sub_options.from(value) if value
        elsif Hash === value
          hash = {}
          value.each do |hash_key, hash_value|
            hash[hash_key] = hash_value
          end
          value = hash
        end

[...]

PR #445, opened 2014-12-18
Fix for issue with string keys in connection options

aliaksandrb:
Hi!
Here is small fix for *Issue with a string keys in Faraday::Connection options #444*


mislav:
Per https://github.com/lostisland/faraday/issues/444#issuecomment-81460929, we won't support string keys on principle. I'll try to make the error message more helpful, though, to resolve the original issue you've opened.

Issue #446, opened 2014-12-24
Persistent http adapter.


Ben-M:
Would you be open to a pull request for a [persistent_http](https://github.com/bpardee/persistent_http) adapter?

Persistent_http is very much like Net::HTTP::Persistent, but provides a connection pool so that connections can be shared between threads. This should provide better results for threaded web servers.

PR #452, opened 2015-01-04
NetHttp uses with_net_http_connection (#446)

Ben-M:
As discussed in #446 this switches Net:Http to use 'with_net_http_connection' internally, so that it can be overridden to allow for connection pooling.

Let me know if you'd prefer me to remove the version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Faraday Changelog
 
+## v0.9.1
+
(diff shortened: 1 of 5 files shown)


Ben-M:
@technoweenie would you prefer this with or without the version bump?

technoweenie:
Typically I do version bumps separately.  The changelog edit is good.

Since there's not much Faraday action going on, I can merge this and push a new version out.

PR #457, opened 2015-01-14
httpclient: Enable gzip compression

apepper:
Motivation: `httpclient` support gzip compression since late 2010 (see https://github.com/nahi/httpclient/issues/42).

### DIFF
--- a/test/adapters/httpclient_test.rb
+++ b/test/adapters/httpclient_test.rb
@@ -17,5 +17,7 @@ module Adapters
         assert_equal host, conn.options[:bind][:host]
       end
     end
+
+    Integration.apply(self, :Compression)
   end
 end
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -59,7 +59,8 @@ module Adapters
     module Compression
       def test_GET_handles_compression
         res = get('echo_header', :name => 'accept-encoding')
-        assert_match(/gzip;.+\bdeflate\b/, res.body)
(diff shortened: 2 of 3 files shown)


technoweenie:
Seems legit, tests pass, etc.  Thanks!

test/adapters/httpclient_test.rb
```diff
@@ -17,5 +17,7 @@ def test_binds_local_socket
         assert_equal host, conn.options[:bind][:host]
       end
     end
+
+    Integration.apply(self, :Compression)
```
mislav: You could have added `:Compression` to the existing `Integration.apply` call above in the same file.

test/adapters/integration.rb
```diff
@@ -59,7 +59,8 @@ def test_no_parallel_support
     module Compression
       def test_GET_handles_compression
         res = get('echo_header', :name => 'accept-encoding')
-        assert_match(/gzip;.+\bdeflate\b/, res.body)
```
mislav: Why was this regex change necessary?

apepper: Since `httpclient` sends `Accept-Encoding: gzip,deflate` with a `,` instead of an `;`. This is valid as  described by http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3 .

test/adapters/httpclient_test.rb
apepper: Good point. I improved it in #459.

PR #459, opened 2015-01-16
[test] Simplified httpclient_test.

apepper:
This is a follow up of https://github.com/lostisland/faraday/pull/457/ .


mislav:
Thanks!

PR #374, opened 2014-05-04
Updated .travis.yml for jruby 2.0 and 2.1 modes

alhafoudh:
Included jruby 2.0 mode and 2.1 mode into build matrix

Issue #348, opened 2014-02-26
faraday 0.9.0 can't use in jruby mode 2.0


mimosa:
`gem install faraday` 

`jruby --2.0 -S pry`

``` ruby
require 'faraday'
conn = Faraday.new(url: 'http://open.shou65.com')
response = conn.get '/weather.json'
response.body # Nothing
```

replace adapter is work :100: 
`Faraday.default_adapter = :httpclient`

1.9 is ok :+1:

iMacTia:
Closing as this should now be handled in 280f08a3b282ef571b113b688623fef35fb65bc7.

PR #464, opened 2015-02-01
Test against Ruby 2.2 and JRuby 2.0/2.1 modes

O-I:
Hi.

This attempts to move issues #348 and #374 forward. I've added JRuby 2.0 and 2.1 modes to the Travis build and allow them to fail. I've also added MRI 2.2 to see if it passes.

Cheers,
O-I

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -74,8 +74,8 @@ module Faraday
       defined? RUBY_ENGINE and 'rbx' == RUBY_ENGINE
     end
 
-    def self.ruby_22?
-      defined? RUBY_ENGINE and 'ruby' == RUBY_ENGINE and /^2\.2\.\d+/ =~ RUBY_VERSION
+    def self.ruby_22_plus?
+      RUBY_VERSION > '2.2'
     end
 
     def self.ssl_mode?
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -74,6 +74,10 @@ module Faraday
       defined? RUBY_ENGINE and 'rbx' == RUBY_ENGINE
     end
 
+    def self.ruby_22?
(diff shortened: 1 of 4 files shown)


O-I:
Ruby 2.2 fails with this [error](https://travis-ci.org/lostisland/faraday/jobs/49046561). I've removed it from .travis.yml for now. Not too confident I can troubleshoot that alone. I can explore. Or someone with more experience can enlighten me at their convenience. :smiley:

mislav:
It seems that old Typheus fails to compile on Ruby 2.2. You could put back 2.2 on the CI list but figure out how to skip the old Typhoeus test. It's not crucial.

O-I:
OK. So I think I figured out how to skip the Typhoeus tests on Ruby 2.2 and get it passing on Travis, but now I've hit some sort of SSL(?) [error](https://travis-ci.org/lostisland/faraday/jobs/49234198) on Ruby 2.1.

test/helper.rb
```diff
@@ -74,6 +74,10 @@ def self.rbx?
       defined? RUBY_ENGINE and 'rbx' == RUBY_ENGINE
     end
 
+    def self.ruby_22?
+      defined? RUBY_ENGINE and 'ruby' == RUBY_ENGINE and /^2\.2\.\d+/ =~ RUBY_VERSION
```
mislav: You should do `RUBY_VERSION > '2.2'` and you don't need any RUBY_ENGINE checks. We want this method to return true for every version from 2.2.0 forward. We don't want Typhoeus tests run on 2.3.

mislav:
Intermittent failure. We sometimes get those. Restarted and the build passd.

O-I:
Wow. Plain old string comparison really makes that method clear. Anything else?

mislav:
Thanks!

PR #468, opened 2015-03-14
Make excon adapter compatible with 0.44 excon version

chipiga:
Now it always use 127.0.0.1 as proxy host


gustavokloh:
+1 here.

mislav:
Thanks! I guess there's no way to test this because our test proxy server is on 127.0.0.1.

PR #469, opened 2015-03-16
Added method sort_params= to Faraday::Connection

nicknux:
- The sort_params= method takes in a boolean which
  indicates whether to sort request parameters.
  By default Faraday sorts the paramerers.

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -349,7 +349,7 @@ module Faraday
 
       query_values = params.dup.merge_query(uri.query, options.params_encoder)
       query_values.update extra_params if extra_params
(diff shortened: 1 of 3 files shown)


mislav:
I'm against turning every little tweak that someone might need once for a specific server into a new Faraday configuration key.

To configure exactly how you need parameters to be encoded for your server, ordering and all, please use the custom `params_encoder` option:

```rb
Faraday.new(url, request: { params_encoder: MyEncoder })
```

The custom encoder object should respond to `encode(params)` and `decode(query)`. Please [see existing encoders](https://github.com/lostisland/faraday/blob/81f16593a0138ec58bb6f25e1c2804e91589662f/lib/faraday/parameters.rb) for inspiration.

PR #470, opened 2015-03-30

aalness:
Testing the actual functionality with net-http-persistent with the current LiveServer seems not possible. Reluctant to refactor that for testing this.

Manual testing with netcat confirmed correct behavior for different settings.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,11 @@ bundler_args: --without development --path vendor/bundle --jobs=3 --retry=3
 cache: bundler
 
 rvm:
-  - 1.8.7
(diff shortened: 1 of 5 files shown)


aalness:
Hmm, `Adapters::Patron#test_OPTIONS` seems to be failing in master.

hudon:
@technoweenie Would be great to get this merged, how does it look?

mislav:
Thanks for the contribution, but we can't accept features that work only with one adapter and not with others. The whole point of Faraday is that you can use the same code unchanged with one adatper and switch to another. Sorry for the late reply!

PR #473, opened 2015-04-04
Add Manticore adapter

cheald:
This adds an adapter for [Manticore](https://github.com/cheald/manticore), a high-performance JRuby-only HTTP client.

I can distribute the adapter with Manticore directly, but I figured I'd try this first, since this makes it much easier to make sure that the adapter doesn't break with Faraday changes; by testing it with the Faraday test suite, it'll be easier to make sure things work as intended in the future.

Parallel requests are supported and work well. Compression is supported. SSL is supported, but it takes different keys than Faraday's defaults, since Manticore can't currently read PEM certificates (it prefers JKS and PKCS12). Streaming reads are easily supported once Faraday's interface for streaming is implemented.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ group :test do
   gem 'simplecov'
   gem 'sinatra', '~> 1.3'
   gem 'typhoeus', '~> 0.3.3', :platforms => [:ruby_18, :ruby_19, :ruby_20, :ruby_21]
+  gem 'manticore', '~> 0.4', :platforms => [:jruby]
 end
 
 gemspec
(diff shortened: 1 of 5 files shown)


cheald:
The MRI suite is failing because of Patreon failures unrelated to Manticore. Jruby-1.8 mode fails due to the lack of `cgi/cookie`, which I'll investigate. The test suite in SSL mode is failing; I'll investigate and figure out what need to happen there.

cheald:
I'm actually going to abandon this in favor of just shipping the adapter in-gem.

Issue #441, opened 2014-12-04
It seems 0.9.0 does not set Faraday::Response object in `env` correctly


taiki45:
I found the regression or change which was brought by v0.9.0. This regression will be fixed by merging https://github.com/lostisland/faraday/pull/360.

---

I tested by this test code:

```
require File.expand_path('../helper', __FILE__)

class RackBuilderTest < Faraday::TestCase
  class TestMiddleware < Faraday::Middleware
    def call(env)
      @app.call(env).on_complete do |e|
        raise ResponseNotGiven unless e[:response]
      end
    end
  end

  class ResponseNotGiven < StandardError
  end

  def setup
    @conn = Faraday.new do |b|
      b.use TestMiddleware
      b.adapter :test do |stub|
        stub.get('ok') { [200, {'Content-Type' => 'text/html'}, '<body></body>'] }
      end
    end
  end

  def test_response_not_given
    assert_raises ResponseNotGiven do
      @conn.get('ok')
    end
  end

  def test_response_given
    assert_kind_of Faraday::Response, @conn.get('ok')
  end
end
```

v0.8.9 sets `Faraday::Response` in yielded `env` correctly.

```
~/.ghq/github.com/lostisland/faraday v0.9.0*
❯ git co v0.8.9
Note: checking out 'v0.8.9'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at 0534b2d... Release 0.8.9

~/.ghq/github.com/lostisland/faraday 0534b2d...*
❯ bundle exec ruby test/rack_builder.rb
Your Gemfile lists the gem rake (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Your Gemfile lists the gem simplecov (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Run options:

# Running tests:

[2/2] RackBuilderTest#test_response_not_given = 0.00 s
  1) Failure:
RackBuilderTest#test_response_not_given [test/rack_builder.rb:25]:
RackBuilderTest::ResponseNotGiven expected but nothing was raised.

[...]

PR #477, opened 2015-05-05
Env stored on middleware response has reference to the response

JoshCheek:
Synopsis
--------

The `Env` that gets stored on the `Response` does not contain a reference to that response. This env is a copy of the env that was passed through the middleware stack. Thus if the env passed to `app.call` is used, it has the response, but if the one from the `Response` is used, then it does not. This happens when, for example, you invoke the `on_complete` hook on the response, after it has finished.

Fixes Longstanding VCR issue
----------------------------

Fixes https://github.com/vcr/vcr/issues/386

A longstanding issue in VCR's integration with v0.9.x It was worked around in this commit: https://github.com/vcr/vcr/pull/439 And is merged into master, but is not released. Merging this will fix the issue, without the changes to VCR needing to be released.

The old code, which is currently released, and gets its env from the Request can be seen [here](https://github.com/vcr/vcr/blob/cb9696c44dee4c0d39b4f531e72516f4bc57272f%5E/lib/vcr/middleware/faraday.rb#L99-105)

Fixes Faraday Issue
-------------------

Fixes https://github.com/lostisland/faraday/issues/441
which identifies this same problem.

Closes Faraday PR
-----------------

I'm going to recommend closing https://github.com/lostisland/faraday/pull/360/files as the issue was probably introduced [here](https://github.com/lostisland/faraday/commit/04a8514cba58f899183dbe2d5c7c4281d1e85f3f#diff-fc9726b31994223d11a212b5169100fdL68) when `Response#finish` changed from storing the `env` directly, to storing a copy:

```diff
- @env = env
+ @env = Env.from(env)
```

Presumably this behaviour is still desired, but the mentioned pull essentially undoes this by returning the same `env` from `Env.from` ([link](https://github.com/lostisland/faraday/pull/360/files#diff-32dfef617839b0b1d5f792c3d3da8420R275))

The commit that introduced the issue is probably good, but is called by the app in the middleware stack that creates the response, before it sets the rsponse onto the env. This means that the copy does not have the link to the response.

https://github.com/lostisland/faraday/blob/81f16593a0138ec58bb6f25e1c2804e91589662f/lib/faraday/rack_builder.rb#L152-155

Not sure about the test name/location
-------------------------------------

I had difficulty figuring out where to put this test, and what to name it. I eventually went with the test on the `middleware_stack_test.rb`, because it deals with the `RackBuilder`, but if there's a better place,
or a better name, let me know and I'll update it.

One other weird thing
---------------------

Depending on when the `Request#on_complete` callback is invoked, you're either getting the original env, or the copy:

https://github.com/lostisland/faraday/blob/81f16593a0138ec58bb6f25e1c2804e91589662f/lib/faraday/response.rb#L53-67

### DIFF
--- a/lib/faraday/rack_builder.rb
+++ b/lib/faraday/rack_builder.rb
@@ -151,8 +151,9 @@ module Faraday
         lock!
         to_app(lambda { |env|
           response = Response.new
-          response.finish(env) unless env.parallel?
           env.response = response
+          response.finish(env) unless env.parallel?
+          response
         })
       end
     end
--- a/test/middleware_stack_test.rb
+++ b/test/middleware_stack_test.rb
@@ -143,6 +143,15 @@ class MiddlewareStackTest < Faraday::TestCase
     assert_match "zomg/i_dont/exist", err.message
   end
 
+  def test_env_stored_on_middleware_response_has_reference_to_the_response
+    env = response = nil
+    build_stack Struct.new(:app) {
+      define_method(:call) { |e| env, response = e, app.call(e) }
+    }
+    @conn.get("/")
+    assert_same env.response, response.env.response
+  end
+
   private
 
   # make a stack with test adapter that reflects the order of middleware


JoshCheek:
The CI failure is due to some other issue. I skipped that test in development, since it wasn't the bug I was trying to fix.

mislav:
Thanks for the investigative work, and sorry for the late reply! I'm going to test this out and merge if everything works out. I'd like to restore VCR compatibility.

JoshCheek:
Np, life gets flung at all of us, thanks for the merge :)

sferik:
:+1:

PR #484, opened 2015-06-01
Update ENV Keys docs to use `options` rather than `request`

natebird:
The documentation for the ENV key isn’t `request`. Rather it is `options`. Since this isn’t very well documented for proxy settings it was unclear how to implement proxy settings. Changing to `options` makes it clear as it can be mapped back to the documentation for `timeout` as well.

### DIFF
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -68,21 +68,21 @@ module Faraday
     end
 
     # ENV Keys
-    # :method - a symbolized request method (:get, :post)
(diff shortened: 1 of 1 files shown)


mislav:
Hello there, sorry you got confused. The key is still `:request` not `:options`, however. This documentation refers to ConnectionOptions keys, not the API for the Request class.

Did you have trouble configuring a proxy server? Can I help you with that?

mislav:
Closing in the meantime because the original issue was likely a misunderstanding.

PR #485, opened 2015-06-02
Add SOCKS support with socksify

chagar:
Use a SOCKS (SOCKS5) connection with Faraday, using the socksify gem.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -89,7 +89,13 @@ module Faraday
 
       def net_http_connection(env)
         if proxy = env[:request][:proxy]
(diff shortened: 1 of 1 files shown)


mislav:
I have no experience with SOCKS, but to add a feature to an adapter in Faraday, the attempt must be made to add the same feature to all other adapters as well. Think you can look into it?

Also, we have automated tests for all HTTP and proxy server functionality. In order to add tests for this, you would have to spin up a SOCKS [server in test mode](https://github.com/lostisland/faraday/blob/033e7b420d662bf42cd7d1bc2c3698509bc6d78b/script/test#L123-124). If you think you can do that, please explore this direction. I don't have the bandwidth to work on this myself. Thanks!

PR #489, opened 2015-06-12
Failing test case for query string parsing

mikegee:
`Rack::Utils.parse_nested_query` (what Rails uses) parses query string differently that Faraday. This test case illustrates the issue.

### DIFF
--- a/test/connection_test.rb
+++ b/test/connection_test.rb
@@ -61,6 +61,11 @@ class TestConnection < Faraday::TestCase
     assert_equal({'a' => '1'}, conn.params)
   end
 
+  def test_initialize_stores_default_params_from_uri_like_rack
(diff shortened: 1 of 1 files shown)


mikegee:
I suspect this test case doesn't belong in this file. And, I haven't looked into the implementation yet. Helpful suggestions are welcome.

mislav:
[parameters_test](https://github.com/lostisland/faraday/blob/1316f4cbd9afb42c5c7a6bb9f09d84130ae8fb0c/test/parameters_test.rb) is the first place to start looking. Also, please search existing issues on the subject first. Thanks!

mislav:
Closing this in favor of my fix in #531. Please take a look!

PR #493, opened 2015-07-01
Maximum interval option for Retry middleware

shiroyasha:
It is often useful to limit the interval in the retry middleware, especially when using a `backoff_factor` that increments the interval exponentially.

Please review.


mislav:
Looks great; thanks!

PR #494, opened 2015-07-06
[httpclient] Fixes default certificate store not having default paths.

johnbellone:
This came up while debugging an issue with the [latest Berkshelf API
release](https://github.com/berkshelf/berkshelf-api-client/issues/9). It seems that the net_http.rb adapter did this in the exact
same way (calls #set_default_paths on the X509::Store) but for some
reason it was neglected when httpclient.rb adapter was written.

It is expectation that, by default, the system certificate store is utilized.

### DIFF
--- a/lib/faraday/adapter/httpclient.rb
+++ b/lib/faraday/adapter/httpclient.rb
@@ -72,14 +72,14 @@ module Faraday
 
       def configure_ssl(ssl)
         ssl_config = client.ssl_config
+        ssl_config.verify_mode = ssl_verify_mode(ssl)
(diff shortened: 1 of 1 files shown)


johnbellone:
Its important to note that this was lifted nearly verbatim from the [net_http.rb implementation](https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http.rb#L98-127).

afiune:
Are you guys going to create a new gem with this fix.. Since it fixes all of our problems! :+1: :smile:

Issue #317, opened 2013-10-28
Logger middleware just crashes


PavelPenkov:
This simple program crashes because when logger is called there no headers and body yet. 

``` ruby
require 'faraday'

conn = Faraday.new(url: "http://google.com") {|config| config.response :logger}
conn.get "/"
```

```
/Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response/logger.rb:31:in `dump_headers': undefined method `map' for nil:NilClass (NoMethodError)
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response/logger.rb:25:in `block in on_complete'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/logger.rb:376:in `add'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/logger.rb:404:in `debug'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response/logger.rb:25:in `on_complete'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response.rb:9:in `block in call'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response.rb:63:in `on_complete'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response.rb:8:in `call'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/response/logger.rb:20:in `call'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/connection.rb:253:in `run_request'
    from /Users/synapse/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/faraday-0.8.8/lib/faraday/connection.rb:106:in `get'
[...]

Issue #121, opened 2012-02-22
Automatically set @handler to Faraday.default_adapter if left unspecified.


envygeeks:
Normally for our simple scripts we set Faraday.default_adapter once at the top however while working with Github who (no offense) loves redirects we have to use faraday_middleware (well not have to use) to follow redirects which requires us to do a block (I assume) I think it would be better if Faraday assumed the handler should be set to Faraday.default_adapater if it's not set with use in the block, this saves some coding and redundant code IMO.

henrik:
If this is implemented, the changes in #496 can probably be reverted (if they were merged in the first place).

bricker:
+1 to this. At the very least, an error that no adapter was specified, or better just use the `default_adapter` if nothing was specified.

ys:
The name is confusing, when you have to add middlewares you expect to just add middlewares and keep the `default_adapter` in place.

This is silly that we end up with no adapter at all. You never want that...

iMacTia:
I just want to ensure everyone that we're internally discussing a strategy to manage this.
I know it sounds like an easy check, but as @mislav already explained we have an issue at the moment related with the freedom people have to add Adapters into the middlewares stack.
In fact, at the moment and "adapter" is nothing else than a middleware, like any other, performing the request.
Now, when you define your own middlewares stack, we loose the ability to detect the adapter from the other middlewares, unless you use the `#adapter` method when building the connection. Unfortunately, the use of it is not mandatory at the moment, making the adapter detection impossible.

ys:
Cool! Thanks for the headsup here:)  
  

  
On Nov 14 2016, at 2:45 pm, Mattia <notifications@github.com> wrote:  

> I just want to ensure everyone that we're internally discussing a strategy
to manage this.  
I know it sounds like an easy check, but as
[@mislav](https://github.com/mislav) already explained we have an issue at the
moment related with the freedom people have to add Adapters into the
middlewares stack.  
In fact, at the moment and "adapter" is nothing else than a middleware, like
any other, performing the request.  
Now, when you define your own middlewares stack, we loose the ability to
detect the adapter from the other middlewares, unless you use the `#adapter`
method when building the connection. Unfortunately, the use of it is not
mandatory at the moment, making the adapter detection impossible.

>

> —  
You are receiving this because you commented.  
Reply to this email directly, [view it on GitHub](https://github.com/lostislan
d/faraday/issues/121#issuecomment-260337804), or [mute the
[...]

PR #496, opened 2015-07-09
More helpful Logger error on missing adapter

henrik:
The current error message has confused people.

As discussed in #317, there are better longer-term solutions, but no one has implemented those yet, so maybe good could be better than perfect here?

I'll add a note in #121 to revert this (if it's merged) if the proper fix is implemented.

I'm sure this isn't the best way to catch this error – please do suggest improvements and I'll do my best.

cc @mislav

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -25,6 +25,11 @@ module Faraday
     end
 
     def on_complete(env)
+      # Not ideal, but better than a cryptic error: https://github.com/lostisland/faraday/issues/317
+      if env.status.nil?
+        raise 'Response data is empty! Did you forget to explicitly specify an adapter?'
+      end
+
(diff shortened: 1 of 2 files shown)


mislav:
Thanks, but as I said in https://github.com/lostisland/faraday/issues/317#issuecomment-120070515, this only helps people who have added the logger middleware on their stack. Anyone else who didn't but also forgot to specify an adapter is still screwed. I don't look forward to maintaining workarounds that only help a tiny percentage of users.

henrik:
@mislav Thank you. Alright, that's fair. I hate running into a problem and not making it less bad for the next person (even if it doesn't fix the wider problem), and this seems it might be relatively easy to maintain (small and localized and tested), but I can't very well argue with a maintainer about maintainability :) I'll close the ticket.

PR #497, opened 2015-07-14
Fix up the README file

pvdb:
@technoweenie, @mislav _et al._ - the main purpose of this PR is to add `Net::HTTP::Persistent` to the list of available adapters, as per the changes in #123, and to make further improvements to the README file started in ffa01e25:

* mark `Net::HTTP` as the default adapter
* add `Net::HTTP:Persistent` as an available but experimental adapter
* add docu hyperlinks for both `Net::HTTP` and `Net::HTTP::Persistent`

Hope these changes meet your approval... thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,6 @@ Faraday supports these adapters:
 * [EventMachine][]
 * [HTTPClient][]
 
-[net_http]: http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTP.html
-[net_http_persistent]: https://github.com/drbrain/net-http-persistent
-
 It also includes a Rack adapter for hitting loaded Rack applications through
 Rack::Test, and a Test adapter for stubbing requests by hand.
 
@@ -213,12 +210,14 @@ of a major release, support for that Ruby version may be dropped.
 Copyright (c) 2009-2013 [Rick Olson](mailto:technoweenie@gmail.com), Zack Hobson.
 See [LICENSE][] for details.
 
-[travis]:    http://travis-ci.org/lostisland/faraday
-[excon]:     https://github.com/geemus/excon#readme
-[typhoeus]:  https://github.com/typhoeus/typhoeus#readme
-[patron]:    http://toland.github.com/patron/
-[eventmachine]: https://github.com/igrigorik/em-http-request#readme
-[httpclient]: https://github.com/nahi/httpclient
-[jruby]:     http://jruby.org/
-[rubinius]:  http://rubini.us/
-[license]:   LICENSE.md
+[net_http]:            http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTP.html
+[net_http_persistent]: https://github.com/drbrain/net-http-persistent
+[travis]:              http://travis-ci.org/lostisland/faraday
+[excon]:               https://github.com/geemus/excon#readme
+[typhoeus]:            https://github.com/typhoeus/typhoeus#readme
+[patron]:              http://toland.github.com/patron/
+[eventmachine]:        https://github.com/igrigorik/em-http-request#readme
+[httpclient]:          https://github.com/nahi/httpclient
+[jruby]:               http://jruby.org/
+[rubinius]:            http://rubini.us/
+[license]:             LICENSE.md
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,18 +1,22 @@
 # Faraday
 
 Faraday is an HTTP client lib that provides a common interface over many
-adapters (such as Net::HTTP) and embraces the concept of Rack middleware when
+adapters (such as `Net::HTTP`, the default adapter) and embraces the concept of Rack middleware when
```
mislav: This note wasn't necessary. Instead, the "default" indicator could be added to the list below.

README.md
```diff
@@ -1,18 +1,22 @@
...
 processing the request/response cycle.
 
 Faraday supports these adapters:
 
-* Net::HTTP
+* [`Net::HTTP`][net_http]
+* [`Net::HTTP::Persistent`][net_http_persistent] _(experimental)_
```
mislav: The "experimental" note here isn't necessary. Net::HTTP::Persistent works pretty well in my experience. Users should switch between adapters at their own risk.

README.md
```diff
@@ -1,18 +1,22 @@
...
 * [Excon][]
 * [Typhoeus][]
 * [Patron][]
 * [EventMachine][]
 * [HTTPClient][]
 
+[net_http]: http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTP.html
```
mislav: Could we move the URLs down to group them with the rest of URLs such as for Excon? Thanks

pvdb:
Thanks for reviewing, @mislav, I'll action your feedback now.

Re. the experimental nature of the `Net::HTTP::Persistent` adaptor: I labeled it as such because it is still labeled as such [in the source code](https://github.com/lostisland/faraday/blob/81f16593a0138ec58bb6f25e1c2804e91589662f/lib/faraday/adapter/net_http_persistent.rb#L5-L8) as well... do you prefer me leave the label in the README file, or shall I remove it from both the README file and the source code?

mislav:
> do you prefer me to leave the experimental label in the README file, or shall I remove it from both the README file and the source code?

I think removing the "experimental" comment from both places is OK, because the adapter works well enough.

pvdb:
> I think removing the "experimental" comment from both places is OK, because the adapter works well enough.

Cool, that and all your other comments addressed in subsequent commits... all looking good?  Thanks!

README.md
```diff
@@ -1,12 +1,13 @@
 # Faraday
 
 Faraday is an HTTP client lib that provides a common interface over many
-adapters (such as Net::HTTP) and embraces the concept of Rack middleware when
+adapters (such as `Net::HTTP`) and embraces the concept of Rack middleware when
```
mislav: Could we not wrap module names as inline code blocks? I am of the conviction that module names can occur in prose without marking them up as code. I think it looks better.

README.md
```diff
@@ -1,12 +1,13 @@
...
 processing the request/response cycle.
 
 Faraday supports these adapters:
 
-* Net::HTTP
+* [`Net::HTTP`][net_http] _(the default adapter)_
```
mislav: Nitpick: it can just say "(default)" here

README.md
```diff
@@ -209,12 +210,14 @@ of a major release, support for that Ruby version may be dropped.
...
-[excon]:     https://github.com/geemus/excon#readme
-[typhoeus]:  https://github.com/typhoeus/typhoeus#readme
-[patron]:    http://toland.github.com/patron/
-[eventmachine]: https://github.com/igrigorik/em-http-request#readme
-[httpclient]: https://github.com/nahi/httpclient
-[jruby]:     http://jruby.org/
-[rubinius]:  http://rubini.us/
-[license]:   LICENSE.md
+[net_http]:            http://ruby-doc.org/stdlib/libdoc/net/http/rdoc/Net/HTTP.html
+[net_http_persistent]: https://github.com/drbrain/net-http-persistent
```
mislav: Nitpick: You can use a shorter alias for the Markdown link here so that the rest of the references don't need to be indented so deep :wink:

pvdb:
All done, @mislav :smile:

mislav:
Thanks!

pvdb:
> Thanks!

W00t, w00t... thanks for merging!

PR #498, opened 2015-07-16

hudon:
Just brought https://github.com/lostisland/faraday/pull/470 up to date

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,11 @@ bundler_args: --without development --path vendor/bundle --jobs=3 --retry=3
 cache: bundler
 
 rvm:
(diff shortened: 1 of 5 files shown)


mislav:
Thanks for the suggestion. I would not include https://github.com/hudon/faraday/commit/8fe2187026abafe5112d2f637ffc1d83feabfd77 here because it doesn't belong to this PR.

Is there a chance we could support `idle_timeout` in other adapters? I'm not a fan of features being exposed through Faraday that only 1 adapter supports.

mislav:
See https://github.com/lostisland/faraday/pull/470#issuecomment-254296415

PR #500, opened 2015-07-19
fix detection of environment variable "no_proxy".

vmi:
Hi,

I found out that Faraday does not detect environment variable "no_proxy".
This PR fixes it.


vmi:
I close this ticket because I didn't have enough consideration for the problem.

PR #501, opened 2015-07-20
Set req.body example as standard json format.

gazeldx:
'{ "name": "Unagi" }' is not a standard json format. 

Rubyist may like pass a hash.to_json for this scenario.


mislav:
Thanks for the suggestion, but I don't know what "standard JSON" is. All I know is that `{ "name": "Unagi" }` is perfectly valid JSON, parseable by any JSON parser, so I'm fine leaving it as-is.

gazeldx:
Yes you are right.

PR #502, opened 2015-07-28
:verify_callback ssl option for Net::HTTP adapter

AlexanderPavlenko:
This option is useful for certificate pinning — https://stackoverflow.com/questions/22093042/implementing-https-certificate-pubkey-pinning-with-ruby

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -106,6 +106,8 @@ module Faraday
         http.ca_path      = ssl[:ca_path]      if ssl[:ca_path]
         http.verify_depth = ssl[:verify_depth] if ssl[:verify_depth]
         http.ssl_version  = ssl[:version]      if ssl[:version]
+
+        http.verify_callback = ssl[:verify_callback] if ssl[:verify_callback]
       end
 
       def ssl_cert_store(ssl)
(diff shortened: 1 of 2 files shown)


thomasgallagher:
This would be very useful

mltsy:
Yes!  Please :+1:

mltsy:
@AlexanderPavlenko would you be able to add a test for this? @mislav would you merge this very simple feature if it has a test?

jimpo:
Can you also add to the NetHttpPersistent adapter which overrides this method?

mltsy:
Is there a reason Net:HTTP::Persistent overrides the configure_ssl method?  Are some of the config options irrelevant for a persistent connection or something? I can't see why they would be... or is it just to avoid automatically setting `use_ssl = true` for some reason?  I feel like if we make an update to that here, we should remove the redundancy at the same time if possible.

mltsy:
And I will answered my own question: `http.use_ssl = true` is not a thing in Net::HTTP::Persistent.  Maybe it would be better to just move that to its own method (`enable_ssl`) and override that method to do nothing in the Persistent adapter :)  Then all the ssl options (which seem to be identical for both libraries) don't have to be repeated.

iMacTia:
Hi All,

I'm sorry but we're not accepting PR of this type as explained on https://github.com/lostisland/faraday/pull/470#issuecomment-254296415.

You're welcome to share your thoughts on #648

PR #504, opened 2015-07-30
Don't set a default cert_store for Net::HTTP

zanker:
When you set a `cert_store` for Net::HTTP, it overrides the `ca_file` and `ca_path` options. Since a default one is always set, you can't actually reconfigure Net::HTTP without doing it through the `cert_store` option.

Net::HTTP by default already does this: [Net::HTTP](https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L897) uses [OpenSSL::SSL::SSLContext::DEFAULT_CERT_STORE](https://github.com/ruby/openssl/blob/master/lib/openssl/ssl.rb#L71) which calls `set_default_paths` and as part of `set_params` it sets `cert_store = DEFAULT_CERT_STORE`.

I tested this by setting an invalid `ca_file`, and it still worked until I removed the code which sets `cert_store` by default, and then it failed validation. I can't find the OpenSSL C code to specifically show that it prioritizes `cert_store` though.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -98,22 +98,14 @@ module Faraday
       def configure_ssl(http, ssl)
         http.use_ssl      = true
         http.verify_mode  = ssl_verify_mode(ssl)
-        http.cert_store   = ssl_cert_store(ssl)
 
+        http.cert_store   = ssl[:cert_store]   if ssl[:cert_store]
         http.cert         = ssl[:client_cert]  if ssl[:client_cert]
         http.key          = ssl[:client_key]   if ssl[:client_key]
         http.ca_file      = ssl[:ca_file]      if ssl[:ca_file]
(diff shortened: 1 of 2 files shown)


mislav:
This was added in #76 because back then Ruby (1.8.7 it seems) sometimes didn't find CA certificates without it. If we remove it, we risk regressing to the same issue since we still support 1.8.7.

Are you saying `ca_file` and `ca_path` are unusable right now? Can we instead avoid setting default `cert_store` if one of those options are specified?

zanker:
That would depend on how `OpenSSL::SSL::SSLContext` works under the hood. I suspect that if you set `cert_store`, it takes precedence over `ca_file` and `ca_path` and so it'll still break.

If MRI 1.8.7 is still required, it might be better to just add a version switch, as ugly as it is?

mislav:
How about what I suggested above: only set `cert_store` if you haven't supplied anything for `ca_file` and `ca_path`?

iMacTia:
Closing as stale

PR #506, opened 2015-08-04
Faraday::Response::Logger: Update tests

darrhiggs:
Looking through the tests for the `Faraday::Response::Logger` I saw that
the tests do not test that only the request/response body are logged when
the logger `:bodies` option is a hash.

Tests updated to use `refutes_match` when either the request/response
bodies are not whitelisted setup options.


Zorbash:
+1

mislav:
Thanks!

PR #507, opened 2015-08-06
Allows passing of adapter specific ssl options

michaelirey:
Running SSL=yes script/test
Gives: 494 runs, 837 assertions, 1 failures, 21 errors, 0 skips

This PR does not cause these failures as they already exist in master.

Added ssl option for excon client_key_pass using a new key in SSLOptions called adapter_specific.


presidentJFK:
:+1:

iMacTia:
Hi @michaelirey  and sorry if it take so long to reply to your PR.
The change you're proposing is actually not needed in your case as the options you want to add can be passed in the `connection_options`.

All you need to do is (WARNING: untested code):

```ruby
Faraday.new(url: api_url) do |f|
  f.adapter :exon, {ssl_version: ..., client_key_pass: ...}
end
```

PR #508, opened 2015-08-06
Update options for TyphoeusAdapter

mehulkar:
I'm not sure if this is the correct way to fix the problem I'm having. Faraday specifies a particularly old version of Typheous. I have other gems that are depending on more recent versions of Typheous. In combination, when I have a more recent version of Typhoeus loaded, my specs start to fail because Faraday uses the TypheousAdapter and is making requests to typhoeus using the old option key. 

This change does not break any existing tests, so I'm not sure what i can do about breaking changes. This change _does_ fix my issue.

### DIFF
--- a/lib/faraday/adapter/typhoeus.rb
+++ b/lib/faraday/adapter/typhoeus.rb
@@ -40,7 +40,7 @@ module Faraday
           :method  => method,
           :body    => env[:body],
           :headers => env[:request_headers],
-          :disable_ssl_peer_verification => (env[:ssl] && env[:ssl].disable?)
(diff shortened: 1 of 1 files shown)


mislav:
Sorry for the late reply.

Faraday specifies an old version of Typhoeus in its tests because newer versions of Typhoeus ship with [their own built-in adapter](https://github.com/typhoeus/typhoeus/blob/01995fab575b2087a8e64ced43404cb199fa8b7d/lib/typhoeus/adapters/faraday.rb). Therefore we stopped maintaining the Faraday adapter.

Checks the code documentation above for an example how to load the adapter!

Issue #476, opened 2015-04-30
ArgumentError: invalid byte sequence in US-ASCII


jaimeiniesta:
We're getting this error using faraday 0.9.1, ruby 2.2.0 and OSX Yosemite:

``` ruby
> url = 'http://www.lemagit.fr/actualites/4500244568/SentinelOne-mise-sur-le-poste-de-travail-pour-detecter-les-attaques?asrc=EM_MDN_42264256&utm_medium=EM&utm_source=MDN&utm_campaign=20150428_SearchSecurity.fr%20:%20toute%20l%27information%20S%e9curit%e9%20-%20Avril%202015_'
 => "http://www.lemagit.fr/actualites/4500244568/SentinelOne-mise-sur-le-poste-de-travail-pour-detecter-les-attaques?asrc=EM_MDN_42264256&utm_medium=EM&utm_source=MDN&utm_campaign=20150428_SearchSecurity.fr%20:%20toute%20l%27information%20S%e9curit%e9%20-%20Avril%202015_"

> conn = Faraday.new(url: url)
 => #<Faraday::Connection:0x007fdf59ac04c8 @parallel_manager=nil, @headers={"User-Agent"=>"Faraday v0.9.1"}, @params={"asrc"=>"EM_MDN_42264256", "utm_medium"=>"EM", "utm_source"=>"MDN", "utm_campaign"=>"20150428_SearchSecurity.fr : toute l'information S\xE9curit\xE9 - Avril 2015_"}, @options=#<Faraday::RequestOptions (empty)>, @ssl=#<Faraday::SSLOptions (empty)>, @default_parallel_manager=nil, @builder=#<Faraday::RackBuilder:0x007fdf59ac0298 @handlers=[Faraday::Request::UrlEncoded, Faraday::Adapter::NetHttp]>, @url_prefix=#<URI::HTTP http://www.lemagit.fr/actualites/4500244568/SentinelOne-mise-sur-le-poste-de-travail-pour-detecter-les-attaques>, @proxy=nil>

> conn.get
ArgumentError: invalid byte sequence in US-ASCII
    from /Users/jaime/.rvm/gems/ruby-2.2.0/gems/faraday-0.9.1/lib/faraday/utils.rb:174:in `gsub'
    from /Users/jaime/.rvm/gems/ruby-2.2.0/gems/faraday-0.9.1/lib/faraday/utils.rb:174:in `escape'
[...]

PR #513, opened 2015-08-21
Treat URL query params as binary data

davoclavo:
If a url with non properly UTF-8 encoded characters as part of the query
is passed to a connection, an `invalid byte sequence` error is raised.
e.g. http://example.com/search?q=%A0

We should treat query strings as binary data to avoid having this issue

references https://github.com/lostisland/faraday/issues/476

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -188,7 +188,11 @@ module Faraday
       }.tr(' ', '+')
     end
 
-    def unescape(s) CGI.unescape s.to_s.force_encoding(Encoding::BINARY) end
+    def unescape(s)
+      string = s.to_s
+      string.force_encoding(Encoding::BINARY) if RUBY_VERSION >= '1.9'
+      CGI.unescape string
+    end
 
     DEFAULT_SEP = /[&;] */n
 
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -188,7 +188,7 @@ module Faraday
       }.tr(' ', '+')
     end
 
-    def unescape(s) CGI.unescape s.to_s end
+    def unescape(s) CGI.unescape s.to_s.force_encoding(Encoding::BINARY) end
 
     DEFAULT_SEP = /[&;] */n
 
--- a/test/utils_test.rb
+++ b/test/utils_test.rb
@@ -65,7 +65,8 @@ class TestUtils < Faraday::TestCase
   end
 
   def test_unescaping_and_escaping_invalid_byte_sequences
-    invalid_utf_sequence = "%A0".force_encoding(Encoding::UTF_8)
+    invalid_utf_sequence = "%A0"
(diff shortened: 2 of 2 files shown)


mislav:
Thanks for the fix. Faraday supports Ruby 1.8, which doesn't have encodings support, so `force_encoding` won't work there. Maybe no workaround is even needed for Ruby 1.8?

In which scenarios Faraday calls unescape on mis-encoded URLs?

davoclavo:
@mislav Yeah, you were right, we don't need to use `force_encoding` for Ruby 1.8 as it doesn't even deal with encodings at all.

AFAIK Faraday calls unescape only on the query params, in this example: http://example.com/search?q=%00%A0 - it will call `Utils.unescape` only on `"%00%A0"`

davoclavo:
Any ideas on why the Travis CI build is failing?

fschwahn:
Just curious about this, because we are hitting that error. The travis error is a segfault in ruby 2.2.0 while running simplecov - it's unrelated to this PR. The specs are all passing.

alfie-max:
It's 2016 guys.. can we get this merged in if this fixes the issue? we are facing a lot of errors with similar url's

iMacTia:
Hi all, coming back to this in the process of going through all issues.
Happy to know form @alfie-max that this is something helpful for others so would like to push it soon.

@davoclavo, I'm really sorry for the delay but tests have been recently fixed and are now passing in the master branch, would you mind pulling the latest version and checking if that fixes this PR as well?

davoclavo:
Tests are 🍏 now!

iMacTia:
Cool, thanks @davoclavo.
I'm trying to replicate the error on the v0.9.2, but I'm getting the following:

```
> Faraday.new.get("http://example.com/search?q=%A0")
ArgumentError: invalid byte sequence in UTF-8
	from /Users/iMacTia/.rvm/gems/ruby-2.3.1/gems/faraday-0.9.2/lib/faraday/utils.rb:186:in `gsub'
	from /Users/iMacTia/.rvm/gems/ruby-2.3.1/gems/faraday-0.9.2/lib/faraday/utils.rb:186:in `escape'
```

So from what I see the issue is on the `escape` method, not `unescape`.
Am I misunderstanding the issue you're trying to solve?
Does the above work with your branch?

iMacTia:
The following example works properly with no issues:

```ruby
Faraday.new.get("http://example.com/search?q=%A0".force_encoding(Encoding::BINARY))
```

So given that:

1. The scope of the issue is not clear enough (still can't understand if the fix works with my example as I get the error on `escape` method and not `unescape`)
2. We discussed internally and we're not sure managing encoding inside Faraday is a good idea (what if the Binary encoding have some side-effects in some edge cases?)

I'm then going to close this and suggesting whoever is having this issue to apply the encoding before passing the url to Faraday.

Happy to reopen and discuss if you want to add more details.

PR #518, opened 2015-09-08
Add MRI 2.2

ekampp:
Travis passes on MRI 2.2. This should be reflected in the README.


mislav:
Thanks! I've changed it to simply say "Ruby 1.8.7+" instead.

ekampp:
:+1:

Issue #454, opened 2015-01-04
Replacing Faraday with Hurley


technoweenie:
Though Faraday has been pretty solid for awhile, it has a number of big issues.

* Requests are made with an array of rack-like middleware and/or adapters.  This makes it cumbersome to swap adapters out.
* Url handling has weird edge case issues.
* Adapters are of varying quality, and depend on specific Faraday releases.

I tried fixing this a couple years ago (#207), but eventually gave up due to the complexity in migrating from Faraday v0.9 to the mythical Faraday v1.0.

Instead, I sat down this last week and wrote [Hurley](https://github.com/lostisland/hurley), a major refresh of the Faraday code base.  

* No more rack internals.
* A `Hurley::Client` can swap out connections without changing its properties.
* Optional support for `Addressable::URI`.
* Zero 3rd party runtime dependencies.
* Url handling is improved, with a clear API for changing it.
* Built-in automatic redirection following.

## What about Faraday?

It hasn't seen much activity since v0.9's release.  Hurley is what Faraday v1.0 would have been.  Being a separate library means it can be loaded alongside Faraday during a gentle code migration.

So, only bug fixes for now (unless someone else wants to take the reins).  I'm hoping to get Faraday-related projects such as Sawyer and Octokit upgraded to Hurley soon.

## WTF is Faraday and Hurley?

Dumb LOST nerd lore :)  Daniel Faraday is a scientist with brain damage that goes back in time and gets shot by his mother years before he was born.  Hurley Reyes is brought to the island by Jacob, and eventually inherits and runs the island for many years.

## Comments?

I know this is a big change.   It's not ideal, but I think better than letting Faraday rot in its current state.

sferik:
:+1:

jfelchner:
I don't see why these code changes couldn't have been made in `faraday`?  A `1.0` release would have allowed breaking backward compatibility.  Now we have another gem which is extremely similar that people have to deal with?  Am I missing something?

mislav:
@technoweenie Would you be willing to write some sort of migration guide from Faraday 0.9 to Hurley? Then we could update our README by linking to it. Such guide could be hosted on Hurley repo, for example.

Having maintained Faraday for a long time, I'm a bit :cry: that Faraday will never see a 1.0, but I'm also proud of the work we did making the 0.8 and 0.9 series useful, stable, and a top HTTP client for a while. Good work everyone :+1:

technoweenie:
[...]

PR #519, opened 2015-09-12
cleanup rakefile

grosser:
- generate_certs is used from script/
 - console == bundle console
 - script/package == rake build (unnecessary since release takes care of it)
 - script/release == rake release
 - script/cached-bundle == obsolete, travis does it automatically

@mislav


grosser:
@mislav you around ... anyone else maintaining this project ?

technoweenie:
Big :-1: Having the explicit scripts in there is precisely the point: http://githubengineering.com/scripts-to-rule-them-all/. Now, I'm sure some of them can be cleaned up. If `script/release` is simply just `bin/rake release` or something, that's fine.

Also, re: Faraday maintenance: https://github.com/lostisland/faraday/issues/454

grosser:
@mislav can we make this Hurley thing official then ... did not know Faraday was dead :/

mislav:
I've merged just the other good stuff but kept our explicit `script/release`. Thanks!

PR #520, opened 2015-09-12
cleanup gemfile and helper

grosser:
@mislav

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,22 +4,20 @@ gem 'ffi-ncurses', '~> 0.3', :platforms => :jruby
 gem 'jruby-openssl', '~> 0.8.8', :platforms => :jruby
 gem 'rake'
 
-group :test do
-  gem 'coveralls', :require => false
-  gem 'em-http-request', '>= 1.1', :require => 'em-http'
-  gem 'em-synchrony', '>= 1.0.3', :require => ['em-synchrony', 'em-synchrony/em-http']
-  gem 'excon', '>= 0.27.4'
-  gem 'httpclient', '>= 2.2'
-  gem 'leftright', '>= 0.9', :require => false
-  gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -4,22 +4,20 @@ gem 'ffi-ncurses', '~> 0.3', :platforms => :jruby
...
-  gem 'minitest', '>= 5.0.5'
-  gem 'net-http-persistent', '>= 2.9.4'
-  gem 'patron', '>= 0.4.2', :platforms => :ruby
-  gem 'rack-test', '>= 0.6', :require => 'rack/test'
-  gem 'rest-client', '~> 1.6.0', :platforms => [:jruby, :ruby_18]
-  gem 'simplecov'
-  gem 'sinatra', '~> 1.3'
-  gem 'typhoeus', '~> 0.3.3', :platforms => [:ruby_18, :ruby_19, :ruby_20, :ruby_21]
-end
+gem 'coveralls'
```
grosser: removing the require stuff since we don't use it, except for leftright which blows up when using `bundle console`

faraday.gemspec
```diff
@@ -18,7 +15,6 @@ Gem::Specification.new do |spec|
   spec.licenses = ['MIT']
 
   spec.add_dependency 'multipart-post', '>= 1.2', '< 3'
-  spec.add_development_dependency 'bundler', '~> 1.0'
```
grosser: weird 1-off dev dependency ...

mislav:
Thanks! I cherry-picked the changes I wanted. I left the Gemfile untouched because I like to see the requires even if we don't use them; a `:require => false` indicates to the reader that this gem will be conditionally required in the test suite, like coveralls.

grosser:
OK, the for merging :)

mislav:
Ouch, I just realized I forgot to preserve your authorship info in 6c0a03419c98df97a4b2001a545e361811e2e4ab; sorry about that :disappointed:

grosser:
Don't care ... You will get the blame now :D

PR #521, opened 2015-09-12
ship only essential files

grosser:
- install faster
 - less space used when putting into vendor/cache

@mislav


mislav:
Agreed!

Issue #523, opened 2015-09-15
APIAuth support


santiagodoldan:
Is it possible to use api-auth?

I'm not able to sign the request object, I already created a middleware but I don't have access to the request object which api-auth receives to create the hmac message.

Thanks

santiagodoldan:
I created a pull request #524

technoweenie:
Sorry, we're not accepting new adapters at this time. This [typhoeus adapter for Faraday](https://github.com/typhoeus/typhoeus/blob/master/lib/typhoeus/adapters/faraday.rb) is a great model of how a another gem can add a Faraday adapter.

Some more info on Faraday, and its Hurley rewrite: https://github.com/lostisland/faraday/issues/454

santiagodoldan:
cool, thanks for letting me know about that, will move that adapter to another gem

technoweenie:
Sounds good :+1: I'll close the issues, then.

PR #524, opened 2015-09-15
ApiAuth support

santiagodoldan:
hey guys,

I wasn't able to sign a Faraday request using ApiAuth, I created a new adapter for that, not sure if that's the best place but I don't have access to the request object from a Faraday Middleware.

Usage:

```ruby
connection = Faraday.new(url: 'url') do |faraday|
  faraday.adapter :api_auth, 'key', 'secret'
end
```

what do you think? maybe there is a cleaner way of doing the same

Thanks


technoweenie:
Rejected /cc https://github.com/lostisland/faraday/issues/523#issuecomment-140765391

Issue #408, opened 2014-09-07
typhoeus parallel middleware processing broken in 0.9


nevans:
I know that the typhoeus adapter is part of the typhoeus gem, but I'm trying to debug this and it looks like faraday must have changed something about how it handles middleware for parallel requests.  In particular, I can see that the middleware *is* running on parallel requests, but the resulting env[:body] seems to be getting lost or thrown away at some point.

test.rb:
```` ruby
require "typhoeus"
require "typhoeus/adapters/faraday"
require "faraday_middleware"

class DebugMiddleware < Faraday::Middleware
  def initialize(app, name)
    super(app)
    @name = name
  end
  def call(env0)
    @app.call(env0).on_complete do |env1|
      $stderr.puts "Debug %s: %p" % [@name, env1[:body]]
    end
  end
end

conn = Faraday.new(url: "http://echo.jsontest.com/") do |faraday|
  faraday.use      DebugMiddleware, "after JSON parse "
  faraday.response :json
  faraday.use      DebugMiddleware, "before JSON parse"
  faraday.adapter  :typhoeus
end

r0 = conn.get("/foo/bar/baz/quux")
puts "Final value r0         : %p" % [r0.body]
puts

r1, r2 = nil
conn.in_parallel do
  r1 = conn.get("/foo/bar/key/value")
  r2 = conn.get("/blah/blah/one/two")
end
puts "Final value r1         : %p" % [r1.body]
puts "Final value r2         : %p" % [r2.body]
````

With faraday 0.8.9, faraday_middleware 0.8.8, and typhoeus 0.6.8:
````
$ bundle exec ruby test.rb 
Debug before JSON parse: "{\n   \"baz\": \"quux\",\n   \"foo\": \"bar\"\n}\n"
Debug after JSON parse : {"baz"=>"quux", "foo"=>"bar"}
Final value r0         : {"baz"=>"quux", "foo"=>"bar"}

Debug before JSON parse: "{\n   \"foo\": \"bar\",\n   \"key\": \"value\"\n}\n"
Debug after JSON parse : {"foo"=>"bar", "key"=>"value"}
Debug before JSON parse: "{\n   \"one\": \"two\",\n   \"blah\": \"blah\"\n}\n"
Debug after JSON parse : {"one"=>"two", "blah"=>"blah"}
Final value r1         : {"foo"=>"bar", "key"=>"value"}
Final value r2         : {"one"=>"two", "blah"=>"blah"}
````

[...]

PR #525, opened 2015-09-21
Env custom members should be copied by Env.from(env)

nevans:
Fixes lostisland/faraday#408.

If middleware or adapters add custom members to env, those members should be
retained.

### DIFF
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -270,6 +270,17 @@ module Faraday
     def_delegators :request, :params_encoder
 
     # Public
+    def self.from(value)
+      env = super(value)
+      if value.kind_of?(self)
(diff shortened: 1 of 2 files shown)


nevans:
N.b. this is only a shallow copy of `custom_members`.  I figure that's better than doing no copy.

nevans:
I need this change (or something like it) to make typhoeus/typhoeus#476 work (in addition to several custom middlewares I'm using in one of my apps).

That typhoeus PR is creating a custom `typhoeus_timed_out` member for more fool-proof error handling (rather than just assume `status == 0` means a time out) after parallel queries.  However this is a problem that should be solved for _all_ parallel capable adapters, and perhaps it makes sense to add a non-custom member to `Env` instead.

nevans:
Thanks!

Issue #465, opened 2015-02-03
Faraday::RequestOptions.params_encoder is not called


hunner:
I am trying to use a custom encoder only on certain requests, but not on every request of a connection. I saw `connection.options.params_encoder` and `request.options.params_encoder` as two things to try. I can get my encoder working on the connection, but not for just one request.

If I set `request.options.params_encoder = CustomEncoder`, it appears that when the execution path follows `Faraday::RackBuilder.build_env` -> `Faraday::Connection.build_exclusive_url` it passes the path and params at https://github.com/lostisland/faraday/blob/v0.9.0/lib/faraday/rack_builder.rb#L191 but leaves the params_encoder behind, so when `options.params_encoder` is called at https://github.com/lostisland/faraday/blob/v0.9.0/lib/faraday/connection.rb#L407, it only calls the connection's encoder, not the request params_encoder. Should it be?

Perhaps I don't understand the difference between a request encoder and the connection encoder? The line at https://github.com/lostisland/faraday/blob/v0.9.0/lib/faraday/request/url_encoded.rb#L13 would call the request encoder, but I don't see anywhere else that it would happen.

mislav:
Thanks for the report. It's quite possible that you found a bug. We probably haven't tested `params_encoder` option in the per-request scenario, only per-connection. Ideally, one should be able to tweak this per-request.

PR #528, opened 2015-09-26
Honour request.options.params_encoder

davidcornu:
Fixes https://github.com/lostisland/faraday/issues/465

When setting up a new request, `build_exclusive_url` only uses the global `params_encoder` setting which means request-specific settings are completely ignored.


davidcornu:
cc @hunner

mislav:
Good catch; thanks!

davidcornu:
:tada:

PR #530, opened 2015-10-01
Provide option to not log request/response headers

Zorbash:
Currently an app using Faraday cannot suppress header logging information.
A typical use case, is an application with api clients gem dependencies, which use Faraday in development environment, where one would like to browse logs of external HTTP requests.
Header logging output can be too verbose for this use case.

This PR suggests we add the equivalent of the already present option to suppress logging of request / response body, but for headers. 

Usage:

```ruby
Faraday.new do |c|
  c.response :logger, Logger.new(STDOUT)
end

Faraday.new do |c|
  c.response :logger, Logger.new(STDOUT), headers: false
end

Faraday.new do |c|
  c.response :logger, Logger.new(STDOUT), headers: { request: false }
end

Faraday.new do |c|
  c.response :logger, Logger.new(STDOUT), headers: { response: false }
end
```

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -4,7 +4,7 @@ module Faraday
   class Response::Logger < Response::Middleware
     extend Forwardable
 
-    DEFAULT_OPTIONS = { :bodies => false }
+    DEFAULT_OPTIONS = { :headers => true, :bodies => false }
 
     def initialize(app, logger = nil, options = {})
       super(app)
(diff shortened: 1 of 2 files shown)


Zorbash:
@mislav Any feedback on this?

AlexVPopov:
:+1: that would be a great thing!

iMacTia:
Hi @Zorbash, just hit this PR and it looks very promising.
Unfortunately, tests are not passing due to gem incompatibilities.
Given this change has already been requested by others, could you please pull changes from master so that tests will pass?
I would be happy to merge this PR as soon as it's green for travis 👍

Zorbash:
@iMacTia updated the PR, tests are :green_heart:.

iMacTia:
Amazing 👍 !
Clean and useful, merging now 😉 
Thanks @Zorbash

PR #532, opened 2015-10-06
VCR compatibility fix: untangle the response env mess

mislav:
Ensures that upon Response#finish, the `env` during `response.on_complete` blocks and the `response` object itself is properly populated with properties such as `status` and `body`, and that `env` is identical to `response.env`.

Includes #477, references #378, partially adopts #360

/cc @JoshCheek, @pengwynn


lib/faraday/response.rb
```diff
@@ -61,8 +61,8 @@ def on_complete
 
     def finish(env)
       raise "response already finished" if finished?
-      @on_complete_callbacks.each { |callback| callback.call(env) }
-      @env = Env.from(env)
+      @env = env.is_a?(Env) ? env : Env.from(env)
```
pengwynn: Could we push this down to `Env.from`?

pengwynn:
Looks solid. :+1:

lib/faraday/response.rb
mislav: I think it's a feature that `Env.from` creates a copy by itself. In this case, we don't want a copy so we put up a guard. That's the reason I didn't cherry pick your original commit.

PR #534, opened 2015-10-08
EMHttp and EMSynchrony  SSL validation host fix

ayanko:
Connection host might be already resolved so that SSL validation fails.
We should use host from request uri instead of connection host.

More details: https://github.com/igrigorik/em-http-request/issues/33


mislav:
Thanks!

PR #540, opened 2015-11-08
Remove extraneous call to_s in string interpolation.

StevenXL:
For documentation purposes:

Ruby automatically calls `to_s` on the value returned in a string interpolation. As a result, the explicit call to `to_s` is unnecessary on Line 21 of `logger.rb`.


sbleon:
This looks like a nice simple simplification. 👍

mislav:
Sorry, I don't think `to_s` is hurting anyone here!

PR #541, opened 2015-11-13
let SOCKS5 avaliable

x235:


mislav:
Thanks for the contribution, but we aren't interested in supporting SOCKS5 right now!

PR #544, opened 2015-11-16
Add additional SSL options to net::http and net::http::persistant

aetherknight:
Replaces PR #339 .

* Add  support for `ciphers`, `verify_depth`, and `ssl_timeout` to net::http and net::http::persistant.
* Also add adding `ca_path` and `verify_depth` to net::http::persistant. The underlying library supports these options, just like net::http.

Not touching the other adapters due to limited experience with some of the underlying libraries, or due to inconsistencies with what is supposed to be passed or supported. The libcurl-based HTTP client libraries expect file paths for the client certificate and client private key instead of their corresponding OpenSSL::X509::Certificate or OpenSSL::PKey::* objects, plus replicating the Ruby OpenSSL `verify_callback` API may require significant effort or patching of the client libraries.


iMacTia:
Thanks @aetherknight and apologies for the late review.
We understand your need to get these additional options but we decided to avoid having more adapter-specific features as this goes against the Faraday view of being a general interface for multiple adapters.

PR #545, opened 2015-11-18
Support 'Expect' header with '100-continue' value (currently only used by Net:HTTP)

pvmeerbe:
HTTP supports the 'Expect' header to allow verification of a specific request before actually sending the body data.
Example of practical use : http://stackoverflow.com/questions/14281628/ssl-renegotiation-with-client-certificate-causes-server-buffer-overflow

This is supported by Net::HTTP, however is not enabled by default. The time to wait for the '100-continue' response is configured by setting the 'continue_timeout' parameter  (default = nil)
See http://docs.ruby-lang.org/en/2.0.0/Net/HTTP.html#attribute-i-continue_timeout

However, faraday does not support configuring such a parameter, hence the merge request

Most-likely a similar timeout parameter would be required for other HTTP adapters supporting the Expect header.

### DIFF
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -189,7 +189,7 @@ module Faraday
   end
 
   class RequestOptions < Options.new(:params_encoder, :proxy, :bind,
-    :timeout, :open_timeout, :boundary,
+    :timeout, :open_timeout, :boundary,:continue_timeout,
     :oauth)
 
     def []=(key, value)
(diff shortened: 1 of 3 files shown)


sbleon:
@pvmeerbe the project prefers that new features get added to all adapters that can support them.

@mislav I'm not sure how `RequestOptions` is used. Is it a good idea to add an option that some adapters may ignore? Is there a translation layer that adapts these options to their `Adapter`-specific equivalents?

lib/faraday/options.rb
```diff
@@ -189,7 +189,7 @@ def self.fetch_error_class
   end
 
   class RequestOptions < Options.new(:params_encoder, :proxy, :bind,
-    :timeout, :open_timeout, :boundary,
+    :timeout, :open_timeout, :boundary,:continue_timeout,
```
sbleon: Code style: need a space before the final symbol on this line.

mislav:
Thanks for the contribution, but see #470 for the reason why we don't support features that aren't available in all adatpers.

Issue #418, opened 2014-09-23
Reason Phrases on the Response object


davidbiehl:
I'm working with an API that uses "descriptive" reason phrases in the status of the response. An example:

    HTTP/1.1 400 Wrong JSON format: A JSONObject text must begin with '{' at 1 [character 2 line 1]

There doesn't appear to be a way to obtain the reason phrase from the response object, just the status as a number (400 in the example above).

Looking at the [The HTTP/1.1 RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1.1) we see ...

> The reason phrases listed here are only recommendations -- they MAY be replaced by local equivalents without affecting the protocol.

It seems like it would be reasonable to make them available on the response object with a `#reason_phrase`, or similar, method.

sbleon:
This is an old issue, and a fix is implemented in #547, so I'm closing it.

PR #547, opened 2015-12-03
Adds reason phrase support

sbleon:
Reason phrases are sometimes used to report additional information about why a request failed, so they're very useful in troubleshooting failed request. Building on @davidbiehl's work mentioned in https://github.com/lostisland/faraday/issues/418, this adds reason phrase support to all adapters (except Rack).

@mislav, you said in #418 that you'd be happy to see a pull request about this, so here you go!

### DIFF
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -34,9 +34,10 @@ module Faraday
       env.clear_body if env.needs_body?
     end
 
-    def save_response(env, status, body, headers = nil)
+    def save_response(env, status, body, headers = nil, reason_phrase = "")
       env.status = status
       env.body = body
+      env.reason_phrase = reason_phrase.to_s.strip
       env.response_headers = Utils::Headers.new.tap do |response_headers|
         response_headers.update headers unless headers.nil?
         yield(response_headers) if block_given?
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -34,10 +34,10 @@ module Faraday
       env.clear_body if env.needs_body?
     end
 
-    def save_response(env, status, body, headers = nil, reason_phrase = "")
+    def save_response(env, status, body, headers = nil, reason_phrase = nil)
       env.status = status
       env.body = body
-      env.reason_phrase = reason_phrase.to_s.strip
+      env.reason_phrase = reason_phrase && reason_phrase.to_s.strip
       env.response_headers = Utils::Headers.new.tap do |response_headers|
         response_headers.update headers unless headers.nil?
         yield(response_headers) if block_given?
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -252,7 +252,8 @@ module Faraday
   end
 
   class Env < Options.new(:method, :body, :url, :request, :request_headers,
-    :ssl, :parallel_manager, :params, :response, :response_headers, :status)
+    :ssl, :parallel_manager, :params, :response, :response_headers, :status,
+    :reason_phrase)
 
     ContentLength = 'Content-Length'.freeze
     StatusesWithoutBody = Set.new [204, 304]
--- a/test/adapters/rack_test.rb
+++ b/test/adapters/rack_test.rb
@@ -14,8 +14,12 @@ module Adapters
     include Integration::Common
     include Integration::NonParallel
 
-    # Rack::MockResponse doesn't provide any way to access the reason phrase
-    undef :test_GET_reason_phrase
+    # Rack::MockResponse doesn't provide any way to access the reason phrase,
+    # so override the shared test from Common.
+    def test_GET_reason_phrase
+      response = get('echo')
+      assert_nil response.reason_phrase
+    end
 
     # not using shared test because error is swallowed by Sinatra
     def test_timeout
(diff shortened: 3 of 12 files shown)


sbleon:
Looks like I missed Typhoeus due to testing on Ruby 2.2. One sec...

sbleon:
This is good to go!

davidbiehl:
@sbleon that's awesome! Thanks for your help with getting this implemented with the rest of the adapters :)

sbleon:
@mislav sorry to bug you about this, but I would love to get back on a proper release of this gem. Can you take a look at this PR, please, or is this another maintainer I should bother?

test/adapters/rack_test.rb
```diff
@@ -14,6 +14,9 @@ def adapter_options
     include Integration::Common
     include Integration::NonParallel
 
+    # Rack::MockResponse doesn't provide any way to access the reason phrase
+    undef :test_GET_reason_phrase
```
mislav: Could we add a test that the reason phrase is nil for this adapter?

lib/faraday/adapter.rb
```diff
@@ -34,9 +34,10 @@ def call(env)
       env.clear_body if env.needs_body?
     end
 
-    def save_response(env, status, body, headers = nil)
+    def save_response(env, status, body, headers = nil, reason_phrase = "")
       env.status = status
       env.body = body
+      env.reason_phrase = reason_phrase.to_s.strip
```
mislav: Could the reason phrase be nil if it's missing, instead of an empty string? That way users can test the presence of the phrase more easily.

lib/faraday/options.rb
```diff
@@ -252,7 +252,8 @@ def new_builder(block)
   end
 
   class Env < Options.new(:method, :body, :url, :request, :request_headers,
-    :ssl, :parallel_manager, :params, :response, :response_headers, :status)
+    :ssl, :parallel_manager, :params, :response, :response_headers, :status,
+    :reason_phrase)
```
mislav: I'm not ecstatic about the word `reason_phrase`. Could we use `status_text` instead?

mislav:
Sorry for the late review; and thanks for the contribution!

test/adapters/rack_test.rb
sbleon: Done in 0818a1e.

lib/faraday/adapter.rb
sbleon: Good idea! Done in 0818a1e.

lib/faraday/options.rb
sbleon: @mislav, I think that `status_text` would be a little more intuitive, as "Status Text" seems to be a common colloquial term. (E.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Response_codes).

However, "Reason Phrase" is the proper technical term per http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html, so I'd prefer to keep the name `reason_phrase`, if that's alright with you.

What do you think?

sbleon:
@mislav this PR should be good to, pending your approval of the term "Reason Phrase" per https://github.com/lostisland/faraday/pull/547#discussion_r48784710. Let me know if you really want me to change it to `status_text`. Otherwise, I'd love to get this merged. Thanks!

sbleon:
Hey, @mislav ! I'm sure you're busy, but please respond to my comment above about "Reason Phrase" vs "Status Text". IMHO, this PR is good to merge ;-)

sbleon:
Ping. I'm hoping we can get this merged. It should be good to go.

I really need to get this merged so I can send a PR to [VCR](https://github.com/vcr/vcr) that will make it support reason phrases when hooked into Faraday. Thanks!

sbleon:
@jingweno, @technoweenie, @pengwynn, @sferik, I'm reaching out to you as members of the lostisland org. I'd really like to get this PR merged, but it seems like @mislav is really busy. I'm hoping that one of you is a committer and can help me get this merged.

Sorry to rope so many of you into this. I'm just trying to reduce the number of forks in my Gemfile!

sbleon:
@mislav this PR should be good to go, pending your approval of the term "Reason Phrase" per https://github.com/lostisland/faraday/pull/547#discussion_r48784710. Let me know if you really want me to change it to `status_text` and I'll defer to you. Otherwise, I'd love to get this merged. Thanks!

davidbiehl:
Another thought: get the best of both worlds by aliasing reason_phrase with
status_text. That way you have what's correct per the RFC and also the
colloquial term.

sbleon:
Yeah, I’ve considered that as well. I prefer just `reason_phrase` because
it will help to educate developers about the proper terminology, but Ruby
does love its method aliases, so I’m open to having both.

sbleon:
I would really, really, love to see this merged and released. @sferik @technoweenie @mislav can we move this forward, please?

sbleon:
@iMacTia can you merge this for me? I addressed all of @mislav's code review feedback a while ago.

iMacTia:
This was reviewed an is passing all tests. Moreover, it was asked many times in other issues. Finally, is backward-compatible as it is simply adding more info to ENV.
Good job!

Merging :)

PR #552, opened 2015-12-22
feature: uri pattern for test mocking

MidnightWonderer:
Add regular expression to middleware matcher.

### DIFF
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -85,7 +85,7 @@ module Faraday
         protected
 
         def new_stub(request_method, path, headers = {}, body=nil, &block)
-          normalized_path = (path.is_a? Regexp) ? path : Faraday::Utils.normalize_path(path) 
+          normalized_path = (path.is_a? Regexp) ? path : Faraday::Utils.normalize_path(path)
           (@stack[request_method] ||= []) << Stub.new(normalized_path, headers, body, block)
         end
 
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -85,7 +85,7 @@ module Faraday
         protected
 
         def new_stub(request_method, path, headers = {}, body=nil, &block)
-          normalized_path = (path.is_a? Regexp) ? path : Faraday::Utils.normalize_path(path)
+          normalized_path = path.is_a?(Regexp) ? path : Faraday::Utils.normalize_path(path)
           (@stack[request_method] ||= []) << Stub.new(normalized_path, headers, body, block)
         end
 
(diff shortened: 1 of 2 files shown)


MidnightWonderer:
The failed test has nothing to do with the feature in this pull request.
```
Gem::InstallError: addressable requires Ruby version >= 1.9.0.
Installing rack-test 0.6.3
An error occurred while installing addressable (2.4.0), and Bundler cannot
```

lib/faraday/adapter/test.rb
```diff
@@ -85,7 +85,7 @@ def verify_stubbed_calls
         protected
 
         def new_stub(request_method, path, headers = {}, body=nil, &block)
-          normalized_path = Faraday::Utils.normalize_path(path)
+          normalized_path = (path.is_a? Regexp) ? path : Faraday::Utils.normalize_path(path)
```
mislav: Style nitpick: could it be `path.is_a?(Regexp)` instead?

lib/faraday/adapter/test.rb
```diff
@@ -96,7 +96,7 @@ def matches?(stack, path, headers, body)
 
       class Stub < Struct.new(:path, :params, :headers, :body, :block)
         def initialize(full, headers, body, block)
-          path, query = full.split('?')
+          path, query = (full.is_a? Regexp) ? full : full.split('?')
```
mislav: I feel like it's better to use `full.respond_to?(:split) ? full.split("?") ...`

lib/faraday/adapter/test.rb
```diff
@@ -108,12 +108,20 @@ def matches?(request_uri, request_headers, request_body)
...
-          request_path == path &&
+          path_match?(request_path) &&
             params_match?(request_params) &&
             (body.to_s.size.zero? || request_body == body) &&
             headers_match?(request_headers)
         end
 
+        def path_match?(request_path)
+          if path.is_a? Regexp
+            path =~ request_path
```
mislav: Could we capture the MatchData object by doing `match = request_path.match(path)` and somehow pass it to the block? That way the block could access the values from capture groups

mislav:
Thanks for the feature proposal! Sorry about the CI failure; you can merge in latest master to fix that.

In addition to addressing my thoughts in inline comments, could you also add code docs for this feature? I realize there is very little documentation for the Test adapter right now. Your PR could try to improve this while adding a new feature. At least some example use would be nice.

MidnightWonderer:
@mislav all done
Let me know if you have any more suggestion for new code added.

iMacTia:
Hi @MidnightWonderer and sorry if your PR remained suspended for so much time.
I would like to merge it in, but I think that tests are outdated now. Can you please pull latest changes from master once more? thanks in advance

MidnightWonderer:
@iMacTia done.
The tests are still passed without doing anything.

iMacTia:
This took a long time, but I'm happy to merge it in as it actually makes tests more powerful.
Good work @MidnightWonderer 👍 !

PR #556, opened 2016-01-11
Add `Faraday.respond_to?`.

jish:
There are some methods that Faraday _does_ respond to because of `method_missing`, but it does not report that it does actually respond to some of those methods.

---

I am not very familiar with `Faraday`'s test suite and could not figure out the right place to put these tests, please let me know if they should move elsewhere.

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -93,7 +93,7 @@ module Faraday
     alias require_lib require_libs
 
     def respond_to?(symbol, include_private = false)
-      default_connection.respond_to?(symbol) || super
+      default_connection.respond_to?(symbol, include_private) || super
     end
 
   private
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -92,7 +92,7 @@ module Faraday
 
     alias require_lib require_libs
 
-    def respond_to?(symbol)
+    def respond_to_missing?(symbol, include_private = false)
       default_connection.respond_to?(symbol) || super
     end
 
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -92,6 +92,10 @@ module Faraday
 
     alias require_lib require_libs
 
+    def respond_to?(symbol)
+      default_connection.respond_to?(symbol) || super
+    end
+
   private
     # Internal: Proxies method calls on the Faraday constant to
     # #default_connection.
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -92,7 +92,7 @@ module Faraday
 
     alias require_lib require_libs
 
-    def respond_to_missing?(symbol, include_private = false)
+    def respond_to?(symbol, include_private = false)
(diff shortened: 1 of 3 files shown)


sbleon:
@jish I like what you've done in a1520b6. Thanks!

Can you remove 986e557 from your branch? It seems orthogonal and I'm not sure if it's a good idea yet. If you can do so I think we can get this merged.

jish:
Reverted the Readme change.

lib/faraday.rb
```diff
@@ -92,6 +92,10 @@ def default_adapter=(adapter)
 
     alias require_lib require_libs
 
+    def respond_to?(symbol)
```
mislav: Please define [`respond_to_missing?` instead](https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding). Also, take note that both `respond_to?` and `respond_to_missing?` accept 2 arguments.

test/connection_test.rb
```diff
@@ -411,6 +411,14 @@ def test_init_with_block_yields_connection
     assert_equal '/omnom', conn.path_prefix
   end
 
+  def test_respond_to
+    assert Faraday.default_connection.respond_to?(:get)
```
mislav: These 2 assertions are not necessary. You didn't make any changes to `default_connection`, why would you add assertions that verify its interface?

jish: > "why would you add assertions"

The concept of the test was, "this object delegates method calls to the `default_connection` object" rather than "this object responds to some arbitrary method".

If you prefer the latter, I'm happy to change it.

jish:
I've updated this branch to use `respond_to_missing?` instead. Unfortunately, now the test suite is failing on Ruby `1.8.7`.

I'm not sure how to proceed, please advise.

sbleon:
`respond_to_missing?` was [added in Ruby 1.9.2](http://apidock.com/ruby/v1_9_3_392/Object/respond_to_missing%3F). I think you should go back to `respond_to?`*, but with two arguments like you did with `respond_to_missing?`. 

\* Unless @mislav feels like it’s time to drop support for 1.8.7.

jish:
Switched back to `respond_to?` for Ruby `1.8.7` compatibility.

lib/faraday.rb
```diff
@@ -92,6 +92,10 @@ def default_adapter=(adapter)
 
     alias require_lib require_libs
 
+    def respond_to?(symbol, include_private = false)
+      default_connection.respond_to?(symbol) || super
```
mislav: `include_private` is not forwarded to the inner `respond_to` call

sbleon: Almost there, @jish! Thanks for sticking with this :-)

jish:
Forwarded the `include_private` parameter to the inner `respond_to?` call.

mislav:
Thanks @jish!

PR #557, opened 2016-01-23
Unify ssl certificate options

### LINKED ISSUE #486: Ssl options type mismatch between net/http adapter and excon adapter
When using client_certs and client_keys, there is a difference in behavior between excon and net/http

Excon expects the file paths to the cert and key, while net/http expects actual Certificate and PSA objects

So if you change the adapter, you need to find every place you're using these options and change them. 

I'd expect Faraday to just work when swapping out adapters.

In this case, it probably makes sense to have faraday take the file paths, and then instantiate the objects in the adapters that need them. For backwards compatibility, the net/http adapter could check to see if it gets a Certificate object and just uses that, or loads the data from a file path if given a string.

This way we move to a single unified interface when usage of Faraday does not depend on the adapter in use. 

I'm happy to make the changes if people find this valuable. Let me know.

embs:
net_http adapter instantiates certificate and key when it receives file
paths instead of objects.

Closes #486.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -100,6 +100,13 @@ module Faraday
         http.verify_mode  = ssl_verify_mode(ssl)
         http.cert_store   = ssl_cert_store(ssl)
 
+        if ssl[:client_cert].class == String
+          ssl[:client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl[:client_cert]))
+        end
+        if ssl[:client_key].class == String
+          ssl[:client_key] = OpenSSL::PKey::RSA.new(File.read(ssl[:client_key]))
+        end
+
         http.cert         = ssl[:client_cert]  if ssl[:client_cert]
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for your pull request! But I'm not sure if I want to support this right now.

1. Having an option accept either a String or a different option seems to "magic" for me.
2. The cert file will be read and cert object instantiated on every HTTP request. It's a much better idea to have the user supply this object when configuring the adapter. That way the cert file doesn't have to be re-read over and over again.
3. This behavior just applies to `net/http` adapter. However, in Faraday, it's a good idea to have all adapters share the same behavior. From the referenced issue, it seems that Excon expects these options to file strings. But how about other adapters?
4. There are no tests for this. The tests that exercise SSL support are a bit tricky to set up, but we should have some nevertheless. I don't think we have any client certificate tests yet. We should look into how to set those up!

iMacTia:
closing after @mislav's comprehensive explanation

PR #559, opened 2016-02-08
WebMock test example in README.md

omouse:
Added an example of how to stub requests in an RSpec test suite using WebMock: https://github.com/bblimke/webmock

Code based on examples found here:
* https://github.com/bblimke/webmock#setting-expectations-in-rspec-with-a_request
* https://github.com/bblimke/webmock#setting-expectations-in-rspec-on-the-stub

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ end
 stub = stub_request(:post, "example.com")
   .with(body: "hello world", status: 200)
   .to_return(body: "a response to post")
-expect(stub).to_have_been_requested
+expect(stub).to have_been_requested
 
 expect(
   a_request(:get, "example.com")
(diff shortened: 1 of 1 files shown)


sbleon:
@omouse I'm not sure Faraday should document how to use WebMock.

Maybe a more general statement about how to test your Faraday-based code would be appropriate. "To test your Faraday-based code, you can use any HTTP mocking library that works with your HTTP adapter of choice." We could include links to WebMock and VCR, or a link to https://www.ruby-toolbox.com/categories/mocking_web_requests.

omouse:
@sbleon up to you, I'm good either way

sbleon:
I prefer my approach 😉 . If WebMock's API changes, we shouldn't need to update our docs. So I'd rather point at them than try to document it ourselves.

mislav:
Thanks for the contribution, but closing with the reason @sbleon outlined above.

PR #563, opened 2016-02-29
Fix EM::HTTP SSL patch to trust certificate verified by trusted intermediates

BRMatt:
The Ubuntu team [recently released an update to the `ca-certificates` package](https://launchpadlibrarian.net/242531582/ca-certificates_20141019ubuntu0.12.04.1_20160104ubuntu0.12.04.1.diff.gz) that removes several root certificates, including one that is used to sign the root of the chain for `login.mailchimp.com` (GTE CyberTrust Global Root).

The patched version of `em-http` that is distributed with faraday requires all certificates in the chain to be trusted before it'll verify the connection. This causes ssl verification for login.mailchimp.com to fail.

One of the intermediaries on login.mailchimp.com's certificate chain (Baltimore CyberTrust Root) is included in Ubuntu's/Mozilla's ca-certificates bundle, and seems like it should be enough to verify that the certificate is valid(disclaimer: I'm definitely not an expert at openssl). At any rate, it appears to mimic what `openssl s_client` is doing.

I've created [a test case](https://github.com/geckoboard/em-http-ssl-patch)(tested against ruby 2.1.0) that demonstrates the problem and the fix. You can see what would happen if there's a break between the trusted certificate and mailchimp's certificate by [commenting out this line](https://github.com/geckoboard/em-http-ssl-patch/blob/85f91c269e0e94d789d04675e8ec5309e152236c/fixed-monkey-patch.rb#L15).

This PR currently does not have any automated tests as I'm not entirely sure of the best way to test this in a reliable way, and there don't seem to be any existing tests for this patch.

If you want I could add a test case that includes the certificates that are currently in `login.mailchimp.com's` chain and verify that the patch treats the chain as valid? I guess the main downside to this approach is that the tests could stop working when the certificates expire in a few years(?).

Thanks for all the hard work you folks do maintaining this gem :heart: :sparkling_heart: :sparkles:

### DIFF
--- a/lib/faraday/adapter/em_http_ssl_patch.rb
+++ b/lib/faraday/adapter/em_http_ssl_patch.rb
@@ -18,20 +18,23 @@ module EmHttpSslPatch
       rescue OpenSSL::X509::StoreError => e
         raise e unless e.message == 'cert already in hash table'
       end
-      true
-    else
-      raise OpenSSL::SSL::SSLError.new(%(unable to verify the server certificate for "#{host}"))
     end
+
+    true
   end
 
   def ssl_handshake_completed
     return true unless verify_peer?
 
+    unless certificate_store.verify(@last_seen_cert)
+      raise OpenSSL::SSL::SSLError.new(%(unable to verify the server certificate for "#{host}"))
+    end
+
     unless OpenSSL::SSL.verify_certificate_identity(@last_seen_cert, host)
       raise OpenSSL::SSL::SSLError.new(%(host "#{host}" does not match the server certificate))
-    else
-      true
     end
+
+    true
   end
 
   def verify_peer?


BRMatt:
I'm not entirely sure why the tests aren't passing, they seem to be for an unrelated adapter?

iMacTia:
Thanks @BRMatt and sorry for the late reply.
I'm not a certificates expert either but I'm trying to understand the issue here.

Issue:
So, I understand that login.mailchimp.com have an intermediary in his certificate chain that is not officially trusted and thus the certificate verification is failing.
However, this intermediary can be considered safe as it is recognised by "Ubuntu's/Mozilla's ca-certificates bundle".

Solution:
Make the certificate verification pass even if there's one certificate in the chain that is not trusted.


Am I getting this correctly?

BRMatt:
Hi @iMacTia,

> Thanks @BRMatt and sorry for the late reply.

No worries, we have this patch vendored internally so this PR was mainly to contribute back changes 😄.

> So, I understand that login.mailchimp.com have an intermediary in his certificate chain that is not officially trusted and thus the certificate verification is failing.

No, the intermediary in the chain was officially trusted. The issue was that EM only considered a server's certificate to be trusted if the root, and every certificate between it and the server's certificate was trusted. My understanding is that the root was removed as part of an effort to remove 1024 bit root certificates. (see [these](https://bugzilla.mozilla.org/show_bug.cgi?id=1047011) [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=936304)).

> However, this intermediary can be considered safe as it is recognised by "Ubuntu's/Mozilla's ca-certificates bundle".

That is correct.

> Make the certificate verification pass even if there's one certificate in the chain that is not trusted.

Almost, but not quite. This patch will cause the library to verify the server's certificate if it can form a path from a trusted certificate at any point in the chain to the certificate that the server is presenting.

iMacTia:
Thanks for the clarification @BRMatt, I'm trying to understand if we're decreasing the level of security to manage a once-in-a-while edge case caused by the Ubuntu team removal of root certificates.

What are the security implications in changing the way we're checking the certificate chain?
What if in the future we find a situation where the root certificate is not trustworthy anymore and there are no intermediaries backed by famous companies like Ubuntu's/Mozilla's?

Having a quick read at [Intermediate_Certificate_Authority](https://en.wikipedia.org/wiki/Intermediate_Certificate_Authority), it looks like the root certificate is also the most important one.

Finally, as soon as Mailchimp devs find out about the certificate issue, they can simply regenerate the certificate against a new trusted root CA, can't they?

In short, I don't want to sacrifice security to solve an issue that can be simply solved re-issuing the certificate. This solution can work and be reasonably secure for Mailchimp, but is it secure in all other cases as well?

BRMatt:
> What if in the future we find a situation where the root certificate is not trustworthy anymore and there are no intermediaries backed by famous companies like Ubuntu's/Mozilla's?

So the root has been removed from the trusted store, and no other certificate are in the trusted store? In that case the server's certificate will not be trusted.

> Having a quick read at Intermediate_Certificate_Authority, it looks like the root certificate is also the most important one.

I think it comes across that way because in the example only the "root" is trusted by the browser. In this situation, one of the intermediates is trusted. [This thread](https://security.stackexchange.com/questions/17391/can-an-intermediate-ca-be-trusted-like-a-self-signed-root-ca) mentions that the X.509 spec doesn't actually have a concept of "root CAs", and instead talks about "trust anchors", which are CA certificates the user has deemed trustworthy (e.g. the "intermediate CA certificate" we're discussing).

> Finally, as soon as Mailchimp devs find out about the certificate issue, they can simply regenerate the certificate against a new trusted root CA, can't they?

I'm not sure that's a given. Sites aren't always in control of the certificates that are being served on their behalf, e.g. if you're using AWS' certificate manager, or fastly's "Shared certificate" service, you have no control over the chain in your certificate. Also in this situation there were likely a bunch of browsers/servers that did not yet trust the new "intermediate", but still trusted the old root. They wouldn't be able to remove the root until they were confident that it was no longer necessary.

> In short, I don't want to sacrifice security to solve an issue that can be simply solved re-issuing the certificate. This solution can work and be reasonably secure for Mailchimp, but is it secure in all other cases as well?

My understanding is that this doesn't make the system less secure, but again, I'm not an cryptographer :). It's worth reiterating that the `openssl` command was successfully verifying mailchimp's certificate based solely on the trusted intermediate, so we just mimicked its behaviour. 

Based on the details in that SE thread/RFC, it seems that this PR actually brings EM closer inline with [the RFC's behaviour](https://datatracker.ietf.org/doc/rfc3280/?include_text=1), e.g.:

>    The selection of a trust anchor is a matter of policy: it could be
>   the top CA in a hierarchical PKI; the CA that issued the verifier's
>   own certificate(s); or any other CA in a network PKI.  The path
>   validation procedure is the same regardless of the choice of trust
>   anchor.  In addition, different applications may rely on different
>   trust anchor, or may accept paths that begin with any of a set of
>   trust anchor.
> ...
>    To meet this goal, the path validation process verifies, among other
>   things, that a prospective certification path (a sequence of n
>   certificates) satisfies the following conditions:
>
>      (a)  for all x in {1, ..., n-1}, the subject of certificate x is
>      the issuer of certificate x+1;
>
>      (b)  certificate 1 is issued by the trust anchor;
>
>      (c)  certificate n is the certificate to be validated; and
>
>      (d)  for all x in {1, ..., n}, the certificate was valid at the
>      time in question.

I understand your hesitation with this - cryptography's an easy thing to get wrong, and the absence of specs for the monkey patch didn't really help matters.

We [have some specs for our version of the monkey patch](https://gist.github.com/BRMatt/636da609cf9b92bafa90d6dde31844d7) that roughly verify its behaviour. It could be modified for this PR, but it's worth noting that the certificates used in it are the actual ones we got from login.mailchimp.com, so will stop working when that certificate expires. If the specs were to be added it'd probably be best to create a custom CA cert/server cert with long expiry dates explicitly for these tests.

iMacTia:
Thanks again @BRMatt, I feel a bit more comfortable now but I want to investigate a little further on the matter. In the meantime, we should get tests passing. Only ruby2.2 (SSL=no) was failing so I'm running the build again to double-check if that was a "false positive".
If the test is still failing, can you please double-check if you see any specific error when you run your tests locally as I don't see useful info in the build log.

iMacTia:
Also, please pull latest changes from master as I can see your PR is still building against ruby1.8.7. We fixed some tests recently and pulled support for that ruby, so the two things could be related!

mislav:
@BRMatt Thanks for the fix and for the productive discussion!

I'm not an expert as well, but I found it completely ridiculous and unacceptable that the EM adapter doesn't verify certificates by default, so I added my workaround. It's definitely not perfect.

I guess your solution is legit if it matches the behavior of net/http. If you make a request to mailchimp with net/http adapter, assuming that Ruby was built in a way that it relies on system ca-certificates package (that should be true by default), does it trust the Baltimore CyberTrust Root cert and not attempt to find the root cert in ca-certificates?

BRMatt:
> I'm not an expert as well, but I found it completely ridiculous and unacceptable that the EM adapter doesn't verify certificates by default, so I added my workaround. It's definitely not perfect.

Thanks for the original implementation, it's a bit surreal that this was the situation before your patch!

> I guess your solution is legit if it matches the behavior of net/http. If you make a request to mailchimp with net/http adapter, assuming that Ruby was built in a way that it relies on system ca-certificates package (that should be true by default), does it trust the Baltimore CyberTrust Root cert and not attempt to find the root cert in ca-certificates?

Given that the mailchimp certs have changed since this patch was originally submitted, I'm trying to generate some certificates to make an independent test-case that we can use to verify against net/http and cURL without "taking random internet stranger's word that this is all ok". I'll let you know when I've got one.

iMacTia:
Amazing 👍 !
Keep us posted and thanks for your work

iMacTia:
Hi @BRMatt, do you have any update on this?

BRMatt:
Hey, sorry @iMacTia I didn't get a chance to revisit this. Given the amount of uncertainty here it's probably best closing this. We're actually moving away from eventmachine so this is less important for us than it was last year. Sorry 😫

iMacTia:
No worries @BRMatt, thanks for letting me know! Happy to reopen in case someone else is facing te same issue

PR #564, opened 2016-03-03
HTTP gem support, with Celluloid::IO integration

bbozo:
Yet another attempt at integrating the HTTP gem :) This time because of Celluloid::IO, apparently you need to use a specific socket constructor in conjunction with the HTTP gem to make full use of the Celluloid::IO reactor.

The author states his motivation here https://github.com/httprb/http#another-ruby-http-library-why-should-i-care including

Streaming support is still on the TODO list

All tests are green locally on ruby 2.2.3

### DIFF
--- a/lib/faraday/adapter/http_gem.rb
+++ b/lib/faraday/adapter/http_gem.rb
@@ -3,6 +3,14 @@ module Faraday
     class HttpGem < Faraday::Adapter
       dependency 'http'
 
+      HTTP_EXCEPTIONS = [
+        HTTP::ConnectionError
+      ]
+
+      HTTP_EXCEPTIONS << OpenSSL::SSL::SSLError if defined?(OpenSSL)
+      HTTP_EXCEPTIONS << Java::JavaNet::ConnectException if defined?(Java)
+
+
       def initialize(app, options = {})
         @client = ::HTTP::Client.new options
         super(app)
@@ -21,10 +29,12 @@ module Faraday
               :headers => env[:request_headers],
               :proxy => proxy_options(env)
             }.merge(socket_options(env)))
-        rescue HTTP::ConnectionError => e
-          raise Faraday::Error::ConnectionFailed, e
-        rescue OpenSSL::SSL::SSLError => e
-          raise Faraday::SSLError, e
+        rescue *HTTP_EXCEPTIONS => err
+          if defined?(OpenSSL) && OpenSSL::SSL::SSLError === err
+            raise Faraday::SSLError, err
+          else
+            raise Faraday::Error::ConnectionFailed, err
+          end
         end
 
         save_response env, res.status, res.to_s, res.headers.to_h
--- a/lib/faraday/adapter/http_gem.rb
+++ b/lib/faraday/adapter/http_gem.rb
@@ -1,0 +1,95 @@
+module Faraday
+  class Adapter
+    class HttpGem < Faraday::Adapter
+      dependency 'http'
+
+      def initialize(app, options = {})
+        @client = ::HTTP::Client.new options
+        super(app)
+      end
+
+      def call(env)
+        super
+
+        # TODO: Support streaming.
+        #   HTTP gem expects body to be String or Enumerable,
+        env[:body] = env[:body].read if env[:body].respond_to? :read
+
+        begin
+          res = @client.request(env[:method], env[:url], {
+              :body => env[:body],
+              :headers => env[:request_headers],
+              :proxy => proxy_options(env)
+            }.merge(socket_options(env)))
+        rescue HTTP::ConnectionError => e
+          raise Faraday::Error::ConnectionFailed, e
+        rescue OpenSSL::SSL::SSLError => e
+          raise Faraday::SSLError, e
+        end
+
+        save_response env, res.status, res.to_s, res.headers.to_h
+
+        @app.call env
+      end
+
+      private
+
+      def proxy_options(env)
+        return unless env[:request][:proxy]
+
+        rv = {
+          :proxy_address => env[:request][:proxy][:uri].host,
+          :proxy_port    => env[:request][:proxy][:uri].port
+        }
+
+        if env[:request][:proxy][:user]
+          rv[:proxy_username] = env[:request][:proxy][:user]
+          rv[:proxy_password] = env[:request][:proxy][:password]
+        end
+
+        rv
+      end
+
+      def socket_options(env)
+        rv = {}
+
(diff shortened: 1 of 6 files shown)


bbozo:
Two issues so far:

1. one of the celluloid's dependencies (hitimes) failes on older rubies (`Gem::InstallError: hitimes requires Ruby version >= 1.9.3.` error) (https://travis-ci.org/lostisland/faraday/jobs/113386704)
2. the tests pass, but are marked failed, because of an unknown stack trace when requiring the `http` gem, I'll put this up as an issue with them (https://travis-ci.org/lostisland/faraday/jobs/113386709)

what do you suggest to do? :)

lib/faraday/adapter/http_gem.rb
```diff
@@ -21,10 +29,12 @@ def call(env)
               :headers => env[:request_headers],
               :proxy => proxy_options(env)
             }.merge(socket_options(env)))
-        rescue HTTP::ConnectionError => e
-          raise Faraday::Error::ConnectionFailed, e
-        rescue OpenSSL::SSL::SSLError => e
-          raise Faraday::SSLError, e
+        rescue *HTTP_EXCEPTIONS => err
+          if defined?(OpenSSL) && OpenSSL::SSL::SSLError === err
```
sbleon: It seems a little odd to `rescue` a group of exception classes and then check their class. I think leaving two `rescue` clauses would be simpler and easier to read. Is `OpenSSL` not defined in some cases? That could justify this implementation.

sbleon:
@bbozo check out https://github.com/bbozo/faraday/blob/httprb_support/test/adapters/patron_test.rb for an example of how to avoid running tests for an adapter in unsupported Ruby versions.

sbleon:
With 1.6M downloads currently, it seems like the `http` gem is popular enough to consider supporting in Faraday. It's more popular than some other supported HTTP libs (e.g. `patron`).

lib/faraday/adapter/http_gem.rb
bbozo: I shamelessly took the exception handling pattern from the `Net::HTTP` adapter implementation https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http.rb (reasoning it's one of the oldest and most mature) and `OpenSSL` isn't a part of the HTTP gem's gemspec so I guess there are scenarios when this would be an issue.

sbleon: Fair enough!

mislav:
@bbozo Instead of adding your own workaround for rake, can you rebase on top of latest master to get tests passing on 1.8? Thanks 🙇

bbozo:
@mislav I merged last master as soon as you approved the other pull request ^_^

@sbleon can you help me understand why for example https://travis-ci.org/lostisland/faraday/jobs/120001584 fails? It says no tests failed and I couldn't find other error messages

mislav:
@bbozo Your best bet is to run the test locally in the same environment.

sbleon:
@bbozo there is a stack trace showing up in the middle of the test output,
so something’s going wrong, but not all of the information is visible in
the output. Let us know if you try running locally on 1.9.3 and the other
failing versions and can’t reproduce the non-zero exit code.

bbozo:
@mislav @sbleon 

the stack trace is not a result of an exception as far as I can see, the `require 'http'` command generates it and then successfully passes and all of the tests pass OK.

I have noticed it before, put it forward to the `HTTP` gem team because I thought it might cause issues, but then closed it when I discovered a problem in something else

I have forwarded reopen the issue to the HTTP gem people https://github.com/httprb/http/issues/311 but it's really strange and it doesn't happen outside of Faraday tests

``` ruby
bbozo@eva:~/dev/faraday (httprb_support) $ irb
2.2.3 :001 > require 'http'
 => true 
```

Do you have any idea what in Faraday could be interacting with the HTTP gem to cause the stack trace?

bbozo:
One more hint:

* https://travis-ci.org/lostisland/faraday/jobs/120004629 - stack trace & test is green

this is probably why I closed the issue with the HTTP people the first time around

bbozo:
PS `Celluloid::IO` fires up several threads, maybe all of them don't get joined / terminated by the end of the test, is it possible for something like that to affect the exit code on that place in the test script?

bbozo:
@tarcieri from `HTTP.rb` says that the stack traces are not related to HTTP

sbleon:
@bbozo, were you able to reproduce the mid-test stack trace locally?

Also, if we can find out what the “2” means in `The command "bundle exec
script/test" exited with 2.`, that might give us a further clue.

bbozo:
@sbleon yes, as I said before, the stack trace is present locally, it happens when I `require 'http'`, but the `http` people maintain it has nothing to do with the http gem, but something in the Faraday test environment.

Exit status 2 according to http://tldp.org/LDP/abs/html/exitcodes.html should mean "Misuse of shell builtins" and the stack trace suggest the culprit is in (or immediately after?) the cleanup code:

```
.........................................................................................................................................................................................................................................................................................................................

Finished in 13.473758s, 39.7810 runs/s, 70.3590 assertions/s.

536 runs, 948 assertions, 0 failures, 0 errors, 0 skips
Coverage report generated for Unit Tests to /home/bbozo/dev/faraday/coverage. 1586 / 1835 LOC (86.43%) covered.
[Coveralls] Outside the CI environment, not sending data.
Coverage (86.43%) is below the expected minimum coverage (87.00%).
+ cleanup
+ '[' 2 -ne 0 ']'
+ '[' -n '' ']'
+ kill 31016
+ kill 31019

$ echo $?
2
```

and again: celluloid fires up a number of threads when loaded, is it possible that some of the kills return a different exit code if there are running threads remaining? The culprit is possibly in:

``` sh
wait_for_server() {
  timeout=$(( `timestamp` + $1 ))
  while true; do
    if server_started "$2"; then
      break
    elif [ `timestamp` -gt "$timeout" ]; then
      echo "timed out after $1 seconds" >&2
      return 1
    fi
  done
}
```

I don't know I'm not that good at debugging shell scripts, and `script/test` has a funny flow with some fancier features inside, at least for me

tarcieri:
Last I looked at the stack trace it's blowing up on:

```ruby
require "http/client"
```

...from here: https://github.com/httprb/http/blob/master/lib/http.rb#L8

however beyond that, nothing in the stack trace seemed related (as if something else is defining `http/client.rb` somewhere in the load path, and it's blowing up in that)

iMacTia:
Hello everyone,

I'm closing this as it seems like it has reached a dead end and a new PR (#591) using a more recent version of the gem is in place.

PR #568, opened 2016-03-25
Correct parsing of uri with anchor in Request#url

hbda:
If pass into the method get url with an anchor, the library parses the parameters are not correct.
```ruby
2.3.0 :012 > Faraday.new.get('http://sushi.com?b=2&a=1#qqq').env[:url].to_s
 => "http://sushi.com?a=1%23qqq&b=2" 
```

### DIFF
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -46,14 +46,12 @@ module Faraday
     end
 
     def url(path, params = nil)
-      if path.respond_to? :query
-        if query = path.query
-          path = path.dup
-          path.query = nil
-        end
-      else
-        path, query = path.split('?', 2)
+      path = Utils.URI(path)
+      if query = path.query
+        path = path.dup
+        path.query = nil
       end
+
       self.path = path
       self.params.merge_query query, options.params_encoder
       self.params.update(params) if params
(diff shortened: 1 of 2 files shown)


sbleon:
This looks like a legitimate bug fix. Once we can fix the Ruby < 1.9.3 build issue, I recommend merging this.

lib/faraday/request.rb
```diff
@@ -46,14 +46,12 @@ def headers=(hash)
...
 
     def url(path, params = nil)
-      if path.respond_to? :query
-        if query = path.query
-          path = path.dup
-          path.query = nil
-        end
-      else
-        path, query = path.split('?', 2)
+      path = Utils.URI(path)
```
mislav: Instead of parsing this with URI, I would simply strip out everything after and including `#` in a string. The fragment identifiers are useless in HTTP requests as they don't actually get sent to the server. So, even if they're supplied, they should be ignored.

hbda: @mislav I rewrite request

iMacTia:
Fixes a legitimate bug and now all tests are passing.
Good job, merging 👍

Issue #570, opened 2016-03-30
Travis builds breaking. Need to drop support for Ruby < 1.9.3 or lock rake version.


sbleon:
Some Travis builds which might otherwise pass (e.g. [the build](https://travis-ci.org/lostisland/faraday/builds/118519888) for PR #568) are breaking on Ruby 1.8.7 and 1.9.2 because [rake has dropped support for those versions](https://github.com/ruby/rake/commit/4ab6eba0d77c7e2d449e9292673099e2f10e193b). I think we need to either:
1. Drop support for Ruby < 1.9.3.
or 
2. Lock the rake version in the Gemfile to '< 11.0.0'.

Since rake is required only as a dev dependency, I'd be in favor of option 2. This doesn't seem like a good reason to drop support for those old Ruby versions. @mislav I can make a PR for your preferred option.

Getting this build problem fixed will make it easier to review PRs by eliminating false negatives from Travis.

mislav:
Could you make a PR that updates our Gemfiles to lock down rake to a version `< 11.0.0` please? That should take care of it.

PR #571, opened 2016-03-31
Restrict rake to less than v11, which drops support for older rubies.

sbleon:


sbleon:
This resolves #570

bbozo:
:+1: it makes tests fail on #564

bbozo:
Note, I added

``` ruby
# Newer versions drop support for Ruby < 1.9.3, but we still support them.
if RUBY_VERSION >= '1.9.3'
  gem 'rake'
else
  gem 'rake', '< 11.0.0'
end
```

into #564 to make the tests pass

mislav:
Thanks @sbleon!!

PR #575, opened 2016-04-18
Explain custom middleware registration in the README

danhodge:
This documentation would have saved me a few minutes the first time I needed to create my own custom middleware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,28 @@ later, response. Some keys are:
 :response_headers
 ```
 
+## Registering custom middleware
+
+Custom middleware classes need to be registered with Faraday before they can be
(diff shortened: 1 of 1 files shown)


iMacTia:
Hi @danhodge thanks for pointing this out.
I agree our documentation is lacking some basic explanation. Good news is that we're going to fix this with the new big release, together with a revamped Wiki.

Appreciating your input, I'm going to close this as it will be part of that revision (see #620).

mislav:
@danhodge I don't think middleware should necessarily use the `register_middleware` mechanism. If you define your own custom middleware, you can just use the constant directly:

```rb
conn.use MyMiddleware
```

I think we'll be documenting that approach first and foremost on the wiki.

PR #576, opened 2016-04-22
Fix comparison error at sort in rare case

supercaracal:
I fixed comparison error at sort. This error is reproduced when duplicate string and symbol keys with different values are included in params.

```ruby
conn = Faraday.new(url: 'http://localhost:3000')
conn.post('/products', { :name => 'hogehoge', 'name' => nil })
```

```
ArgumentError at /products
=> comparison of Array with Array failed
```


mislav:
Thanks for the fix, but we don't support mixed symbol and string parameters! I suggest that you switch to only using strings.

PR #585, opened 2016-05-17

ashleym1972:
Allow SNI host header to be specified


mislav:
Thanks, but see https://github.com/lostisland/faraday/pull/470#issuecomment-254296415

PR #586, opened 2016-05-17
Fix typo in method documentation

RobinDaugherty:
According to the `build_request` method, a `Faraday::Request` object is yielded, not a Response object.


mislav:
Good catch; thanks!

PR #588, opened 2016-05-31
Add Zlib::BufError to net/http error handling

joevandyk:
Got this error once:

```
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:363 in finish
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:363 in finish
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:265 in ensure in inflater
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:264 in inflater
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:280 in read_body_0
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:201 in read_body
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1134 in block in get
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1423 in block in transport_request
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http/response.rb:162 in reading_body
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1422 in transport_request
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1384 in request
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1377 in block in request
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:853 in start
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1375 in request
vendor/ruby-2.2.4/lib/ruby/2.2.0/net/http.rb:1133 in get
vendor/bundle/ruby/2.2.0/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:80 in perform_request
vendor/bundle/ruby/2.2.0/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:40 in block in call
vendor/bundle/ruby/2.2.0/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:87 in with_net_http_connection
vendor/bundle/ruby/2.2.0/gems/faraday-0.9.2/lib/faraday/adapter/net_http.rb:32 in call
```

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -21,6 +21,7 @@ module Faraday
         Net::HTTPHeaderSyntaxError,
         Net::ProtocolError,
         SocketError,
+        Zlib::BufError,
         Zlib::GzipFile::Error,
       ]
 


joevandyk:
Is there any chance of this being released in a fix?

iMacTia:
Hi @joevandyk and sorry if it took this long to give you a response.

Ideally we want to keep exceptions in `NET_HTTP_EXCEPTIONS` at the bare minimum (there are already too much in my opinion) and they should be the most common exceptions you could get while using the adapter.

The exception you're adding looks to me like a very edge-case exception, so I would prefer no to include it in the list and close this PR.
However,  if you think this is not just an isolated edge-case and you can provide a clear way to reproduce this issue, I will be happy to reconsider.

Thanks

joevandyk:
I suppose the problem would be reproducible if the remote server's response was corrupted somehow. That's obviously hard to reproduce. :)

We've found when using Ruby and doing millions of http requests a day that there's dozens of different types of exceptions that can be raised when doing a http request, and we are struggling with how to catch the only the network-related exceptions. Any ideas? The only thing I can think of is to rescue all exceptions but that could hide other problems.

iMacTia:
As a general practice, you usually proceed by one of the following:

1. If your application needs to behave differently based on the exceptions, then you need to rescue each of them singularly and write code to manage each of them. If you find new exceptions later, you can always add them with a code update.
2. If your application needs to manage each exception at the same time, but you don't want to loose the information about what happened exactly (helpful for future investigation as well), then you can rescue all exceptions but log the actual issue on a log file to keep trace of it.

In your specific case, you could rescue `Zlib::Error` that according to the [doc](https://ruby-doc.org/stdlib-2.1.1/libdoc/zlib/rdoc/Zlib/Error.html) is the superclass of all Zlib exceptions.

Finally, `Zlib` is a compression library, not a networking one, so the issue in this case was not only related to networking (I would argue that it was not actually due to networking at all). It seems that your server is sending a broken response that is causing Zlib to fail somehow.
I agree this sounds strange as we already have `Zlib::GzipFile::Error` in the list of catches, but I personally don't agree with that decision as well for the same reasons I gave to you.
Unfortunately, it's too late to remove that exception from the list as this would break backward-compatibility, but at least we can avoid adding new ones.

PR #591, opened 2016-06-10
Add adapter for the HTTP.rb gem

jhbabon:
The gem link: https://rubygems.org/gems/http

You can initialize the adapter with any of the options that the gem allows:

```ruby
Faraday.new do |faraday|
  faraday.adapter :http, :keep_alive_timeout => 10
end
```

I know there is a PR trying to do something similar (https://github.com/lostisland/faraday/pull/564) but is kind of stuck and is using and old version of the `http` gem, so I decided to try my adapter. Feel free to close this PR if you want to push the other one and sorry if is bothering someone.

About the stack trace in the specs, I opened a PR for the `http` gem as well to see if it can be fixed. I think there is nothing we can do in the Faraday side. This is the PR: https://github.com/httprb/http/pull/353

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,7 @@ group :test do
   gem 'simplecov'
   gem 'sinatra', '~> 1.3'
   gem 'typhoeus', '~> 0.3.3', :platforms => [:ruby_18, :ruby_19, :ruby_20, :ruby_21]
+  gem 'http', '~> 2.0'
 
   # Below are dependencies of the gems we actually care about that have
   # dropped support for older Rubies. Because they are not first-level
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ group :test do
   gem 'sinatra', '~> 1.3'
   gem 'typhoeus', '~> 0.3.3', :platforms => [:ruby_18, :ruby_19, :ruby_20, :ruby_21]
   gem 'http', '~> 2.0'
+  gem 'cookiejar', '= 0.3', :platforms => [:ruby_18, :ruby_19]
 
   # Below are dependencies of the gems we actually care about that have
   # dropped support for older Rubies. Because they are not first-level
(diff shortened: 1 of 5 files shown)


iMacTia:
@jhbabon I can see the PR on http was merged, can you update your branch to make tests passing?
Also, for completeness, here is the requirements for new adapters.
Please double-check you followed all of them:

> We will accept adapters that:
- support SSL & streaming;
- are proven and may have better performance than existing ones; or
- if they have features not present in included adapters.

@mislav, @technoweenie, this would be a new adapter so is probably better if you review it.

technoweenie:
I think this would work better as a `faraday-http` gem. We can provide access to a repo on the lostisland account for that. Ideally, Faraday maintainers shouldn't have to maintain compatibility with all the various HTTP ruby gems. This also gives adapter contributors more freedom over their own adapters.

I realize that Faraday depends on plenty of gems today. Some plans in a completely rewritten Faraday 1.0 and separate Hurley gem included supporting _only_ the core `net/http` adapter.

iMacTia:
Hi @jhbabon I just want to update you on the situation.
We're currently in the process of drawing Faraday's next big release 1.0 (#620).
One of the things we're discussing is how to deal with new adapters proposals as the more we support inside Faraday, the more work we have to do in order to support new features in the future, so we need to plan this carefully.

I appreciate your patience. In the meantime, you can update your branch with the latest `ruby-http` gem version including your fix.

jhbabon:
Hi @iMacTia!

Thanks for the all the explanations. I think is better to close this PR because right now I can't confirm that this adapter has the same or more of the features other adapters have, specially the streaming part. Also, moving this to a separate gem once the version `1.0` of Faraday is released makes total sense. I'll wait until then and if I have time I'll try to take a look at this again.

Thanks for your patience.

iMacTia:
Thank you for your understanding!
I really hope this will see the light in the future as there are some people requesting it.

PR #594, opened 2016-07-08
Update net_http.rb

prokizzle:
Add recognition of uncaught `Errno::EPIPE: Broken pipe` error to `NET_HTTP_EXCEPTIONS`

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -17,6 +17,7 @@ module Faraday
         Errno::EHOSTUNREACH,
         Errno::EINVAL,
         Errno::ENETUNREACH,
+        Errno::EPIPE,
         Net::HTTPBadResponse,
         Net::HTTPHeaderSyntaxError,
         Net::ProtocolError,


iMacTia:
Hi @prokizzle, thanks for your PR!

Makes sense to me adding this error to the list, but tests are failing due to gems versions incompatibilities. Would you mind updating your branch with the latest master branch version to fix this before merging?

mislav:
@prokizzle Thanks for the contribution! Can you share more about when does EPIPE happen during the request cycle?

prokizzle:
As far as I know @mislav it's one of those errors that pops up when there's an unexpected network problem, i.e., an incomplete response, or a network drop mid request. It's pretty difficult to reproduce. I read this [PR588]( https://github.com/lostisland/faraday/pull/588) which seems kind of similar, and while I agree it's annoying to have to add a bunch of arbitrary exceptions to faraday, the randomness of the error makes me think it should be handled by an umbrella error in faraday, rather than having to rescue it in code.

iMacTia:
EPIPE is strictly related to SIGPIPE as explained [here](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Operation-Error-Signals).

It also states:

> Another cause of SIGPIPE is when you try to output to a socket that isn’t connected. 

So what's probably happening in this case is that the socket was closed for some reason before the request was sent.
Without trying to understand the issue behind this, I think this error can be network-related enough to be included in the `NET_HTTP_EXCEPTIONS` list.

If you can merge latest changes from master and make tests pass again, I'll merge this in.

prokizzle:
Okay, I think it's all set.

iMacTia:
Amazing, thanks @prokizzle 👍
Merging

PR #600, opened 2016-08-24
Fix Ruby 1.9.3 tests on Travis. Drop support for Ruby 1.8.7.

sbleon:


sbleon:
@iMacTia I'd like to get this merged ASAP, since all other PRs are failing to build because of the test issues I've fixed herein.

@technoweenie @mislav does anyone have major concerns about dropping 1.8.7 from our build matrix? It's becoming increasingly difficult to support due to so many gems dropping support for it.

iMacTia:
I personally believe this is ok, we can always mark the next release as 0.10 and explain that support have been removed for some VERY OLD ruby versions. If people don't want to update their ruby, they simply need to force the proper version on the Gemfile:

```ruby
gem 'faraday', '~> 0.9'
```

For this reason, it will be extremely important to increase the version number in the next release. Please do not forget about it!

iMacTia:
I would wait for a day the chance to the other lostiland member to approve, than you can proceed with the merge yourself without waiting for me :)

=== OT ===
I'll be incresingly unresponsive in the next days so please proceed without me if you don't see me replying!
=========

technoweenie:
I'm :+1: on it for a big revision (v0.10.0, v1.0, whatever). v0.9.x will always be available for Ruby 1.8.7 users. Any major bug fixes could be backported if really necessary. 

Honestly, I'm so out of the ruby game these days, I have no idea what a reasonable ruby version is. Do people still run 1.9.x too?

sbleon:
Thanks, Rick! I’m not sure how much of the community is stuck on Ruby 1.8.7
or 1.9.x. My thinking is to keep supporting old versions until it becomes
painful, and it just became painful to support 1.8.7.

I was also just thinking it was probably time for a 1.0 release of faraday.
Regarding releases, do you want to handle the release to rubygems, or would
you feel comfortable sharing that responsibility with @iMacTia and myself?
If the latter, I’m on rubygems at leon@singlebrook.com. So, `gem owner
faraday -a leon@singlebrook.com` would add me as an owner.

On Wed, Aug 24, 2016 at 4:32 PM, risk danger olson <notifications@github.com

iMacTia:
I would like to share this article to give an overall idea of the ruby situation: https://semaphoreci.com/blog/2015/12/01/ruby-version-usage-in-commercial-projects-2015-edition.html

Considerations:

* this chart is from 2015, in the article they state that ruby 2.x usage was 79% in 2015 and 51% in 2014, so it looks like a fast growth. This year we're probably over 85-90%.
* majority (if not all) projects **in development** today are based on ruby 2.x, so affected projects by dropping ruby <2.0 support are those in production that are rarely kept up-to-date.
* Most of the famous gems we depend on have already dropped the support

So IMHO, we're not causing any issue to anyone.
The only addition I would do is to add to the gemspec the ruby dependency as described here: http://stackoverflow.com/questions/5004492/how-do-you-specify-a-minimum-ruby-version-in-a-gemspec 
This way, bundler won't allow people with ruby < 2.0 to upgrade.
the other advantage is that we can thus completely remove every support for ruby < 2.0, included forced gem versions and all tests, making version 0.10 more maintainable and secure.


My suggestion about version 1.0 is to proceed fixing/merging issues/PRs using small updates (0.10 as we drop old ruby support, then 0.10.1, 0.10.2 and so on...) and leave the v1.0 as a "prize" when
we'll close most of them 😄 
Ideally we want that to be a major, as stabe as possible release

sbleon:
I got rid of most of the Ruby version checks. I don't think dropping support for 1.9.3 gets us much added benefit, so I'm in favor of still supporting it for now. 

I spec'ed support for the Ruby version in the gemspec, so I think this is good to merge now.

sbleon:
@iMacTia your idea about 0.10 vs 1.0 sounds good to me.

iMacTia:
I'm glad you agree and that it makes sense :)!
Regarding ruby 1.9.3, I thought tests were failing like for 1.8.7 due to gems not supporting it anymore.
If we can update our dependencies and tests are passing with 1.9.3, then happy to keep supporting it!

iMacTia:
Oh I see! Faraday works with ruby 1.9.3 if you force those 2 gems in the gemfile!
Unfortunately, AFAIK, gemfile is only used by tests and is not taken into consideration the gem is used :( 
This means that people will need to manually instruct their rails peojects to use the correct version (if using ruby 1.9.3).
I would suggest we make this clear in the Readme to avoid people wasting time and opening issues, if not there already

sbleon:
The README has a section about supported ruby versions, and I updated it to remove 1.8.7.

All of the restrictions in the Gemfile are for test dependencies, so users of the library don't have to worry about version locking. But thanks for thinking of them anyway!

Any other concerns before we merge this?

iMacTia:
Nope, merging now :)!
Thanks for the fix, next time a PR is failing we can ask to merge changes from master and see the green tick again!

sbleon:
Thanks, Mattia!

Rick, what do you think about getting the recent changes released to
rubygems?

mislav:
👍 for droppping 1.8 support in 0.10. Thanks @sbleon!

PR #601, opened 2016-08-26
Closes #599 Update run_request docs

### LINKED ISSUE #599: Inconsistent documentation for `run_request`
[The rubydoc for `Connection#run_request` say the signature is:](http://www.rubydoc.info/gems/faraday/0.9.0/Faraday/Connection#run_request-instance_method)

* method => Symbol 
* url => String|URI
* **body => String**
* headers => Hash

```ruby
    # Builds and runs the Faraday::Request.
    #
    # method  - The Symbol HTTP method.
    # url     - The String or URI to access.
    # body    - The String body
    # headers - Hash of unencoded HTTP header key/value pairs.
    #
    # Returns a Faraday::Response.
```

However, the README examples demonstrate should be a hash:
https://github.com/lostisland/faraday#usage

I am using strings like the docs demonstrate, but it would be helpful to know which the correct/supported way.

Thanks for your time & your work on Faraday!

djds23:


iMacTia:
Thank you @djds23!
Failing tests are ruby < 2.0 as usual so that's not an issue (Also because only a comment line was changed).
Merging this and closing #599

iMacTia:
Thanks to you for your contribution!
I'm sure it will save some time in the future to new developers :)!

PR #603, opened 2016-08-29
Add optional content-length on parts in multipart requests

dblommesteijn:
This will fix an issue I had with the middleware `Content-Length` layer of Faraday. When connecting to a HTTP server that was not accepting `Content-Length` of all chunked payload.

Some exotic server required the `Content-Length` of the part pushed with the multipart request, and not the entire length of the content including newline characters an other parts. Moreover, the `Content-Length` needed to be stripped of the trailing `\r\n` to work correctly. 

The fix is enabled by adding a parameter to the `Faraday::UploadIO` object initialiser, a snipped you can find below. By assigning the key (index) of the Faraday::UploadIO to the `:use_part_content_length` parameter, the `Content-Length` of the part configured is used.

```ruby
payload = {:c => Faraday::UploadIO.new(__FILE__, 'text/x-ruby', 'request_middleware_test.rb', use_part_content_length: :c)}
```

A test has been added to the `test/request_middleware_test.rb` file: `test_multipart_with_content_length_limit`.

### DIFF
--- a/lib/faraday/request/multipart.rb
+++ b/lib/faraday/request/multipart.rb
@@ -35,12 +35,17 @@ module Faraday
     def create_multipart(env, params)
       boundary = env.request.boundary
       parts = process_params(params) do |key, value|
-        Faraday::Parts::Part.new(boundary, key, value)
+        part = Faraday::Parts::Part.new(boundary, key, value)
+        # use content-length of given part
+        if value.is_a?(UploadIO) && value.opts[:use_part_content_length] == key
+          env.request_headers[Faraday::Env::ContentLength] = (part.length - 2).to_s
+        end
+        part
       end
       parts << Faraday::Parts::EpiloguePart.new(boundary)
-
       body = Faraday::CompositeReadIO.new(parts)
-      env.request_headers[Faraday::Env::ContentLength] = body.length.to_s
+      # set content-length when not set for the uploaded part
+      env.request_headers[Faraday::Env::ContentLength] ||= body.length.to_s
       return body
     end
 
--- a/test/request_middleware_test.rb
+++ b/test/request_middleware_test.rb
@@ -117,6 +117,25 @@ class RequestMiddlewareTest < Faraday::TestCase
     assert_equal [], regexes
   end
 
+  def test_multipart_with_content_length_limit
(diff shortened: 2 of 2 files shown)


dblommesteijn:
Any news guys?

iMacTia:
Hi @dblommesteijn, thanks for your PR and sorry for the delay in the response!
I understand the issue you're experiencing, but there are a couple of reasons on why I'm not really happy with this to be added to Faraday:

1. From what I understand, this is an issue specific of this server which is not following the standard way to check the Content-Length on this type of payloads. I looked into the issues backlog and I couldn't find other people with the same issue either. I'm not really a fan of having an additional check on the base library (executed everytime) just to support this out-of-standard server. I can accept to have this merged if it turns out the server is indeed using a standard (just different from the most famous one), if this standard is not too old and the server have a good distribution (e.g. some version of IIS).

2. Even though point 1 gets passed, I still don't like the solution provided as you're storing into the Faraday::UploadIO object an option (`use_part_content_length`) that will then be used by the `Multipart` middleware. I would rather prefer a solution where this option is passed directly to the middleware.


My personal point of view is that this exceptional case should be managed with a small custom middleware in your project that change the `Content-Length` header before sending the request and set it to the correct value accepted by the server.

dblommesteijn:
@iMacTia Thanks for your message, when I would refactor it into the middleware would this patch be accepted (if done correctly)?

iMacTia:
I still think this middleware would not be broadly used and thus I don't see the need to include it in the `faraday` or `faraday_middleware` gems. My point was that you can write this custom middleware and use it in your project only so that you don't need any change on Faraday's code. Reasons are explained in point 1. If there were more people experiencing this issue, than a PR for the middleware would make sense in `faraday_middleware` gem.

iMacTia:
Closing for the time being.
Happy to reopen if this gets a "broader audience" as stated in our [CONTRIBUTING](https://github.com/lostisland/faraday/blob/master/CONTRIBUTING.md) guidelines.

dblommesteijn:
@iMacTia thanks for your feedback i'll solve it for us in a more sustainable way via the middleware.

PR #522, opened 2015-09-12
Streaming requests for Net::HTTP

grosser:
Based on work by @bmaddy, @james-lawrence and @ANorwell
https://github.com/backupify/faraday/tree/streaming-0.9.0
https://github.com/bmaddy/faraday/tree/streaming

rebased https://github.com/lostisland/faraday/pull/461

PR #461, opened 2015-01-26
Streaming requests for Net::HTTP

jcoyne:
Based on work by @bmaddy, @james-lawrence and @ANorwell
https://github.com/backupify/faraday/tree/streaming-0.9.0
https://github.com/bmaddy/faraday/tree/streaming

PR #604, opened 2016-08-30
Streaming requests for Net::HTTP

nightscape:
This PR is a rebased version of https://github.com/lostisland/faraday/pull/461 and https://github.com/lostisland/faraday/pull/522 including an additional test and fix for empty initial chunks mentioned by @grosser in https://github.com/lostisland/faraday/pull/461#issuecomment-139620539.
Tests are all green on my machine, so it would be great if someone (maybe @mislav or @iMacTia) could have a look at this :smiley:
The other two PRs can be closed in favor of this one.

### DIFF
--- a/lib/faraday/adapter/em_synchrony.rb
+++ b/lib/faraday/adapter/em_synchrony.rb
@@ -26,6 +26,11 @@ module Faraday
         # Queue requests for parallel execution.
         if env[:parallel_manager]
           env[:parallel_manager].add(request, http_method, request_config(env)) do |resp|
+            if (req = env[:request]).stream_response?
+              warn "Streaming downloads for #{self.class.name} are not yet implemented."
+              req.on_data.call(resp.response, resp.response.bytesize)
+            end
+
             save_response(env, resp.response_header.status, resp.response) do |resp_headers|
               resp.response_header.each do |name, value|
                 resp_headers[name.to_sym] = value
@@ -54,6 +59,10 @@ module Faraday
 
           raise client.error if client.error
 
+          if env[:request].stream_response?
+            warn "Streaming downloads for #{self.class.name} are not yet implemented."
+            env[:request].on_data.call(client.response, client.response.bytesize)
+          end
           status = client.response_header.status
           reason = client.response_header.http_reason
           save_response(env, status, client.response, nil, reason) do |resp_headers|
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -73,9 +73,38 @@ module Faraday
       end
 
       def perform_request(http, env)
+        if env[:request].stream_response?
+          size = 0
+          http_response = perform_request_with_wrapped_block(http, env) do |chunk|
+            size += chunk.bytesize
+            env[:request].on_data.call(chunk, size)
+          end
+          # Net::HTTP returns something, but it's not meaningful according to the docs.
+          http_response.body = nil
+          http_response
+        else
+          http_response = perform_request_with_wrapped_block(http, env)
+        end
+      end
+
+      def perform_request_with_wrapped_block(http, env, &block)
         if :get == env[:method] and !env[:body]
           # prefer `get` to `request` because the former handles gzip (ruby 1.9)
-          http.get env[:url].request_uri, env[:request_headers]
+          request_via_get_method(http, env, &block)
+        else
+          request_via_request_method(http, env, &block)
+        end
+      end
+
+      def request_via_get_method(http, env, &block)
+        http.get env[:url].request_uri, env[:request_headers], &block
+      end
+
+      def request_via_request_method(http, env, &block)
+        if block_given?
+          http.request create_request(env) do |response|
+            response.read_body(&block)
+          end
         else
           http.request create_request(env)
         end
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -75,10 +75,15 @@ module Faraday
       def perform_request(http, env)
         if env[:request].stream_response?
           size = 0
+          yielded = false
           http_response = perform_request_with_wrapped_block(http, env) do |chunk|
-            size += chunk.bytesize
-            env[:request].on_data.call(chunk, size)
+            if chunk.bytesize > 0 || size > 0
+              yielded = true
+              size += chunk.bytesize
+              env[:request].on_data.call(chunk, size)
+            end
           end
+          env[:request].on_data.call("", 0) unless yielded
           # Net::HTTP returns something, but it's not meaningful according to the docs.
           http_response.body = nil
           http_response
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -96,6 +96,16 @@ module Adapters
 
         assert_equal "", response.body
       end
+
+      def test_GET_streaming_empty_response
+        _, streamed = streaming_request(create_connection, :get, 'empty_stream')
+        assert_equal [["", 0]], streamed
+      end
+
+      def test_non_GET_streaming_empty_response
+        _, streamed = streaming_request(create_connection, :post, 'empty_stream')
+        assert_equal [["", 0]], streamed
+      end
     end
 
     module NonStreaming
@@ -345,7 +355,7 @@ module Adapters
 
         # Check that the total size of the chunks (via the last size returned)
         # is the same size as the expected_response
-        assert_equal sizes.last, expected_response.size
+        assert_equal sizes.last, expected_response.bytesize
 
         start_index = 0
         expected_chunks = []
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -1,5 +1,6 @@
 require 'forwardable'
 require File.expand_path("../../helper", __FILE__)
+require File.expand_path("../../shared", __FILE__)
 Faraday.require_lib 'autoload'
 
 module Adapters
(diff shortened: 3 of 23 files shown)


lib/faraday/adapter/em_synchrony.rb
```diff
@@ -26,6 +26,11 @@ def call(env)
         # Queue requests for parallel execution.
         if env[:parallel_manager]
           env[:parallel_manager].add(request, http_method, request_config(env)) do |resp|
+            if (req = env[:request]).stream_response?
```
grosser: extract into a method to avoid duplication ?

lib/faraday/adapter/net_http.rb
```diff
@@ -75,9 +75,45 @@ def create_request(env)
       end
 
       def perform_request(http, env)
+        if env[:request].stream_response?
+          size = 0
+          never_yielded = true
```
grosser: reverse the logic to simplify ? `yielded = false ... yielded = true`

lib/faraday/adapter/net_http.rb
```diff
@@ -75,9 +75,45 @@ def create_request(env)
...
+      end
+
+      def request_via_get_method(http, env, &block)
+        http.get env[:url].request_uri, env[:request_headers], &block
+      end
+
+      def request_via_request_method(http, env, &block)
+        if block_given?
+          http.request create_request(env) do |response|
+            response.read_body do |chunk|
```
grosser: can simplify to `response.read_body(&block)`

nightscape:
Hmm, I'm rather clueless why Travis is failing. Tests are all green, but there are some cryptic `warning: previous definition of xxxxx was here` syntax warnings here https://travis-ci.org/lostisland/faraday/jobs/156358909#L216 and a lot of
```
ERROR WEBrick::HTTPStatus::ProxyAuthenticationRequired
```
at the end of the build.

lib/faraday/adapter/em_synchrony.rb
nightscape: I'm running in circles trying to find a nice way to extract a method here.
Do you have something in mind?

lib/faraday/adapter/net_http.rb
nightscape: :+1:

lib/faraday/adapter/net_http.rb
nightscape: :+1:

lib/faraday/adapter/em_synchrony.rb
grosser: nope ... could put it on the request or make it static ...

nightscape:
Giving this a little more thought, I think a much nicer API would be a `Faraday::Response#chunks` method which returns an `Enumerator`. That way, you could achieve the same as with a callback using `each` on the Enumerator, but you could also `map`, `select`, ... on it and pass the `Enumerator` around like any object.
Trying this now...

lib/faraday/adapter/em_synchrony.rb
iMacTia: I'm not 100% sure of it as well, but here are my 2 cents.
I personally like having a `perform_request` method on most of the adapters. I would suggest having a `perform_stream_request` method to manage stream logics and having the `call` method calling the proper "perform" based on a check with `stream_response?`.

The common logic will therefore be something like (extremely simplified):

```ruby
def call(env)
  if req.stream_response?
    perform_stream_request
  else
    perform_request
  end
end
```

At this point, `perform_stream_request` can be implemented in the common `Faraday::Adapter` triggering the warning message and override this method only in the adapters where
you want to provide a proper implementation.
I really hope this is clear, but please do let me know if that's not the case!

Good work btw 👍!

iMacTia:
@nightscape first of all thanks for this important contribution!
I've never had the need to use Faraday to read from a streaming so I'm not sure I will be able to contribute on this, but surely will come useful for future users :)

I've already commented on the repetition of the code with the warning with my suggestion to limit that issue and provide a common interface on all adapters (checking `stream_response?` on the `call` method and then delegating to the proper method).

Regarding the failing tests, I'll need a computer to see what's wrong and I won't have one in the next couple of weeks. @sbleon I suppose this is not the same issue you fixed last day?

Finally, the Enumerator seems also a good idea, but there's still something I don't really get...
Since we're talking about a stream, this means that the connection will be kept open and new chunks will be added to the response over time, so the response could potentially never end.
We should design this so that people can easily use it in most situations.

Just to make this easier to understand, can you show a couple of example on how you would use this streaming functionality?

test/live_server.rb
```diff
@@ -1,4 +1,5 @@
 require 'sinatra/base'
+load 'test/helper.rb'
```
sbleon: I'm guessing the ruby syntax warnings about re-declarations are due to this line, as some of the things it's complaining about are defined in `test/helper.rb`. I'm pretty sure this test helper script shouldn't try to include the setup for the test suite itself.

I'm not sure why you have this `load` here at all. Is it because you want to use `Faraday::TestCase.big_string`? If so, you can just copy that method over to `Faraday::LiveServer`. I think it's too small to worry about the duplication.

iMacTia: Good point, if it is complaining is probably because that file gets loaded twice. Maybe changing `load` with `require` will fix the issue at requiring the same file twice only loads it once

sbleon: That might do it, but I still think you should bring in only what you need.
Maybe `big_string` could be moved into a module that can be used by both
test/helper and live_server?

iMacTia: Makes sense to me 😄 
Any DRY solution is welcome!

lib/faraday/adapter/em_synchrony.rb
nightscape: I still can't find a nice way to make this work.
If the `call` method you sketched above was located in `Faraday::Adapter` then sub-classes would need to have the same signature for `perform_request` which doesn't seem to be the case.
The same applies to `perform_stream_request` which - if I understood correctly - should delegate to `perform_request` unless overridden in a sub-class.
The only way I can currently think of to make this work is to copy the above code snippet into the `Adapter` sub-classes and have individual implementations of `perform_streaming_request` which would have a similar signature as `perform_request`.
But maybe I'm just overlooking something.

Would you be OK with merging even in the current unDRY state of the code?
Maybe someone comes up with a nicer solution later, I'm currently not seeing a proper way out.

nightscape:
I've pondered the Enumerator idea again, and I think it has one fundamental flaw: Control over the reading the response is given to the user of the Enumerator. Endless responses would not be so much of an issue (that's one of the typical use cases of Enumerators), but if a client just called `first` on the Enumerator and kept the Enumerator instance around it might be that the response and the connection remain open indefinitely (or until some timeout cuts the connection).
I might be wrong on this, but maybe it's better to continue with the original callback style which should not have this issue.

I've incorporated @sbleon's proposal of extracting the `big_string` method into a shared module, so except for the duplication of warning message creation (for which I can't find a good solution right now) this would be good to merge from my side.

What do you think, @iMacTia and @sbleon?

lib/faraday/adapter/em_synchrony.rb
iMacTia: I suppose that's ok if we make tests work so we don't break everything later refactoring the code 😁

iMacTia:
How is the documentation? Do we have a good explanation on how to use te callback to manage the streaming request?

Apologies again for not checking myself but doing everything with the mobile is kind of painful 😥

nightscape:
I just added a small example to the README.
Is there another place where documentation should be added?

lib/faraday/adapter/em_synchrony.rb
nightscape: Travis tests are working now :+1:

iMacTia:
Not as far as I know, README should be fine 😁
I'm ok with merging at this point. if @sbleon agrees as well, he can proceed 👍

akshaylive:
Good PR! Looking forward to this one! 👍

iMacTia:
@nightscape wait a sec, are warning still there even though you followed @sbleon advice? Did you remove the `load test/helper.rb` after extracting the `big_string` method?

nightscape:
Hi @iMacTia, the warnings due to `big_string` are solved.
What I was referring to above is that the code that generates warnings for adapters that don't support streaming is still duplicated because I haven't found a nice way to DRY it.
So from my side this is good to merge and maybe later someone comes up with a nice idea of how to DRY or improve the code.

iMacTia:
I see, sorry @nightscape, reading again after some days I confused the build warnings with those ones 😅
So again, all good for me if @sbleon agreed as well. If he's still busy and can't merge, don't worry, I will merge early next as soon as I have a proper computer to check it 😉

lib/faraday/adapter/net_http.rb
```diff
@@ -75,9 +75,43 @@ def create_request(env)
...
 
       def perform_request(http, env)
+        if env[:request].stream_response?
+          size = 0
+          yielded = false
+          http_response = perform_request_with_wrapped_block(http, env) do |chunk|
+            if chunk.size > 0 || size > 0
+              yielded = true
+              size += chunk.bytesize
+              env[:request].on_data.call(chunk, size)
```
nurse: In `README.md` it says the call's 2nd argument is chunk_size, but `size` here is total size.
Which does you intend?

lib/faraday/adapter/net_http.rb
```diff
@@ -75,9 +75,43 @@ def create_request(env)
       end
 
       def perform_request(http, env)
+        if env[:request].stream_response?
+          size = 0
+          yielded = false
+          http_response = perform_request_with_wrapped_block(http, env) do |chunk|
+            if chunk.size > 0 || size > 0
```
nurse: `chunk.size` should be `chunk.bytesize`.

test/adapters/integration.rb
```diff
@@ -257,6 +333,44 @@ def create_connection(options = {})
...
+          :prefix => '',
+          :streaming? => true
+        }.merge(options)
+        expected_response = opts[:prefix] + Faraday::Shared.big_string
+
+        chunks, sizes = streamed.transpose
+
+        # Check that the total size of the chunks (via the last size returned)
+        # is the same size as the expected_response
+        assert_equal sizes.last, expected_response.size
```
nurse: `expected_response.size` should be `expected_response.bytesize`.

iMacTia:
@nightscape: @nurse highlighted a couple of things that are not completely clear, do you think you can have a look at them before we proceed? Thanks

lib/faraday/adapter/net_http.rb
nightscape: Good catch :) Which one would be preferred? Bytes or Unicode characters? I don't care either way.

iMacTia: I think what the documentation states is probably the best option in this case:

> Set a callback which will receive tuples of chunk Strings
and the sum of characters received so far

as the chunk_size can be easily obtained calling `chunk.bytesize`, making it quite useless as parameter.

nurse: I think it should be total byte size too because people can get chunk bytesize by `chunk.bytesize`.

nightscape:
@nurse: I added two fixup commits. If those changes are what you meant and Travis gives us green light, I'll squash them into their corresponding commits.
I also rebased upon the latest master version.

nurse:
@nightscape Looks fine! thanks!!

nightscape:
I squashed the fixups, so Git history is clean now.
From my side the PR is good to go now :+1:

iMacTia:
@nightscape apologies I still didn't have the time to look properly into this.
Will try to do it in the next days.
@mislav, @technoweenie this seems quite an important change, it would be nice if you could have a look as well.

nightscape:
Bumping this, did any of you have the time to look at this? @iMacTia, @mislav, @technoweenie

iMacTia:
Hi @nightscape sorry I labeled the task but forgot to leave you a comment (I was sure I did, but Github doesn't think so).

We have discussed internally and decided that streaming will eventually be part of the next big release (1.0) rather than next minor ones.
This is because we think streaming will be quite a big feature and we would prefer to give it enough tests (beta and RC releases will help mitigate any issue).
The other reason is that I really like the simplicity of your PR, but it only address the Net::HTTP adapter. Good news is we're considering the idea of focusing our attention into that adapter in v1.0.

I understand this will be annoying as it means that in order to use this feature for the time being you'll need to use your branch, but I hope you understand we want to take this with the due caution.

I'm preparing a plan for the release (you can already have a look at #620) and I'm definitely considering streaming as a feature for it. This will probably means that your PR will be redirected to a new "v1.0" branch in the next days.

I'll try to keep you and the community up-to-date on it, sorry I can't give you deadlines yet.

nightscape:
Hi @iMacTia,

that sounds reasonable. We'll be using our own fork or maybe just wait for 1.0 :+1:

iMacTia:
@nightscape Coming back to this as part of the v1.0 works.
Have you been using this for the past months? Is there any change you would like to add before this gets merged?
Would you be able to rebase against v1.0 branch? (please note I've changed the PR base branch already

nightscape:
Hi @iMacTia,

because of shifting priorities and a partial move to Scala we haven't been using the code in the past months.
I can try to rebase the PR against 1.0, let's see if I still find my way around :wink:

nightscape:
Hi @iMacTia,
I just rebased the code (there was only one minor conflict in option handling).
Travis still says everything's good :smiley:

iMacTia:
Hi @nightscape, thank you very much!
I really appreciate you finding some time to come back to this!
It was an highly requested feature so I'm happy to make this the first important merge of v1.0 👍
We'll probably need to do some refinement, but this looks like a great start!
Thanks

nightscape:
Thank you for looking into it and merging!

grosser:
amazing, it finally happened :D

iMacTia:
I never said this wouldn't happen :wink:
I'm just sorry this took so long.
@grosser are you still interested in using this? It would be nice to have some real-world testing of these changes. Maybe not in production but on a side project :smile:

grosser:
no idea where/why I needed that :D

PR #607, opened 2016-09-14
Updates docs for custom params encoding

7sun:


iMacTia:
Thanks!

Hopefully will save some time in the future to other devs!

Merging

PR #610, opened 2016-09-28
Object#tap no longer required after dropping 1.8.7

swalkinshaw:
Since Faraday only supports Ruby 1.9+ (from https://github.com/lostisland/faraday/pull/600), `Object#tap` is a core method and
this monkey patch is no longer needed.

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -239,25 +239,3 @@ module Faraday
     require_lib 'autoload'
   end
 end
-
-# not pulling in active-support JUST for this method.  And I love this method.
-class Object
-  # The primary purpose of this method is to "tap into" a method chain,
-  # in order to perform operations on intermediate results within the chain.
-  #
-  # Examples
-  #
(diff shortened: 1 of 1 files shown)


iMacTia:
Good spot @swalkinshaw, always happy to get rid of monkey-patches!
Unfortunately tests are failing for 1.9.3 but not for your fix.
I'll need to force some gem version on the master to make them pass again.

iMacTia:
@swalkinshaw just pushed in master the fix for ruby1.9.3 tests.
Can you please pull these changes into your branch?

swalkinshaw:
@iMacTia updated 👌

iMacTia:
All green, that's great! Merging straight away.
Thanks again for the work!

mislav:
Thanks for the contribution! I would also be happy if we stopped using `tap` altogether in the codebase. It's syntax sugar that is seldom useful.

PR #613, opened 2016-10-05
Optionally log headers as they may contain sensitive information

timlinquist:
If youre using basic auth for instance the credentials will be included
in the logs. Headers are also used for passing tokens regularly.

Ideally this would be a whitelist. Suggesting this simple change now so
that folks can take advantage of it.

https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -4,7 +4,7 @@ module Faraday
   class Response::Logger < Response::Middleware
     extend Forwardable
 
-    DEFAULT_OPTIONS = { :bodies => false }
+    DEFAULT_OPTIONS = { :bodies => false, :headers => false }
 
     def initialize(app, logger = nil, options = {})
       super(app)
@@ -33,6 +33,7 @@ module Faraday
     private
 
     def dump_headers(headers)
+      return unless @options[:headers]
       headers.map { |k, v| "#{k}: #{v.inspect}" }.join("\n")
     end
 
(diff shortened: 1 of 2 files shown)


iMacTia:
Hi @timlinquist, thanks for the PR!
I like the idea and understand your needs, but I have 2 changes requests for you:

1. I would prefer the behaviour to be the same as the `bodies` option. If you check the code, you will see there's a `log_body?` method accepting a `type` param and this method is called in both `call` and `on_complete` methods with different types (:request and :response). This allows you to choose if logging only the request body, only the response one, both of them or none of them. It would be great to have the same behaviour for headers as well rather than a binary "yes" or "no".

2. The default value should be true, not false. This is because the current behaviour is equivalent to `true` (in fact headers are logged for both request and response), and we need to guarantee backward-compatibility.

If you change these two things I'll be happy to merge your PR as it's actually adding a new useful feature to the logger middleware 👍

timlinquist:
@iMacTia Ok great. I'll work up the changes. I was hoping to sneak in the breaking change since it's insecure if users overlook their sensitive headers being logged. I was surprised to see it as default to logging everything.

I'll whip these up and @ you when I've made the fixed. Thanks!

iMacTia:
@timlinquist sorry for interrupting you, but I found this PR while going through them: #530 
This actually implements exactly what you need, the way I asked.
There's also test coverage and so on.
I'll try to ping the author as the branch needs to be updated so please hang on for the time being

timlinquist:
@iMacTia Ah great! I'll look for that to land. Thanks for all your hard work!

timlinquist:
I'll re-open if the other pr doesn't land

Issue #615, opened 2016-10-06
Net::HTTP::Persistent removed Ruby 1.9.3 support


mikefogg:
Faraday is using the [net-http-persistent](https://github.com/drbrain/net-http-persistent) gem, which looks like it has just [recently removed support for Ruby 1.x](https://github.com/drbrain/net-http-persistent/commit/e62e2bd6d54b1d5504aeb975d28e4315b633399f) in it's latest 3.0 version.

This causes faraday not to install on Ruby 1.9.3 at all:

```
Installing net-http-persistent 3.0.0
Gem::InstallError: net-http-persistent requires Ruby version ~> 2.1.
```

Because Faraday supports Ruby 1.9.3+, we should update the gemfile to not pull in version 3.0 of the net-http-persistent gem.

PR #616, opened 2016-10-06
Fix gemfile to pull in the net-http-persistent gem for Ruby 1.9.3+

mikefogg:
This allows the gem to work again on Ruby 1.9.3 to fix https://github.com/lostisland/faraday/issues/615.


iMacTia:
Looks like `net-http-persistent` 3.0.0 has some braking changes for Faraday.
This is the only fix we can do for the time being, thanks for the PR.

PR #619, opened 2016-10-14
Adapter support for Net::HTTP::Persistent v3.0.0

olleolleolle:
This PR attempts to support [Net::HTTP::Persistent](https://github.com/drbrain/net-http-persistent) 3.x or 2.x with its existing Faraday Adapter.

## Changes introduced

- For Net::HTTP::Persistent 3.x, there were changed APIs which needed changes in the existing Faraday Adapter
  - a constructor had gotten **keyword** arguments instead of **positional** ones
  - a cleaning-up method after SSL connections had been renamed
- Introduce the use of `ruby RUBY_VERSION` Ruby version hinting for Bundler in the Gemfile. 
  - The Gemfile is used during development and during test runs in CI.
  - with this, it was possible to **avoid pinning** to existing versions, and have the latest versions available supported by the _currently running platform_

## TODO

- [x] use `parameters` instead of `Net::HTTP::Persistent::VERSION` to find out which version is running

## Out of scope for this PR:

- tests for all 2.x/3.x of Net::HTTP::Persistent on all Ruby platforms. Example solution: [Thoughtbot's Appraisal gem](https://github.com/thoughtbot/appraisal) with sets of generated Gemfiles

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby RUBY_VERSION
+
 gem 'ffi-ncurses', '~> 0.3', :platforms => :jruby
 gem 'jruby-openssl', '~> 0.8.8', :platforms => :jruby
 gem 'rake'
@@ -13,24 +15,13 @@ group :test do
   gem 'httpclient', '>= 2.2'
   gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
   gem 'minitest', '>= 5.0.5'
-  gem 'net-http-persistent', '~> 2.9.4'
+  gem 'net-http-persistent'
   gem 'patron', '>= 0.4.2', :platforms => :ruby
   gem 'rack-test', '>= 0.6', :require => 'rack/test'
   gem 'rest-client', '~> 1.6.0', :platforms => [:jruby, :ruby_18]
   gem 'simplecov'
   gem 'sinatra', '~> 1.3'
   gem 'typhoeus', '~> 0.3.3', :platforms => [:ruby_18, :ruby_19, :ruby_20, :ruby_21]
-
-  # Below are dependencies of the gems we actually care about that have
-  # dropped support for older Rubies. Because they are not first-level
-  # dependencies, we don't need to specify an unconstrained version, so we can
-  # lump them together here.
-
-  if RUBY_VERSION < '2'
-    gem 'json', '< 2'
-    gem 'tins', '< 1.7.0'
-    gem 'term-ansicolor', '< 1.4'
-  end
 end
 
 gemspec
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -8,6 +8,7 @@ module Faraday
       dependency 'net/http/persistent'
 
       def net_http_connection(env)
+        proxy_uri = nil
         if (proxy = env[:request][:proxy])
           proxy_uri = ::URI::HTTP === proxy[:uri] ? proxy[:uri].dup : ::URI.parse(proxy[:uri].to_s)
           proxy_uri.user = proxy_uri.password = nil
@@ -16,10 +17,13 @@ module Faraday
             define_method(:user) { proxy[:user] }
             define_method(:password) { proxy[:password] }
           end if proxy[:user]
-          return Net::HTTP::Persistent.new 'Faraday', proxy_uri
         end
 
-        Net::HTTP::Persistent.new 'Faraday'
+        if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
+          Net::HTTP::Persistent.new(name: 'Faraday', proxy: proxy_uri)
+        else
+          Net::HTTP::Persistent.new('Faraday', proxy_uri)
+        end
       end
 
       def perform_request(http, env)
--- a/test/adapters/net_http_persistent_test.rb
+++ b/test/adapters/net_http_persistent_test.rb
@@ -10,8 +10,11 @@ module Adapters
         if defined?(Net::HTTP::Persistent)
           # work around problems with mixed SSL certificates
           # https://github.com/drbrain/net-http-persistent/issues/45
-          http = Net::HTTP::Persistent.new('Faraday')
-          http.ssl_cleanup(4)
+          if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
+            Net::HTTP::Persistent.new(name: 'Faraday').reconnect_ssl
+          else
+            Net::HTTP::Persistent.new('Faraday').ssl_cleanup(4)
+          end
         end
       end if ssl_mode?
     end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: ruby
+before_install: gem install bundler
 script: bundle exec script/test
 cache: bundler
 


Gemfile
```diff
@@ -13,7 +13,11 @@ group :test do
   gem 'httpclient', '>= 2.2'
   gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
   gem 'minitest', '>= 5.0.5'
-  gem 'net-http-persistent', '~> 2.9.4'
+  if RUBY_VERSION >= '2.1'
```
mislav: We can't use conditionals like these in a Gemfile. However, you can use the `:platform` to scope only to certain Rubies.

lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -17,8 +17,11 @@ def with_net_http_connection(env)
             define_method(:password) { proxy[:password] }
           end if proxy[:user]
         end
-
-        yield Net::HTTP::Persistent.new 'Faraday', proxy_uri
+        if Net::HTTP::Persistent::VERSION >= '3'
```
mislav: Instead of checking a version string like this, could we simply check the arity of the constructor?

```
Net::HTTP::Persistent.method(:new).arity
```

We could detect how many arguments does the constructor expect this way, and switch between two different invocation styles without being dependent on the version.

test/adapters/net_http_persistent_test.rb
```diff
@@ -10,8 +10,10 @@ def setup
         if defined?(Net::HTTP::Persistent)
           # work around problems with mixed SSL certificates
           # https://github.com/drbrain/net-http-persistent/issues/45
-          http = Net::HTTP::Persistent.new('Faraday')
-          http.ssl_cleanup(4)
+          if Net::HTTP::Persistent::VERSION < '3'
```
mislav: Ditto my comment about version string checking.

lib/faraday/adapter/net_http_persistent.rb
olleolleolle: Our problem here is that they both (3.x and 2.x) return `-1` for that arity - they both default on each of the given arguments.

Inspired by this comment, I changed the test invocation to avoid testing for the version.

Gemfile
olleolleolle: http://bundler.io/v1.13/man/gemfile.5.html#PLATFORMS seems not to support this use-case.

http://bundler.io/v1.13/man/gemfile.5.html#INSTALL_IF is perhaps a candidate?

iMacTia: @mislav This looks like the only working solution and is what it was done already in the past for other gems at the bottom of the Gemfile. Given the Gemfile is only used for tests, I think that is acceptable.

olleolleolle: @iMacTia Would you want a squash-and-rebase? Any other changes?

test/adapters/net_http_persistent_test.rb
```diff
@@ -10,8 +10,10 @@ def setup
         if defined?(Net::HTTP::Persistent)
           # work around problems with mixed SSL certificates
           # https://github.com/drbrain/net-http-persistent/issues/45
-          http = Net::HTTP::Persistent.new('Faraday')
-          http.ssl_cleanup(4)
+          if Net::HTTP::Persistent::instance_methods.include?(:ssl_cleanup)
+            http = Net::HTTP::Persistent.new('Faraday')
+            http.ssl_cleanup(4)
+          end
```
olleolleolle: @iMacTia: There's no longer any "cleanup" method in 3.0.0.

olleolleolle: (Investigating.)

Is this the current cleanup? https://github.com/drbrain/net-http-persistent/commit/dad5d682a6299e1bb9ef4fed3e2e5837011490bb

I begin to think that this _is_ the current way of "cleaning up" after finishing a connection.

Here's the `ssl_cleanup` implementation's comment (in v2.9.4):

>   ##
>   # Finishes all connections on the given +thread+ that were created before
>   # the given +generation+ in the threads +generation_key+ list.
>   #
>   # See #shutdown for a bunch of scary warning about misusing this method.

Gemfile
iMacTia: No, I personally think that even though it's not the most elegant solution, this is the only one we can use in this specific case. So I'm saying that for me it's ok as it is.

test/adapters/net_http_persistent_test.rb
iMacTia: right, looks like you have to use this `reuse_ssl_sessions`. Did you spot also what the default value is?

olleolleolle: By default, SSL connections are reused: https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent.rb#L409

iMacTia: Ok so in order for tests to work properly, we need to set this to false at the beginning.
I'm wondering if this option exists on v2.9 as well given it was there in 2011 😮

lib/faraday/adapter/net_http_persistent.rb
drbrain: 🙇🏻

I switched this to open the possibility of adding arbitrary configuration keyword argument to the constructor.  Keeping the few positional arguments seemed silly.

test/adapters/net_http_persistent_test.rb
drbrain: The comment immediately above this references issues with SSL client certificates, I think the correct thing to do is to call [`reconnect_ssl`](https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent.rb#L881-L886) to address that.

olleolleolle:
@drbrain Sorry I took forever, but I made a conditional `else` case with the v3.0.0 method you named. The tests still pass. Did I make it the way you mentioned?

olleolleolle:
@iMacTia The last change to the Gemfile is one I was really aching to get in: using the `ruby` version marker to hint to Bundler's Rubygem-finding that we only need matching gems for our version.

See Bundler's docs on Gemfile for (spare) details: http://bundler.io/v1.13/man/gemfile.5.html#RUBY

iMacTia:
Maybe I'm missing something, but looking at [this](http://bundler.io/v1.3/gemfile_ruby.html) I don't find a reason to specify something like `ruby RUBY_VERSION` in the Gemfile.
I understand using it to fi it to a specific version, but using `RUBY_VERSION` doesn't really make sense to me. Am I missing something?

test/adapters/net_http_persistent_test.rb
```diff
@@ -10,8 +10,13 @@ def setup
         if defined?(Net::HTTP::Persistent)
           # work around problems with mixed SSL certificates
           # https://github.com/drbrain/net-http-persistent/issues/45
-          http = Net::HTTP::Persistent.new('Faraday')
-          http.ssl_cleanup(4)
+          if Net::HTTP::Persistent::instance_methods.include?(:ssl_cleanup)
```
drbrain: #reconnect_ssl exists back to net-http-persistent v2.4.1, so you probably don't need this `if`

lib/faraday/adapter/net_http_persistent.rb
drbrain: This will break if there is ever a net-http-persistent 10 release, but otherwise is fine.

test/adapters/net_http_persistent_test.rb
olleolleolle: Thanks!

olleolleolle:
> @iMacTia I understand using it to fi it to a specific version, but using RUBY_VERSION doesn't really make sense to me. Am I missing something?

Oh, the very thing is that we're not a deployed application where dependencies are to be locked down to 1 and only 1 gem selection, but a library which needs 1 selection per Ruby version we support.

_Warning_: This is more of a current measure, Bundler's future has better support for knowing this on its own.

So by leveraging Bundler's "Ruby version hinting", we can try to minimize `if`s in the Gemfile.

Bundler outsources the decision about sub-dependencies' version needs to the gemspecs involved,  by taking the [`required_ruby_version`](http://guides.rubygems.org/specification-reference/#required_ruby_version=) gemspec data into account. 

(You know, I'll try improving the Bundler website with an FAQ item about just this.)

joevandyk:
any update on this?

iMacTia:
There are some important changes to the Gemfile, so I would like @mislav to review this PR as well before merging it in.

aaronmcadam:
Is the issue about how to support both 2.x and 3.x of `net-http-persistent`?

olleolleolle:
@aaronmcadam Yes!

aaronmcadam:
Hey @olleolleolle I've not used it myself, but I've seen https://github.com/thoughtbot/appraisal being used for supporting multiple gem versions.

olleolleolle:
@aaronmcadam I will extend the Description of this PR, in order to make it clear what the changes are and why they're needed.

JulienItard:
Hi @olleolleolle ! Any idea when this will be available ?

Thanks 👍

iancward:
@iMacTia @mislav any update on this? I just ran into this issue as well.

olleolleolle:
@iancward To meet earlier requested change: Formulate a parameters-based version picking, instead of using the VERSION constant.

mtwentyman:
If it is helpful, I've been using a version of the patch provided by @olleolleolle for a few projects (by way of making a new adapter available) and it has worked superbly.

iMacTia:
Tests are passing and changes from @mislav have been implemented so it should now be mergeable.
I'll need to go through it one more time though and then I'll decide if including it into 0.12.2 (soon to be released) or 0.13.0 as it's mid-way between the feature and the fix type of PR 😅.
I know everyone is voting for 0.12.2 so no need to point that out 😄

iMacTia:
Tests are green and the Gemfile is cleaner now!
Moreover, Travis is testing both v2 and v3 of net_http_persistent as v2 will be installed for ruby < 2.1, v3 otherwise.
This was long and suffered, but thanks to @olleolleolle perseverance we can finally merge this in!

olleolleolle:
Celebrate!

Issue #622, opened 2016-11-03
HTTPClient adapter should set SSL timeout


dgolombek:
For reasons that I'm not clear on, HTTPClient on JRuby requires you to set SSL timeout differently than regular timeouts (and only has a singular timeout value, instead of different timeouts for connecting and reading/writing). I've asked about merging the timeouts, but until that's done, Faraday should set the ssl timeout separately.

iMacTia:
Hi @dgolombek and thanks for opening this issue.
Before progressing with this I would like to have a better understanding of the situation.
From what I can see, the issue lies on HTTPClient storing the `timeout` and `ssl_timeout` into two separate variables.
On Faraday's HTTPClient adapter we only set the "normal" timeout, so when performing ssl (https) calls this gets ignored.
However, you already asked HTTPClient gem owner to merge the two timeouts on their side.
I can also see you have already opened an [issue](https://github.com/nahi/httpclient/issues/341) on their side.

By the look of it, I would be more inclined to wait for a fix on HTTPClient gem as I agree with you on this being an inconsistency on their side (the gem behaves differently based on the Ruby version).

I personally don't think having this temporary fixed on Faraday will be of any help, plus:

1. The solution in #623 is a bit dirty: it has to check if the instance responds to a given method that is only available with JRuby.
2. They will fix this at some time on their side, and we could forget this temporary fix inside Faraday. This wouldn't be the first time this happens, so I would rather prefer to avoid it.
3. Even though you get this fixed in faraday, I can't guarantee you a release date for it.

My suggestion at this point is to wait for an answer from HTTPClient gem owner and see what they want to do. If you're eager to get this fixed because you need it on your project, then best thing to do in this case is to monkey-patch either Faraday or the Client until a solution is merged.

dgolombek:
I went back and forth before opening this bug, for the reasons you mention. I will say:

1. The ssl_config.timeout= method is available in both MRI and JRuby -- the respond_to? check is to handle the field possibly going away in the future. The field just isn't used in MRI
2. This problem will continue to hurt anyone who neglects to upgrade HTTPClient, which may take a while

In any case, I'm happy to wait to see the results of the discussion on nahi/httpclient#341 before making a decision either which way here -- it's certainly possibly that the maintainers have a rationale for the separate SSL timeout that I'm not aware of.

iMacTia:
[...]

PR #623, opened 2016-11-03
Set SSL Timeout on HTTPClient adapter

dgolombek:
As mentioned in #622, HTTPClient on JRuby currently requires you to set
the SSL timeout separately than the normal timeouts (although there are
bugs with it with many JRuby versions).


iMacTia:
Closing following [this comment](https://github.com/lostisland/faraday/issues/622#issuecomment-259528819)

PR #625, opened 2016-11-13
HTTPClient adapter shouldn't reset SSL each request

### LINKED ISSUE #624: HTTPClient adapter with SSL enabled should not reset sessions on every request
There is a very bad interplay between Faraday and HTTPClient that causes HTTPClient to flush its session store on every request, causing SSL re-negotiation each request. This kills performance, obviously. Part of this is a bug in HTTPClient but part is hard to work-around there and hence I'm opening an issue here to discuss the best fix.

On every Faraday request the adapter calls configure_ssl, which in turn calls httpclient.ssl_config.verify_mode= and httpclient.ssl_config.cert_store= which have the horrible side effect of calling reset_all on the httpclient object, which closes all open SSL sessions.

In HTTPClient, verify_mode= should only reset all sessions if the verify mode is changed, which I'll submit a patch to HTTPClient to fix. But for cert_store= it's much harder, since the cert store is a OpenSSL::X509::Store object which doesn't have a comparator and writing one would be challenging (it's a C struct). I assume that Faraday sets the cert store so that it can ensure that set_default_paths is called on it, since HTTPClient.ssl_config.cert_store is already a OpenSSL::X509::Store. 

The caller can work around this by providing it's own cert_store, but it's not reasonable to require that of the caller to get decent performance. So I'm trying to figure out best fix here. My first thoughts:

1. Store the cert_store inside the request ssl options, provide the same one to HTTPClient each time, allowing a simple comparator in HTTPClient to determine when it has changed
2. Call set_default_paths on HTTPClient.ssl_config.cert_store directly
3. Figure out how to write a comparator on OpenSSL::X509::Store
4. Add some intelligence to the HTTPClient adapter to only do SSL configuration the first time a call is made to a given URL.

Option 1 is the easiest and safest, since we could basically just do
ssl[:cert_store] ||= OpenSSL::X509::Store.new
but I wanted to see if there were other thoughts before I submitted a patch doing so.

dgolombek:
This fixes #624, in conjunction with nahi/httpclient#343. The same
cert_store is passed to HTTPClient each request, so equality check
inside HTTPClient will skip the reset.

### DIFF
--- a/lib/faraday/adapter/httpclient.rb
+++ b/lib/faraday/adapter/httpclient.rb
@@ -100,12 +100,12 @@ module Faraday
         # Memoize the cert store so that the same one is passed to
         # HTTPClient each time, to avoid resyncing SSL sesions when
         # it's changed
-        if @cert_store.nil?
+        @cert_store ||= begin
           # Use the default cert store by default, i.e. system ca certs
-          @cert_store = OpenSSL::X509::Store.new
-          @cert_store.set_default_paths
+          cert_store = OpenSSL::X509::Store.new
+          cert_store.set_default_paths
+          cert_store
         end
-        @cert_store
       end
 
       def ssl_verify_mode(ssl)
--- a/lib/faraday/adapter/httpclient.rb
+++ b/lib/faraday/adapter/httpclient.rb
@@ -97,10 +97,15 @@ module Faraday
 
       def ssl_cert_store(ssl)
         return ssl[:cert_store] if ssl[:cert_store]
-        # Use the default cert store by default, i.e. system ca certs
-        cert_store = OpenSSL::X509::Store.new
-        cert_store.set_default_paths
-        cert_store
+        # Memoize the cert store so that the same one is passed to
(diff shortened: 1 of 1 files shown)


iMacTia:
Hi @dgolombek, I can see tests are failing due to the following:

```
FAILED: this test suite doesn't tolerate Ruby syntax warnings!
```

I suggest you amend your code trying with the lazy-evaluation:

```ruby
@cert_store ||= begin
  # Use the default cert store by default, i.e. system ca certs
  @cert_store = OpenSSL::X509::Store.new
  @cert_store.set_default_paths
 end
```

And run tests locally to check they're passing (even though I'm not sure you'll see the warning error locally)

dgolombek:
Interesting, tests were passing locally for me, with jruby-9.1.5.0 - I should've run with MRI as well. Updated

iMacTia:
Thanks @dgolombek for digging on this and coordinating the work between Faraday and HTTPClient.
Adding this to the Faraday 0.10.1 release, please give me some time to finish my issues/PR review to check if there's anything else that can be added to the same release

PR #629, opened 2016-11-21
handle default connection options set with hash

### LINKED ISSUE #628: calling `url` on a `Hash` (NoMethodError)
faraday-0.10.0/lib/faraday/connection.rb:60 can now call url on a hash because of the changes on the line above it.

```
     Failure/Error: response = Faraday.new(url: gateway_url).run_request(*request_parameters)

     NoMethodError:
       undefined method `url' for #<Hash:0x007f757ad12f58>
     # ./vendor/bundle/ruby/gems/faraday-0.10.0/lib/faraday/connection.rb:63:in `initialize'
     # ./vendor/bundle/ruby/gems/faraday-0.10.0/lib/faraday.rb:74:in `new'
     # ./vendor/bundle/ruby/gems/faraday-0.10.0/lib/faraday.rb:74:in `new'
     # ./app/middleware/gateway_proxy.rb:24:in `proxy_response'
```

0.9.2:
```
        options = ConnectionOptions.from(url)
        url     = options.url
```

0.10.0:
```
        options = options ? options.merge(url) : ConnectionOptions.from(url)
        url     = options.url
```

Called with code:
`response = Faraday.new(url: gateway_url).run_request(*request_parameters)`

Cohen-Carlisle:
Fixes #628.

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -34,9 +34,6 @@ module Faraday
     #     Faraday.get "https://faraday.com"
     attr_writer :default_connection
 
-    # Public: Sets the default options used when calling Faraday#new.
-    attr_writer :default_connection_options
-
     # Public: Initializes a new Faraday::Connection.
     #
     # url     - The optional String base URL to use as a prefix for all
@@ -120,6 +117,11 @@ module Faraday
   # Returns a Faraday::ConnectionOptions.
   def self.default_connection_options
     @default_connection_options ||= ConnectionOptions.new
+  end
+
+  # Public: Sets the default options used when calling Faraday#new.
+  def self.default_connection_options=(options)
+    @default_connection_options = ConnectionOptions.from(options)
   end
 
   unless const_defined? :Timer
--- a/test/connection_test.rb
+++ b/test/connection_test.rb
@@ -2,6 +2,10 @@ require File.expand_path('../helper', __FILE__)
 
 class TestConnection < Faraday::TestCase
 
+  def teardown
+    Faraday.default_connection_options = nil
+  end
+
   def with_env(key, proxy)
     old_value = ENV.fetch(key, false)
     ENV[key] = proxy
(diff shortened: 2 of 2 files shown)


iMacTia:
Hi @Cohen-Carlisle and thanks for your PR.
Would you mind adding some tests to your PR in order to double-check that all behaviours are still working? In particular, I would like the following to be carefully tested:

1. The following should now work **and be respected**:

```ruby
Faraday.default_connection_options = { request: { timeout: 0.0001 } }
Faraday.new(url: "http://www.example.com/").get("/")
```

2. The following should still work and be respected:

```ruby
Faraday.default_connection_options = { request: { timeout: 0.0001 } }
Faraday.new("http://www.example.com/").get("/")
```

3. We should allow to specify additional options when creating the connection (and check that these are respected):

```ruby
Faraday.default_connection_options = { request: { timeout: 0.0001 } }
Faraday.new("http://www.example.com/", { request: { open_timeout: 0.0001 }}).get("/")
# Same for Faraday.new(url: "http://www.example.com/", request: { open_timeout: 0.0001 }).get("/")
```

Cohen-Carlisle:
I'll work on adding tests in.

iMacTia:
Hi @Cohen-Carlisle , how are the test works going :)?

Cohen-Carlisle:
My bad, got busy between work, play, and family. I will work on this tonight.

Cohen-Carlisle:
@iMacTia while writing tests I discovered an interesting behavior with default vs object-specific connection options. In the last example you provided:
```ruby
Faraday.new("http://www.example.com/", { request: { open_timeout: 0.0001 }})
```
The `request` options in new overwrite any nested `request` options. E.g., either:
```ruby
Faraday.default_connection_options.request.timeout = 10
```
or 
```ruby
Faraday.default_connection_options = { request: { timeout: 10 } }
```
will lose the `request.timeout` setting. Basically, merges are not deep merges.

Cohen-Carlisle:
I pushed up tests that should cover the cases requested, with the possible exception of the merge behavior described above.

iMacTia:
That's unfortunately what I feared 😞
Question now is: was it working properly on 0.9.2?
I'll try to do the same there and see what happens, my feeling is that we just uncovered an old bug

iMacTia:
Aaaaaaaand yes, we have an old bug.
But this will open up discussions like "what should I do to remove a default option for a single request?" and so on, so better to address it on a separate issue.
Thanks @Cohen-Carlisle. Great work, nice addition 👍

PR #633, opened 2016-11-22
Travis: use ruby-2.3.3

olleolleolle:
This PR changes the Travis configuration to use latest stable 2.3 ruby.

- https://www.ruby-lang.org/en/news/2016/11/15/ruby-2-3-2-released/ [link to its changelog](http://svn.ruby-lang.org/repos/ruby/tags/v2_3_2/ChangeLog)
- https://www.ruby-lang.org/en/news/2016/11/21/ruby-2-3-3-released/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ rvm:
   - 2.0.0
   - 2.1
   - 2.2
-  - 2.3.1
+  - 2.3.3
   - jruby-19mode
   - jruby-20mode
   - jruby-21mode


iMacTia:
Apologies for my ignorance on Travis, but is there any way to tell it "use the latest available Ruby 2.3.x"?

olleolleolle:
@iMacTia Sorry, not yet.

mislav:
Since Travis uses rvm, it should be possible to use `2.3`, but the last time I tried that for the 2.3.x line of releases, it didn't work as advertised. So, it's best to stick to an exact version for now.

However, I don't think we need to be updating patch versions of Ruby for our CI suite. We literally don't gain anything by doing it. See my explanation here https://github.com/mislav/will_paginate/pull/505#issuecomment-250900275

**TL;DR:** since a patch version number change in Ruby means no feature changes, we're not going to uncover any potential incompatibilities in Faraday by upgrading from e.g. 2.3.1 to 2.3.3.

olleolleolle:
Thanks for explaining. <3

PR #634, opened 2016-11-22
Carry #579: Faraday::Utils::Headers YAML can be serialized

olleolleolle:
This PR is #579 and one more commit that has a test.

It exercises Header through to_yaml and YAML.load, and then uses its special methods with a simple assert on the output.

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -113,7 +113,7 @@ module Faraday
       end
 
       def encode_with(coder)
-        coder['names'] = names
+        coder['names'] = @names
       end
 
       protected
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -112,6 +112,10 @@ module Faraday
         @names = coder['names']
       end
 
+      def encode_with(coder)
+        coder['names'] = names
+      end
+
       protected
 
       def names
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -108,6 +108,10 @@ module Faraday
           }
       end
 
+      def init_with(coder)
+        @names = coder['names']
+      end
+
       protected
 
       def names
--- a/test/utils_test.rb
+++ b/test/utils_test.rb
@@ -63,5 +63,15 @@ class TestUtils < Faraday::TestCase
       Faraday::Utils.default_uri_parser = old_parser
     end
   end
+
+  # YAML parsing
+
+  def test_headers_yaml_roundtrip
+    headers = Faraday::Utils::Headers.new('User-Agent' => 'safari', 'Content-type' => 'text/html')
+    result = YAML.load(headers.to_yaml)
+
+    assert result.include?('user-agent'), 'Unable to hydrate to a correct Headers'
+    assert result.include?('content-type'), 'Unable to hydrate to a correct Headers'
+  end
 end
 


iMacTia:
Thanks @olleolleolle, happy to see tests are passing now 👍 !
The only thing I'm not completely convinced with is the use of the getter for `names` when encoding

```ruby
def encode_with(coder)
  coder['names'] = names
end
```

The issue I see is that if in the future we change the getter to perform some operations on the raw value, we would end up encoding the result rather than the raw value.
I agree that getters should always be the first choice to access an instance variable, but in the context of serialisation (YAML in this case) I think using the raw value could be better

olleolleolle:
@iMacTia Happy to use the `@names` instead. I'll pen that in on the site right now.

olleolleolle:
@iMacTia there.

iMacTia:
Thanks @olleolleolle , we can finally close this and add it to the next release 👍

qzio:
Really happy about this  :+1:

seanbjornsson:
Hey @olleolleolle (and @iMacTia ), I have been hunting down an issue in my application and it has led me to this PR. 
I saw you added `encode_with` [here](https://github.com/lostisland/faraday/pull/579#issuecomment-262380312) to make the YAML round trip function. I would argue that, while it does not error, and your test passes, it no longer works correctly. Specifically, I am referring to the original functionality of accessing headers with hash syntax, and generally reconstructing the object to its original state.  Here is your test and example with more lines:
```ruby
# YAML parsing

def test_headers_yaml_roundtrip
  headers = Faraday::Utils::Headers.new('User-Agent' => 'safari', 'Content-type' => 'text/html')

  assert headers.include?('user-agent'), 'Unable to assert inclusion of initialized Headers'
  assert headers.include?('content-type'), 'Unable to assert inclusion of initialized Headers'
  assert headers['user-agent'] == 'safari', 'Unable to access initialized Headers'
  assert headers['content-type'] == 'text/html', 'Unable to access initialized Headers'
  
  result = YAML.load(headers.to_yaml)

  assert result.include?('user-agent'), 'Unable to assert inclusion of rehydrated Headers'
  assert result.include?('content-type'), 'Unable to assert inclusion of rehydrated Headers'
  assert result['user-agent'] == 'safari', 'Unable to access rehydrated Headers'
  assert result['content-type'] == 'text/html', 'Unable to access rehydrated Headers'
end
``` 
Everything except the last two assertions pass. I am unfamiliar with Test::Unit, so here it is in the ruby console:
```ruby
irb(main):001:0> headers = Faraday::Utils::Headers.new('User-Agent' => 'safari', 'Content-type' => 'text/html')
=> {"User-Agent"=>"safari", "Content-type"=>"text/html"}
irb(main):002:0> headers.include?('user-agent')
=> true
irb(main):003:0> headers.include?('content-type')
=> true
irb(main):004:0> headers['user-agent']  == 'safari'
=> true
irb(main):005:0> headers['content-type']  == 'text/html'
=> true
irb(main):006:0> result = YAML.load(headers.to_yaml)
=> {}
irb(main):007:0> result.include?('user-agent')
=> true
irb(main):008:0> result.include?('content-type')
=> true
irb(main):009:0> result['user-agent'] == 'safari'
=> false
irb(main):010:0> result['content-type'] == 'text/html'
=> false
```
You'll notice that the inspect value of `headers` and `result` also differ, where `result` no longer presents as a populated hash, but an empty one. 
Your original test passed, because `Faraday::Utils::Headers#include?` ([here](https://github.com/lostisland/faraday/blob/master/lib/faraday/utils.rb#L68)) just checks the `@names` variable, which is the only thing retained through the YAML round trip, due to logic in [`init_with`](https://github.com/lostisland/faraday/blob/master/lib/faraday/utils.rb#L118) and [`encode_with`](https://github.com/lostisland/faraday/blob/master/lib/faraday/utils.rb#L122) as added by this PR. 

That all being said: Why was this added? Before this addition, a Faraday::Utils::Headers instance converted to YAML would present as `ruby/hash` with the correct key value pairs, while now it converts to a `ruby/object` and does not retain key value pairs, but instead retains the names collection. 

For example:
```ruby
# Before encode_with was added
irb(main):001:0> puts headers.to_yaml
--- !ruby/hash:Faraday::Utils::Headers
User-Agent: safari
Content-type: text/html
=> nil

# After encode_with was added
irb(main):001:0> puts headers.to_yaml
--- !ruby/object:Faraday::Utils::Headers
names:
  user-agent: User-Agent
  content-type: Content-type
=> nil
```

Finally, I would like to point out that what lead me to this PR, and identifying this issue, was through upgrading from Faraday v9.1 to Faraday v11.0, thereby getting this new code. The upgrade broke my Faraday::HttpCache when storing the headers of Faraday requests and responses as YAML in a Redis store. They would not correctly survive the YAML round trip, as described above, so the essential header information, particularly the Cache-Control header, was missing. 
Specifically it fails on the storage of the YAML version, [here](https://github.com/plataformatec/faraday-http-cache/blob/master/lib/faraday/http_cache/storage.rb#L47), as serialized by [Psych](https://github.com/ruby/psych), [here](https://github.com/ruby/psych/blob/master/lib/psych.rb#L408)

Also, to be clear, apart from the two versions of Faraday, I discovered this in an upgrade from Rails 4.0 to 4.2 and a bundle update, but I tested this under both rails versions, and both pointed to the Faraday update.

Any insight you (or anyone else reading this) might have would be appreciated. For now, I have fixed my application by removing that method from `Faraday::Utils::Headers`, but I dislike that solution as a temporary monkey patch. 
```
Faraday::Utils::Headers.send(:remove_method, :encode_with)
```

iMacTia:
Hi @seanbjornsson and thanks for the detailed report of the issue.
We definitely missed that because of the successful `include?`.
I find a bit strange that `include?` returns true while the hash access doesn't return anything.
I'll definitely investigate further as I think we're just missing a small bit.

> That all being said: Why was this added?

You can see the details of why we added this fix in #579

PR #635, opened 2016-11-23
Integration test: raise error when no live server is configured

olleolleolle:
This PR is a developer-experience change to the integration tests: Before using a piece of configuration ("live_server") which may be `nil`, the test setup fails early and presents the user with an explanatory error.

  - Add a discoverable tip when trying to use a live server that is not configured

The benefit is that people who hack on the test suite learn their way around.

This happens to the user who tries to run:

```shell
$ ruby -Ilib:test test/adapters/net_http_test.rb
```

### DIFF
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -9,7 +9,7 @@ module Adapters
   # `#adapter_options` optional. extra arguments for building an adapter
   module Integration
     def self.apply(base, *extra_features)
-      if base.live_server?
+      if base.live_server
         features = [:Common]
         features.concat extra_features
         features << :SSL if base.ssl_mode?
@@ -247,8 +247,6 @@ module Adapters
         end
 
         server = self.class.live_server
-        raise 'Integration test suite: Can not continue without live_server ' \
-              'configured. See CONTRIBUTING for usage.' unless server
         url = '%s://%s:%d' % [server.scheme, server.host, server.port]
 
         options[:ssl] ||= {}
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -246,6 +246,8 @@ module Adapters
         end
 
         server = self.class.live_server
+        raise 'Integration test suite: Can not continue without live_server ' \
+              'configured. See script/test for more.' unless server
         url = '%s://%s:%d' % [server.scheme, server.host, server.port]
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
@iMacTia Perhaps the failure should even point to a section in the README? Or say "Just run `rake` if you're unsure"?

olleolleolle:
Or, is it better with a 

```ruby
warn 'Integration test suite: Can not continue without live_server configured. See script/test for more.' unless server
```

iMacTia:
Hi @olleolleolle, thanks for pointing this out.
To be honest I never had this issue because I always run `script/test`.
I think the issue here is that you should use another syntax to run a single test, as explained in the [CONTRIBUTING](https://github.com/lostisland/faraday/blob/master/CONTRIBUTING.md) file.

What did you do to run tests after you've seen that `ruby -Ilib:test test/adapters/net_http_test.rb` was failing?
If you didn't already, can you please try the solution shown in the `CONTRIBUTING` and see if it works?

olleolleolle:
What I did was:

- give up, used `rake` to have more tests run
- then I noted `script/test` exists. Its explanation was not as good as CONTRIBUTING's

This PR is about having the new developer discover this instruction.

olleolleolle:
@iMacTia But close it if you want to - now _I_ know.

iMacTia:
I'm happy to improve contributors experience, but I would like to do it effectively.
If we can make things clearer and save time to future contributors, I think that's a good step forward.
There's something puzzling me though, I can see the same warning is already in place on the `apply` method [here](https://github.com/lostisland/faraday/blob/a712938071bb0997410b5de4a4f2517d30fc4d3d/test/adapters/integration.rb#L19).

So I'm wondering now why you didn't see it? As far as I can see that method is executed for all tests (as `Integration` is required on top of each file and `Integration.apply` called before tests start)

olleolleolle:
Oh, look: 

`if base.live_server` is different from `if base.live_server?` and the former actually does the right thing.

olleolleolle:
Thanks for holding off merging half-baked things.

iMacTia:
No problem, that's why I'm here reviewing the code 👍 !
Sometimes I can be annoying, but the most important thing is that we fix things up together.

I can see the warning as well now, merging :)

PR #636, opened 2016-11-23
CONTRIBUTING: Update link to Changelog as wiki page has moved

olleolleolle:
This PR changes a URL which pointed at the wiki before. Now it points to Releases on Github.


iMacTia:
good spot 👍

PR #361, opened 2014-04-04
WIP: Support for 'no_proxy' environment variable.

johnbellone:
These are fixes that @mislav suggest in #247 comments.

PR #637, opened 2016-11-23
DO NOT MERGE, WIP: Carry #361: Support 'no_proxy' env variable

olleolleolle:
This PR, which is #361 and changes to make tests pass, is in progress :fire:

**Superseded by** #658 which does the same thing without adding the concept of "default middleware".

TODO: 

- [ ] DESIGN DECISION: decide how to get the added middleware into `default` status
- [ ] make the `default` case pass its 2 test failures due to same

```
  1) Failure:
Adapters::DefaultTest#test_proxy_auth_fail [/home/travis/build/lostisland/faraday/test/adapters/integration.rb:212]:
Faraday::ConnectionFailed expected but nothing was raised.

  2) Failure:
Adapters::DefaultTest#test_proxy [/home/travis/build/lostisland/faraday/test/adapters/integration.rb:203]:
Proxy was not able to append Via header.
Expected /:3998$/ to match nil.
```

### DIFF
--- a/test/adapters/integration.rb
+++ b/test/adapters/integration.rb
@@ -192,7 +192,7 @@ module Adapters
 
       def test_proxy
         proxy_uri = URI(ENV['LIVE_PROXY'])
-        conn = create_connection(:proxy => proxy_uri)
+        conn = create_connection(:proxy => { uri: proxy_uri })
 
         res = conn.get '/echo'
         assert_equal 'get', res.body
@@ -206,7 +206,7 @@ module Adapters
       def test_proxy_auth_fail
         proxy_uri = URI(ENV['LIVE_PROXY'])
         proxy_uri.password = 'WRONG'
-        conn = create_connection(:proxy => proxy_uri)
+        conn = create_connection(:proxy => { uri: proxy_uri, password: proxy_uri.password })
 
         err = assert_raises Faraday::Error::ConnectionFailed do
           conn.get '/echo'
@@ -235,10 +235,17 @@ module Adapters
       end
 
       def create_connection(options = {})
+        proxy_options = options.delete(:proxy)
         if adapter == :default
           builder_block = nil
         else
           builder_block = Proc.new do |b|
+            if proxy_options
+              b.request :proxy, proxy_options
+            else
+              b.request :proxy
+            end
+
             b.request :multipart
             b.request :url_encoded
             b.adapter adapter, *adapter_options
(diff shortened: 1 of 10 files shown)


test/adapters/integration.rb
```diff
@@ -235,17 +235,26 @@ def adapter_options
...
+        builder_block =
+          if adapter == :default
+            nil
+          else
+            Proc.new do |b|
+              if options[:proxy]
+                b.request :proxy, options[:proxy]
+              else
+                b.request :proxy
+              end
```
olleolleolle: Perhaps this part will just fade away when _all_ connections will have the Request::Proxy middleware?

somethingnew2-0:
Any updates @olleolleolle?

olleolleolle:
The main question to resolve is how to introduce the concept of enabled-by-default middleware that the user can also disable or replace. Deciding on a design for that. (Then this middleware can be in that collection.)

Because of this practical difficulty of introducing a new concept, I've considered rewriting it all to land the NO_PROXY supper in the current location in the initialize - instead of as a middleware. 

So, the status is that I'm stalled and kind of have nobody to talk to it with.

somethingnew2-0:
Yeah, I've been looking through this PR and the Faraday source most of today.  It looks like rewriting the proxy code as a middleware for just this one feature of NO_PROXY is a little overkill.  

However, the proxy code definitely looks cleaner as a middleware module.  I can see the problem with designing this middleware to be as enabled-by-default as it makes it a unique component and almost "unmodular" because it has to be disabled to be excluded.

olleolleolle:
@somethingnew2-0 You nailed it.

olleolleolle:
See also #658.

olleolleolle:
Closing in favor of #658.

iMacTia:
Thanks @olleolleolle for attempting this.

I agree with you and @somethingnew2-0 that having a middleware as suggested by @mislav is indeed the best thing, but I also understand this solution can be complicated to be implemented in the current gem structure.
I decided to review @somethingnew2-0 PR that address this without refactoring the proxy management into a middleware, I think it will be enough for the next release.

Thanks for your work so far, I would be happy if in the future you want to start from @somethingnew2-0 changes to create the middleware and we can discuss together the challenges you were facing while doing that. Maybe we'll need to introduce some new feature in Faraday beforehand (default middleware?)

olleolleolle:
Thank _you_ @iMacTia and @somethingnew2-0 for great teamwork and stewardship.

Thank you to all the people involved in all the earlier steps of this effort. Props to @johnbellone who took the idea and ran with it! Thanks @mislav, for _having_ the idea! Thanks all commenters, you add tons of insight. (Hi @reggieb!)

(Does Faraday have a community chat channel?)

iMacTia:
> (Does Faraday have a community chat channel?)
We're getting that done 😄!

PR #638, opened 2016-11-24
Include response on ClientErrors raised while parsing

percyhanna:
The `ClientError` class allows a `response` object to be defined, and yet it does not seem to be defined in many cases. For parsing errors in particular, it can be beneficial for logging and debugging if the response object was properly available on the error.

Originally I was looking to incorporate this change in the [`FaradayMiddleware::ResponseMiddleware#parse` method](https://github.com/lostisland/faraday_middleware/blob/master/lib/faraday_middleware/response_middleware.rb#L45-L56) by passing `env` instead of `env[:body]`, but then I realized it was part of the standard `Faraday::Response::Middleware` interface to only pass the `body` in to the `parse` method.

An alternative implementation could be to raise a new `ClientError` instance instead of setting the `response` value on the error. In order to maintain the original error's class, one would need to create a new instance of the original error, e.g.: `raise err.class.new(err, env.response)`. This would also require changing the `TimeoutError#initialize` method to accept `response` as an optional parameter.

I'm also wondering if perhaps there was a specific reason the `response` object was not being set more consistently on `ClientError`s.

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -3,7 +3,7 @@ module Faraday
   class MissingDependency < Error; end
 
   class ClientError < Error
-    attr_reader :response
+    attr_accessor :response
 
     def initialize(ex, response = nil)
       @wrapped_exception = nil
--- a/lib/faraday/response.rb
+++ b/lib/faraday/response.rb
@@ -14,6 +14,9 @@ module Faraday
       # Calls the `parse` method if defined
       def on_complete(env)
         env.body = parse(env.body) if respond_to?(:parse) && env.parse_body?
+      rescue Faraday::Error::ClientError => err
+        err.response ||= env.response
+        raise err
       end
     end
 
--- a/test/response_middleware_test.rb
+++ b/test/response_middleware_test.rb
@@ -70,3 +70,32 @@ class ResponseNoBodyMiddleWareTest < Faraday::TestCase
     assert_equal nil, @conn.get('not_modified').body
   end
 end
+
+class ClientErrorResponseMiddleWareTest < Faraday::TestCase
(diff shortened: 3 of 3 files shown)


iMacTia:
Hi @percyhanna and thanks for your PR.

I had a closer look to the issue and I believe your first thought was correct.
What I see in this case is that `FaradayMiddleware::ResponseMiddleware` is not raising the exception properly, ignoring the `response` parameter. Main reason for this, as you noticed, is probably that only `env[:body]` is passed to the `parse` method.

With the above in mind, I don't think your solution of "catch-all-and-readapt" is the best solution in this context, I'd rather prefer to go with your original idea of changing `FaradayMiddleware::ResponseMiddleware`.

I noticed that `parse` is never called directly by subclasses implementation.
They simply define the parser to use. The normal flow is then the following:

```ruby
call(env)
process_response(env)
parse(env[:body])
```

Best place to me to fix this up is either `process_response(env)` or `call(env)`.
I think the best solution would be similar to the one you proposed:

> An alternative implementation could be to raise a new ClientError instance instead of setting the response value on the error. In order to maintain the original error's class, one would need to create a new instance of the original error, e.g.: raise err.class.new(err, env.response). 

In this case it will be easy though because you only need to catch `Faraday::Error::ParsingError` rather than the more generic `ClientError`.

Would this solve your issue?

percyhanna:
Thanks, @iMacTia. Is https://github.com/lostisland/faraday_middleware/pull/144 what you had in mind?

iMacTia:
Yes, definitely better!
Closing here in favour of https://github.com/lostisland/faraday_middleware/pull/144

PR #642, opened 2016-12-17
Fix Net::HTTP proxy issue

shawzt:
1. Net::HTTP.Proxy is obsolete;
2. Net::HTTP use proxy from env as default

> Please refer to https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb
Net::HTTP use proxy from env as default, so even though we set `''` to faraday proxy and use `net_http` as the adapter, in the end, `Net::HTTP` will still use the proxy from system env.
>
> And `Net::HTTP.Proxy` is obsolete, so move to use `Net::HTTP`

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -90,15 +90,14 @@ module Faraday
 
       def net_http_connection(env)
         proxy = env[:request][:proxy]
+        addr  = env[:url].host
+        port  = env[:url].port || (env[:url].scheme == 'https' ? 443 : 80)
         # HTTP.Proxy is obsolete, and Net::HTTP will use proxy from env as default
-        Net::HTTP.new(
-          env[:url].host,
-          env[:url].port || (env[:url].scheme == 'https' ? 443 : 80),
-          proxy[:uri].host,
-          proxy[:uri].port,
-          proxy[:user],
-          proxy[:password]
-        )
+        if proxy.nil?
+          Net::HTTP.new(addr, port, nil)
+        else
+          Net::HTTP.new(addr, port, proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])
+        end
       end
 
       def configure_ssl(http, ssl)
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -89,11 +89,16 @@ module Faraday
       end
 
       def net_http_connection(env)
-        if proxy = env[:request][:proxy]
-          Net::HTTP::Proxy(proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])
-        else
-          Net::HTTP
-        end.new(env[:url].host, env[:url].port || (env[:url].scheme == 'https' ? 443 : 80))
+        proxy = env[:request][:proxy]
+        # HTTP.Proxy is obsolete, and Net::HTTP will use proxy from env as default
+        Net::HTTP.new(
+          env[:url].host,
+          env[:url].port || (env[:url].scheme == 'https' ? 443 : 80),
+          proxy[:uri].host,
+          proxy[:uri].port,
+          proxy[:user],
+          proxy[:password]
+        )
       end
 
       def configure_ssl(http, ssl)
--- a/test/adapters/net_http_test.rb
+++ b/test/adapters/net_http_test.rb
@@ -40,5 +40,24 @@ module Adapters
       assert_equal 1234, http.port
     end
 
+    def test_no_proxy_with_nil_input
+      url = URI('http://example.com')
+
+      adapter = Faraday::Adapter::NetHttp.new
+      http = adapter.net_http_connection(:url => url, :request => { :proxy => nil })
+
+      assert_equal false, http.proxy_from_env?
+    end
+
+    def test_no_proxy_with_empty_string
+      url = URI('http://example.com')
+
+      adapter = Faraday::Adapter::NetHttp.new
+      proxy = Faraday::ProxyOptions.from('')
+      http = adapter.net_http_connection(:url => url, :request => { :proxy => proxy })
+
+      assert_equal false, http.proxy_from_env?
+    end
+
   end
 end
--- a/test/options_test.rb
+++ b/test/options_test.rb
@@ -78,6 +78,14 @@ class OptionsTest < Faraday::TestCase
     assert_equal 'example.org', options.host
     assert_equal 'http', options.scheme
   end
+  
+  def test_proxy_options_from_empty_string
+    options = Faraday::ProxyOptions.from ''
+    assert_nil options.user
+    assert_nil options.password
+    assert_kind_of URI, options.uri
+    assert_nil options.host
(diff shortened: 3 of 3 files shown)


iMacTia:
Hi @shawzt and thanks for your PR.
I can see you're right, `Net::HTTP::Proxy` have been declared obsolete so we should start using the new parameters in `Net::HTTP.new`.
However, your code doesn't take into consideration that `env[:request][:proxy]` is optional and thus could be `nil`.
This is causing many tests to fail. After getting `proxy` you should still check for his presence with `if proxy.nil?` and then use the 2 params or 6 params constructor accordingly.

Happy to merge your changes if you address the above issue.

Thanks

lib/faraday/adapter/net_http.rb
```diff
@@ -89,11 +89,16 @@ def with_net_http_connection(env)
       end
 
       def net_http_connection(env)
-        if proxy = env[:request][:proxy]
-          Net::HTTP::Proxy(proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])
-        else
-          Net::HTTP
-        end.new(env[:url].host, env[:url].port || (env[:url].scheme == 'https' ? 443 : 80))
+        proxy = env[:request][:proxy]
```
iMacTia: `env[:request][:proxy]` could be nil, which causes all `proxy[...]` calls to fail later in this case.
You will need to test it with `if proxy.nil?` and use the correct `Net::HTTP.new` method accordingly

shawzt:
@iMacTia Thanks for the review and comments, I have fixed the issue and made the CI green. Please help to merge if the PR is ok.

lib/faraday/adapter/net_http.rb
```diff
@@ -89,11 +89,15 @@ def with_net_http_connection(env)
...
 
       def net_http_connection(env)
-        if proxy = env[:request][:proxy]
-          Net::HTTP::Proxy(proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])
+        proxy = env[:request][:proxy]
+        addr  = env[:url].host
+        port  = env[:url].port || (env[:url].scheme == 'https' ? 443 : 80)
+        # HTTP.Proxy is obsolete, and Net::HTTP will use proxy from env as default
+        if proxy.nil?
+          Net::HTTP.new(addr, port, nil)
```
iMacTia: Sending `nil` as the 3rd parameter we change the default behaviour (use the env proxy) and we also make it impossible to enable it back.
This is not acceptable as it could break existing code.

shawzt: Hi @iMacTia 

Thanks for your comments.

However, sending `nil` as the 3rd parameter to `Net::HTTP.new(addr, port, nil) if proxy.nil?` is the correct behavior.

Why needs to send `nil` as the 3rd parameter to `Net::HTTP.new if proxy.nil?`?
+ Net::HTTP will still use the proxy from system env since the changes made in ruby/ruby@9416bec
+ Faraday has its own implementation regarding to proxy in its net/http adapter

---
Generally speaking, here are the cases configured for proxy in `Net::HTTP` adapter:
+ no proxy key when `Faraday.new`
  - use __ENV['http_proxy']__
+ proxy: nil when `Faraday.new`
  - no proxy
+ proxy: { url: xxx, port: xxx } when `Faraday.new`
  - use __proxy: { url: xxx, port: xxx }__
---

Here are relevant code:
+ new Faraday
```
lib/faraday.rb

def new(url = nil, options = nil)
   block = block_given? ? Proc.new : nil
   options = options ? default_connection_options.merge(options) : default_connection_options.dup
   Faraday::Connection.new(url, options, &block)
end
```

+ Faraday connection
```
@proxy = nil
proxy(options.fetch(:proxy) {
    uri = ENV['http_proxy']
    if uri && !uri.empty?
       uri = 'http://' + uri if uri !~ /^http/i
       uri
    end
})
```
```
def proxy(arg = nil)
   return @proxy if arg.nil?
   @proxy = ProxyOptions.from(arg)
end
```

So per my investigation and fix, it will not break the existing code based on current Faraday without this fix.

shawzt:
Hi @iMacTia 

Happy new year 🎉 . Have you got time to review my comments?

~
Thanks,
Shaw

iMacTia:
Hi @shawzt, thanks for the little bump.
Indeed I read your comment as soon as you posted it, but need some time to review carefully.
This is what I'm not convinced about:

```ruby
1    if proxy.nil?
2      Net::HTTP.new(addr, port, nil)
3    else
4      Net::HTTP.new(addr, port, proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])
5    end
```

Above is the current code form your PR. As far as I understood:

Line 2: `Net::HTTP.new(addr, port, nil)` will create a connection with no proxy at all. This ignore also any `http_proxy` environment variable.

Line 4: `Net::HTTP.new(addr, port, proxy[:uri].host, proxy[:uri].port, proxy[:user], proxy[:password])` will create a connection using the proxy settings from the user (if any is provided).

What I'm missing is how do you create a connection using the `http_proxy` environment variable?

iMacTia:
I think I got now what you said @shawzt:

```ruby
@proxy = nil
proxy(options.fetch(:proxy) {
    uri = ENV['http_proxy']
    if uri && !uri.empty?
       uri = 'http://' + uri if uri !~ /^http/i
       uri
    end
})
```

The above is causing the env proxy being used.
I'm having a closer look

iMacTia:
Ok @shawzt, I think I got the full picture now, sorry if it took me some time 😅 
Do you think you would be able to add a simple test to demonstrate that fails on the master branch but passes on your fixed branch?

shawzt:
@iMacTia sure, will add the tests later

iMacTia:
Thanks @shawzt 
Release for v0.11.0 is set for tomorrow, I'd be really happy to merge your branch in if you add the tests

shawzt:
Hi @iMacTia 

The tests are failed at __master__ branch.
image

How should I test my fixed branch? I tested it locally.

iMacTia:
Hi @shawzt.

By **master** I mean the Faraday master branch (the one without your changes), your fixed branch is **your master** (so where your changes are applied).
From what I can see, we have 2 issues here:

1. Your tests are failing on Faraday master because we still support ruby1.9.3 and method `proxy_from_env?` is not available there
2. I added the following test (**I would suggest you add it as well**):

```ruby
def test_proxy_from_env_by_default
  url = URI('http://example.com')

  adapter = Faraday::Adapter::NetHttp.new
  http = adapter.net_http_connection(:url => url, :request => { })

  assert_equal true, http.proxy_from_env?
end
```

because as we said the ENV proxy should be the default one if you don't specify any "proxy" option.
However, this test is passing on Faraday master branch, but not on your branch, which means that your changes are affecting the default behaviour.

I don't think we'll be able to fix this properly for today's release, so I would suggest we delay it to the next one so that you have time to address these issues properly.
Anyway, good work so far and thanks for your contribution 👍

iMacTia:
Proxy management has been completely refactored now (0.12.2) so this PR is not needed anymore

PR #645, opened 2016-12-21
Support additional options for NetHttpPersistent adapter.

jimpo:
The idle_timeout and retry_change_requests request options are documented at http://docs.seattlerb.org/net-http-persistent/Net/HTTP/Persistent.html.


iMacTia:
Hi @jimpo.
I'm sorry but we're not accepting PR of this type as explained on https://github.com/lostisland/faraday/pull/470#issuecomment-254296415.

PR #286, opened 2013-07-12
Add log scrubbing feature

petems:
Added a feature to use a regex parameter to scrub data from logs.

I implemented it in a similar way to what's being done in https://github.com/lostisland/faraday/pull/277

Fixes https://github.com/lostisland/faraday/issues/282 :+1:

Issue #282, opened 2013-07-11
Ability to Scrub Logs


petems:
If you have used faraday to make request that has sensitive data, such as api keys, passwords and the like, it would be nice if there was an option to scrub the logs.

Current behaviour when DEBUG=true:

``` shell
I, [2013-07-11T10:30:54.013810 #64663]  INFO -- : get https://api.supersecretplace.com/get_all?client_id=supersecret_client_key&api_key=supersecret_api_key
D, [2013-07-11T10:30:54.046199 #64663] DEBUG -- request: User-Agent: "Faraday v0.8.7"
I, [2013-07-11T10:30:56.026899 #64663]  INFO -- Status: 200
...
```

Enhancement possible behaviour eg. DEBUG=true, SCRUB=true:

``` shell
I, [2013-07-11T10:30:54.013810 #64663]  INFO -- : get https://api.supersecretplace.com/get_all?client_id=[SCRUBBED]&api_key=[SCRUBBED]
D, [2013-07-11T10:30:54.046199 #64663] DEBUG -- request: User-Agent: "Faraday v0.8.7"
I, [2013-07-11T10:30:56.026899 #64663]  INFO -- Status: 200
...
```

I've got a few ideas, like a flag for the `logger.rb` that gsubs a specific pattern or something of the like?

PR #650, opened 2016-12-29
Added filter method for logger middleware

petems:
Allows you to filter sensitive information with a regex, allowing a gsub with redaction:

```ruby
conn = Faraday.new(:url => 'http://sushi.com/api_key=s3cr3t') do |faraday|
  faraday.response :logger do | logger |
    logger.filter(/(api_key=)(\w+)/,'\1[REMOVED]')
  end
end
```

It's back from the dead! 😄 

For more information see previous threads #286 and #282

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -12,22 +12,28 @@ module Faraday
         require 'logger'
         ::Logger.new(STDOUT)
       end
+      @filter = []
       @options = DEFAULT_OPTIONS.merge(options)
+      yield self if block_given?
     end
 
     def_delegators :@logger, :debug, :info, :warn, :error, :fatal
 
     def call(env)
-      info "#{env.method} #{env.url.to_s}"
-      debug('request') { dump_headers env.request_headers } if log_headers?(:request)
-      debug('request') { dump_body(env[:body]) } if env[:body] && log_body?(:request)
+      info "#{env.method} #{apply_filters(env.url.to_s)}"
+      debug('request') { apply_filters( dump_headers env.request_headers ) } if log_headers?(:request)
+      debug('request') { apply_filters( dump_body(env[:body]) ) } if env[:body] && log_body?(:request)
       super
     end
 
     def on_complete(env)
       info('Status') { env.status.to_s }
-      debug('response') { dump_headers env.response_headers } if log_headers?(:response)
-      debug('response') { dump_body env[:body] } if env[:body] && log_body?(:response)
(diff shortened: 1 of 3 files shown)


dentarg:
I was looking at #286 some when I built my own logger middleware for Faraday (used in a private project unfortunately) and had an idea I wanted to share: also supporting [Procs](https://ruby-doc.org/core-2.3.0/Proc.html) (and/or lambdas) as filters. I think that could be powerful.

Feel free to ignore me if it doesn't sounds interesting. I'm afraid I don't I have any time soon to try it out myself, just wanted to share the idea. :)

petems:
@dentarg Yeah, I ended up making my own logger middleware for filtering a long time ago: https://github.com/petems/faraday-scrublogs. Plus I ended up implementing it directly in the project I needed it for: https://github.com/pearkes/tugboat/blob/3f2f9cfabbd7ae0a1e56a60a52da00324afb56af/lib/tugboat/middleware/custom_logger.rb

lib/faraday/response/logger.rb
```diff
@@ -12,22 +12,28 @@ def initialize(app, logger = nil, options = {})
...
+      debug('request') { apply_filters( dump_body(env[:body]) ) } if env[:body] && log_body?(:request)
       super
     end
 
     def on_complete(env)
       info('Status') { env.status.to_s }
-      debug('response') { dump_headers env.response_headers } if log_headers?(:response)
-      debug('response') { dump_body env[:body] } if env[:body] && log_body?(:response)
+      debug('response') { apply_filters (dump_headers env.response_headers ) } if log_headers?(:response)
+      debug('response') { apply_filters (dump_body env[:body] ) } if env[:body] && log_body?(:response)
```
iMacTia: You'll need to remove spaces between `apply_filters` and the brackets in order to make this compatible with Ruby 1.9.3

petems: Fixed 👍

iMacTia:
Great, nice work, really powerful feature if you know how to use it 😃 
I've marked this for release v0.11.0 so I'll keep the PR open for the time being, but ready to be merged from my point of view

petems:
Yay! Thanks so much to @iMacTia for the merge and @mislav for the original review and help developing the feature 😄

shilpi230:
If we have to remove it from the body can we do it?

iMacTia:
@shilpi230 yes these filters are applied to the following:
* URL
* headers
* body (see https://github.com/lostisland/faraday/blob/main/lib/faraday/logging/formatter.rb#L101)

Issue #648, opened 2016-12-28
Allow adapter-specific customisations 


iMacTia:
Faraday was born as an abstraction layer between developers and their networking adapters. For this reason, when a feature is supported by Faraday, this should be also supported by ALL its adapters.
This allows you to swap your adapter freely without changing your code.

For this reason, the Faraday team is reluctant to the idea of accepting new features supported by only one or two adapters. However, I can see an increase in the number of Issues/PRs that ask for a way to access and customise adapters to support ad-hoc features.

`ActiveRecord` works in a very similar way, providing an abstraction layer with your DB adapter, that you're free to swap anytime, but with some limitations.
AR however gives you the possibility to run arbitrary SQL to query or migrate your DB, even though this could cause you to lose the possibility to use another adapter in the future (unless you change this custom code, of course).

Wanting to do something similar to AR on Faraday, we could allow to interact with the adapter to setup some custom features.
One of the most popular ways to do this, is using the block syntax:

```ruby
conn = Faraday.new(...) do |f|
  f.adapter :net_http_persistent
end

conn.configure_adapter do |adapter|
  adapter.idle_timeout = 100
end
```

This of course requires Faraday to know where the adapter is in the middleware stack, which is currently not possible, but a nice feature I would like to include in v 1.0 making the call to `#adapter` method mandatory while building the stack.

I'm happy to hear the thoughts of the community, and especially the ones from @mislav, or to hear any alternative solution on the above :)

iMacTia:
Example of Issues/PRs that could be addressed by this feature:

#645 #585 #572 #545 #502

mltsy:
I agree.  Restricting the library to be a lowest common denominator of all the supported adapters is an unnecessary cost.  You get the best of both worlds by consolidating any changes or incompatibilities into a configure block for a specific adapter as proposed :+1:

Faraday is actually doing 2 things:
1) Providing an easy, familiar, streamlined interface for creating and reusing HTTP connections, regardless of the underlying adapter (the reason most people use it, I imagine)
[...]

Issue #572, opened 2016-04-02
document how to give options to adapter


alexanderadam:
I'm using the patron adapter and occasionally running into the `Maximum (5) redirects followed` exception.
Patron allows to configure it with `max_redirects` but I'm can't find how to configure it with faraday. :cry:

iMacTia:
Hi @alexanderadam,

I'm in the process of adding the feature you need with #651 
Please have a look at the PR for additional information 😃

iMacTia:
#651 released with v0.11.0
This is the usage in your case:

```ruby
conn = Faraday.new(...) do |f|
  # Insert more middlewares here  
  f.adapter :patron do |session| # yields Patron::Session
    session.max_redirects = 10
  end
end
```

PR #651, opened 2016-12-30
Support adapter customization

iMacTia:
Work in progress that wants to address #648

Usage:

### NetHttp (from #545 and #502)
```ruby
conn = Faraday.new(...) do |f|
  f.adapter :net_http do |http| # yields Net::HTTP
    http.idle_timeout = 100
    http.verify_callback = lambda do | preverify_ok, cert_store |
      # do something here...
    end
  end
end
```

### NetHttpPersistent (from #645, #585 and #470)
```ruby
conn = Faraday.new(...) do |f|
  f.adapter :net_http_persistent do |http| # yields Net::HTTP::Persistent
    http.idle_timeout = 100
    http.retry_change_requests = true
  end
end
```

### Patron (from #572)
```ruby
conn = Faraday.new(...) do |f|
  f.adapter :patron do |session| # yields Patron::Session
    session.max_redirects = 10
  end
end
```

### HTTPClient (from #622)
```ruby
conn = Faraday.new(...) do |f|
  f.adapter :httpclient do |client| # yields HTTPClient
    client.keep_alive_timeout = 20
    client.ssl_config.timeout = 25
  end
end
```

### DIFF
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -30,9 +30,10 @@ module Faraday
     extend Parallelism
     self.supports_parallel = false
 
-    def initialize(app = nil, &block)
-      super
-      @custom_config = block
+    def initialize(app = nil, opts = {}, &block)
+      super(app)
+      @connection_options = opts
+      @config_block = block
     end
 
     def call(env)
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -32,7 +32,7 @@ module Faraday
 
     def initialize(app = nil, &block)
       super
-      @custom_config = block if block_given?
+      @custom_config = block
     end
 
     def call(env)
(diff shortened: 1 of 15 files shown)


iMacTia:
`Excon`, `EMHttp` adapters won't support this feature as they only accept connection configuration when initialised.
For this reason, I'm adding another parameter to the Adapters initialiser that will allow to pass them these optional config (this was already implemented on Excon, I'm just adding it to EMHttp and EMSynchrony as well)

iMacTia:
Done, covered all adapters excluding Thypoeus (will be removed soon) and the test ones.
@mislav, it would be great if you could give it a look.
If everything goes fine, I'd merge and release this with v0.11.0 👍

mltsy:
Wow, awesome work on adding both kinds of configs!  Thanks again for this great update :smiley:

PR #652, opened 2017-01-01
Default options appear overwritten once set on a Connection instance

alexford:
We noticed this when the timeout we set in our project (via `Faraday.default_connection_options.request.timeout = 10` in a Rails initializer), was not being used once another piece of code (in a 3rd party gem) created a Connection with a shorter timeout. Once a shorter timeout is used for a Connection, it seems to become the default for any new Connections that did not explicitly set their timeout.

This PR includes a failing unit test that illustrates the behavior. I am not familiar enough with the internals of Faraday to offer a fix yet, but I'll keep looking.

UPDATE: This bug only appears when the "offending" Connection instance is instantiated with only a Hash parameter to the constructor (i.e. `Faraday.new :url => 'https://foo.com'`).

_For example..._

````ruby
Faraday.default_connection_options.request = 10
conn_a = Faraday.new 'https://foo.com'  # default still 10 as expected

conn_b = Faraday.new 'https://bar.com'
conn_b.options.timeout = 1   # default still 10 as expected

conn_c = Faraday.new :url => 'https://baz.com' # note Hash style
conn_c.options.timeout = 1   # default is now 1!
````

### DIFF
--- a/test/connection_test.rb
+++ b/test/connection_test.rb
@@ -453,6 +453,8 @@ class TestConnection < Faraday::TestCase
   def test_default_connection_options_persist_with_an_instance_overriding
     Faraday.default_connection_options.request.timeout = 10
     conn = Faraday.new 'http://nigiri.com/bar'
+    conn.options.timeout = 1
+    assert_equal 10, Faraday.default_connection_options.request.timeout
 
     other = Faraday.new :url => 'https://sushi.com/foo'
     other.options.timeout = 1
--- a/test/connection_test.rb
+++ b/test/connection_test.rb
@@ -450,6 +450,16 @@ class TestConnection < Faraday::TestCase
     assert_equal 1, conn.options.open_timeout
   end
 
+  def test_default_connection_options_persist_with_an_instance_overriding
+    Faraday.default_connection_options.request.timeout = 10
+    conn = Faraday.new 'http://nigiri.com/bar'
+
+    other = Faraday.new :url => 'https://sushi.com/foo'
+    other.options.timeout = 1
+
+    assert_equal 10, Faraday.default_connection_options.request.timeout
+  end
+
   def env_url(url, params)
     conn = Faraday::Connection.new(url, :params => params)
     yield(conn) if block_given?
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -55,8 +55,7 @@ module Faraday
     #                     :user     - String (optional)
     #                     :password - String (optional)
     def initialize(url = nil, options = nil)
-
-      options = ConnectionOptions.from(options)
+      options = ConnectionOptions.from(options) unless options and options.is_a?(ConnectionOptions)
 
       if url.is_a?(Hash)
         options = options.merge(url)
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -51,7 +51,14 @@ module Faraday
     def merge(value)
       dup.update(value)
     end
+    
+    # Public
+    def dup
+      self.class.from(self)
+    end
 
+    alias clone dup
+    
     # Public
     def fetch(key, *args)
       unless symbolized_key_set.include?(key.to_sym)

### CODE REVIEW

File: test/connection_test.rb [hunk 1]
```diff
@@ -450,6 +450,16 @@ def test_default_connection_options_as_hash_with_instance_connection_options
     assert_equal 1, conn.options.open_timeout
   end
 
+  def test_default_connection_options_persist_with_an_instance_overriding
+    Faraday.default_connection_options.request.timeout = 10
+    conn = Faraday.new 'http://nigiri.com/bar'
+
+    other = Faraday.new :url => 'https://sushi.com/foo'
```
alexford: I have found that this test will pass if the `sushi.com` url here is passed a string argument rather than in this hash form, but fails when passed as seen here.

iMacTia:
Hi @alexford and thanks for your PR!
I think we have a reference issue here, good catch.
I'll try to have a closer look to the code using also your test and see what the issue is.

iMacTia:
@alexford I forgot asking you, can you please confirm me the Faraday version you're using? Is it v0.10.1?

alexford:
Hi @iMacTia —yes we are using v0.10.1. I have merged my co-worker @ctrlShiftBryan's change which appears to fix the issue.

ctrlShiftBryan:
Would a better fix be to use ConnectionOptions.from(default_connection_options) here https://github.com/lostisland/faraday/blob/master/lib/faraday.rb#L66 instead or in addition to the Connection.initialize change?

Cohen-Carlisle:
@ctrlShiftBryan That `.dup` is definitely the problem. It only produces a shallow copy. So when
```ruby
other.options.timeout = 1
```
is run later on, it mutates the `timeout` on the global default `Faraday::RequestOptions` because `Faraday::ConnectionOptions` holds a pointer to it, not a copy.

iMacTia:
Thanks everyone for the input and sorry if I took some time to reply, but I wanted to be sure about what was going on.
So, as @Cohen-Carlisle is correctly pointing out, the issue is due to a, IMHO, wrong implementation of `Faraday::ConnectionOptions#dup`. In fact, this method should create a full copy of the options, but in reality is keeping all the pointers to the sub-options, which allows you to change them through the instance.
I tried overriding `dup` and implement it correctly, but most of tests start failing, so that's not an option, unfortunately.

@alexford and @ctrlShiftBryan solution seems fine to me as it's forcing a proper copy of the options with `ConnectionOptions.from(options)` at the beginning of `Faraday::Connection.new`.
However, there are a couple of small fixes I would ask you to do:

1. Remove the `.dup` call from https://github.com/lostisland/faraday/blob/master/lib/faraday.rb#L66 as it's now useless (@ctrlShiftBryan proposed something similar, I believe we don't need to use `ConnectionOptions.from` in this case)
2. Add the override and check also for `conn` in your test, as you're currently doing it only for `other`

Cohen-Carlisle:
@iMacTia I was able to redefine `ConnectionOptions#dup` as follows to make the tests pass:
```ruby
def dup
  self.class.from(self)
end

alias clone dup
```
And I figured if we redefine `dup` we should get `clone` as well for safety.
Would this solution be preferred?

iMacTia:
Thanks @Cohen-Carlisle !
I just realised the (embarrassing) error I made in my attempt to do the same.
I personally think we should include your fix in this PR as it addresses the issue to the root.
On the other hand, I also like @alexford and @ctrlShiftBryan refactoring of the `Faraday::Connection#initialize`, but I would make the first line optional based on the type of the parameter:

```ruby
options = ConnectionOptions.from(options) unless options and options.is_a?(ConnectionOptions)
```

This is because otherwise we're duplicating the object anyway, even if it's already usable as it is.
Since ``Faraday::Options#dup` is now fully working, we can keep it on https://github.com/lostisland/faraday/blob/master/lib/faraday.rb#L66

So tu summarize, @alexford and @ctrlShiftBryan can you pelase:

1. Add @Cohen-Carlisle fix to `Faraday::Options`
2. Test the options override also for `conn` in your test, as you're currently doing it only for `other`


Happy to merge this in v0.11 if it make it on time 😄

alexford:
@iMacTia your changes are integrated

iMacTia:
Amazing, thanks @alexford for the changes and thanks @Cohen-Carlisle for the `dup` suggestion!
Will pack this in on release v0.11.0 👍

Issue #372, opened 2014-04-30
token auth newline with passenger standalone


madaray:
v0.9.0 /lib/faraday/request/authorization.rb line 20 adds a newline between options 

which works fine on localhost but using nginx it does not. This is from the rails request in the API app - Phusion Passenger version 4.0.41: 

"HTTP_AUTHORIZATION"=>"Token token=\"<token>\",", "HTTP_USER_AGENT"=>"Faraday v0.9.0"

the data after the comma is missing.

This is from the localhost - Thin web server (v1.6.2 codename Doc Brown):

"HTTP_AUTHORIZATION"=>"Token token=\"<token>\",\n                     option_1=\"<option_1_value>\"", "HTTP_USER_AGENT"=>"Faraday v0.9.0"

monkey patching faraday to not use \n fixes the problem. Rails doesn't add the newline when working with ActionController::HttpAuthentication::Token.encode_credentials:

"Token token=\"<token>\", option_1=\"<option_1_value>\"

So I assume that removing it would not make the header invalid?

cavi21:
Hi there! probably I missed something but why this is closed? that `\n` in the **[#build_hash](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/authorization.rb#L20)** method is causing issue between different setups on different servers and not sure the need of it?

Does anyone have any idea why is there? happy to open a PR if it's need it

Have a great day!

iMacTia:
Hi @cavi21, I'm sorry this issue was not addressed before. I suppose @madaray closed it because no one replied in more than 2 years.
I've recently started contributing to Faraday and I'm now trying to address issues more quickly.

That said, I can see the issue here and I agree that headers should not contain any "\n" as that will be read as the end of the header value.
I think the author of that code was just trying to make it more user-friendly to read.

That said, I could review a PR that removes the newlines making all tests still passing.

We have a release (v0.11.0) scheduled for tomorrow, compatibly with your time I could merge this is in time for it.

Please mention this Issue (which I'll now reopen) in your PR

cavi21:
Awesome news @iMacTia I'll open a PR as you said and we will go from there!

Thanks for the prompt reply on this!

PR #657, opened 2017-01-12
Remove the presence of \n on the authorization headers

cavi21:
We talk about this in the issue #372 with @iMacTia.

I change as little as possible in order to prevent issue with older versions of ruby. And make sure that the test are all green (at least in my local machine). 

Let me know if you would like another approach on this. Also I could not found another place where I should change this... if you have an idea or know about other place let me know and I'll change it as well.

Let me know what you think!
Have a great Thursday

### DIFF
--- a/lib/faraday/request/authorization.rb
+++ b/lib/faraday/request/authorization.rb
@@ -16,8 +16,7 @@ module Faraday
 
     # Internal
     def self.build_hash(type, hash)
-      offset = KEY.size + type.size + 3
-      comma = ",\n#{' ' * offset}"
+      comma = ", "
       values = []
       hash.each do |key, value|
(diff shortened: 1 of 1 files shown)


iMacTia:
Thanks @cavi21 
I think it's fine because it's simple.
Tests are green, so I'm happy to merge this in tomorrow for the v0.11.0 release.
Can you confirm using your branch fixes the issues you were experiencing?

cavi21:
Hey @iMacTia sorry for the late reply! yes that change fix the issues I was experiencing in the project I'm working right now.

iMacTia:
Great 👍 ! Thanks for the feedback.
Will merge this in today and release with v0.11.0, good timing 😄

PR #658, opened 2017-01-17
Implement no_proxy feature

somethingnew2-0:
A third stab at supporting the no_proxy environment variable.  Previously we had #361 and #637 which implemented no_proxy using a new Proxy middleware.  Although more modular in design, it runs into the issue where middleware is not typically enabled by default whereas this proxy middleware should be.  I implemented the feature differently by adding onto the existing proxy code instead of creating a middleware.  I reused some code and tests from the previous PRs, so they weren't entirely at a loss.  Feedback and comments welcome.

cc: @olleolleolle

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -79,13 +79,15 @@ module Faraday
       @params.update(options.params)   if options.params
       @headers.update(options.headers) if options.headers
 
-      @proxy = nil
+      @proxy = @no_proxy = nil
       proxy(options.fetch(:proxy) {
         uri = ENV['http_proxy']
         if uri && !uri.empty?
-          uri = 'http://' + uri if uri !~ /^http/i
+          uri = 'http://' + uri unless uri.downcase.start_with?('http')
           uri
         end
+      }, options.fetch(:no_proxy) {
+        ENV['no_proxy']
       })
 
       yield(self) if block_given?
@@ -281,9 +283,10 @@ module Faraday
     end
 
     # Public: Gets or Sets the Hash proxy options.
-    def proxy(arg = nil)
-      return @proxy if arg.nil?
-      @proxy = ProxyOptions.from(arg)
+    def proxy(proxy = nil, no_proxy = nil)
+      return @proxy if proxy.nil?
+      @no_proxy = no_proxy.gsub('*.','').split(',').map(&:downcase) unless no_proxy.nil?
+      @proxy = ProxyOptions.from(proxy)
     end
 
     def_delegators :url_prefix, :scheme, :scheme=, :host, :host=, :port, :port=
@@ -371,6 +374,7 @@ module Faraday
         req.url(url)                if url
         req.headers.update(headers) if headers
         req.body = body             if body
+        req.options.merge!(:proxy => self.proxy) if proxy_allowed?(build_exclusive_url(req.path))
         yield(req) if block_given?
       end
 
@@ -384,8 +388,27 @@ module Faraday
       Request.create(method) do |req|
         req.params  = self.params.dup
         req.headers = self.headers.dup
-        req.options = self.options.merge(:proxy => self.proxy)
+        req.options = self.options
         yield(req) if block_given?
+      end
+    end
+
+    def proxy_allowed?(url)
+      return true if @no_proxy.nil?
+      uri = Utils.URI(url)
+      @no_proxy.none? do |no_proxy_domain|
+        no_proxy_uri = Utils.URI(no_proxy_domain)
+        # assume strings without scheme are of http
+        if no_proxy_uri.class == URI::Generic && !no_proxy_domain.empty?
+          no_proxy_uri = Utils.URI("http://#{no_proxy_domain}")
+        end
+        if no_proxy_uri.port == 80
+          # domain matching is fine
+          uri.host.downcase.end_with?(no_proxy_uri.host)
+        else
+          # need to match ports exactly
+          "#{uri.host.downcase}:#{uri.port}".end_with?("#{no_proxy_uri.host}:#{no_proxy_uri.port}")
+        end
       end
     end
 
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -285,7 +285,9 @@ module Faraday
     # Public: Gets or Sets the Hash proxy options.
     def proxy(proxy = nil, no_proxy = nil)
       return @proxy if proxy.nil?
-      @no_proxy = no_proxy.gsub('*.','').split(',').map(&:downcase) unless no_proxy.nil?
+      @no_proxy = no_proxy
+        .scan(/(?!\.)([^:,\s]+)(?::(\d+))?/)
+        .map {|host, port| [host, port, /(\A|\.)#{Regexp.quote host}\z/i]} unless no_proxy.nil?
       @proxy = ProxyOptions.from(proxy)
     end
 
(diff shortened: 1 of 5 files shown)


somethingnew2-0:
@iMacTia care to take a look?

lib/faraday/connection.rb
```diff
@@ -384,11 +388,30 @@ def build_request(method)
...
+        if no_proxy_uri.class == URI::Generic && !no_proxy_domain.empty?
+          no_proxy_uri = Utils.URI("http://#{no_proxy_domain}")
+        end
+        if no_proxy_uri.port == 80
+          # domain matching is fine
+          uri.host.downcase.end_with?(no_proxy_uri.host)
+        else
+          # need to match ports exactly
+          "#{uri.host.downcase}:#{uri.port}".end_with?("#{no_proxy_uri.host}:#{no_proxy_uri.port}")
+        end
```
reggieb: `uri.port` will default to 80. It would be simpler and remove the need to insert the hosts and ports into strings by just comparing the hosts and ports separately. 

So replace the `if no_proxy_uri.port == 80` statement (lines 405-411) with:

`uri.host.downcase.end_with?(no_proxy_uri.host) && uri.port == no_proxy_uri.port
`

reggieb: Also I think using `end_with?` could be problematic (both "foo" and "oo" end with "oo"). Would this work:

`uri.port == no_proxy_uri.port && uri.host.downcase == no_proxy_uri.host.downcase`

reggieb:
+1 As I've just hit a problem due to Faraday not picking up the no_proxy settings from my environment - I think this would be a worthwhile update.

lib/faraday/connection.rb
```diff
@@ -281,9 +283,10 @@ def in_parallel(manager = nil)
     end
 
     # Public: Gets or Sets the Hash proxy options.
-    def proxy(arg = nil)
-      return @proxy if arg.nil?
-      @proxy = ProxyOptions.from(arg)
+    def proxy(proxy = nil, no_proxy = nil)
+      return @proxy if proxy.nil?
+      @no_proxy = no_proxy.gsub('*.','').split(',').map(&:downcase) unless no_proxy.nil?
```
olleolleolle: nitpick: whitespace missing after comma in gsub call.

iMacTia:
I'm really sorry guys, I'm keeping an eye on this as it really is interesting (we had the same issue some time ago, but lucky enough to be in the position to remove the proxy).
Unfortunately I'm having a busy time in the office but hopefully will be able to have a look at this in the next days.
Apologies

lib/faraday/connection.rb
somethingnew2-0: I copied this comparison code from the previous [PRs](https://github.com/lostisland/faraday/pull/637/files#diff-fc586c64171977c4441ff2d0f8bba0e2R115).

For your first comment, if the port is the default (port 80) we don't want to compare port numbers, but only compare domain names.  Without this the [test_proxy_allowed_when_port_is_not_set_in_no_proxy_list](
https://github.com/lostisland/faraday/pull/658/files#diff-6c3548c0c264b7967fcecf8c452ee633R410) fails.

For your second comment, we need the end_with? because the `*.` was `gsub`ed out of the no_proxy string assuming that we want to disallow any domain ending.  For example, a certain shape company might want to exclude internal domains ending with `*.circle` from being proxied.  They could just add `circle` to their no_proxy environment variable then.  This is how it works with other tools such as `curl`.  I can't find a spec or RFC on no_proxy environment variable usage, but this seems to be the general practice.

reggieb: Ah - OK - so if no port number specified then any port is allowed. That makes sense.

As for the `circle` example - my concern is that other domains that end with "circle" would also be blocked, for example `semicircle`. That is, a test like this is needed:

```ruby
def test_proxy_allowed_when_only_end_of_domain_matches
  with_env 'http_proxy' => 'http://proxy.com', 'no_proxy' => 'circle.com' do
    conn = Faraday::Connection.new
    assert_equal conn.proxy_allowed?('http://some.semicircle.com'), true
  end
end
```

somethingnew2-0: Yeah, that would make sense.  Perhaps I should use this code instead?
https://github.com/ruby/ruby/blob/trunk/lib/uri/generic.rb#L1528-L1545

somethingnew2-0: Ok, I updated it to use that code which fixes the semicircle.com issue discussed. I also included a test for it.

test/connection_test.rb
```diff
@@ -369,6 +376,86 @@ def test_proxy_requires_uri
...
+  end
+
+  def test_proxy_allowed_when_url_is_not_in_no_proxy_list_but_ip_address_is
+    with_env 'http_proxy' => 'http://proxy.com', 'no_proxy' => '127.0.0.1' do
+      conn = Faraday::Connection.new
+      assert_equal conn.proxy_allowed?('http://localhost'), false
+    end
+  end
+
+  # def test_proxy_allowed_when_url_is_not_in_no_proxy_list_but_ip_address_is_public
```
reggieb: Is there a good reason to leave this commented out test in the code? I'd suggest it is removed.

reggieb:
Looks good to ship

test/connection_test.rb
somethingnew2-0: Removed

somethingnew2-0:
Should I bump the gem version?

lib/faraday.rb
```diff
@@ -14,7 +14,7 @@
 #   conn.get '/'
 #
 module Faraday
-  VERSION = "0.11.0"
+  VERSION = "0.11.1"
```
iMacTia: No need to bump the version, I'll bump it before releasing the gem next time 😄

somethingnew2-0: Ah ok, I removed the commit to bump the version. 👍

test/env_test.rb
```diff
@@ -72,11 +72,6 @@ def test_request_create_stores_ssl_options
     assert_equal false, env.ssl.verify
   end
 
-  def test_request_create_stores_proxy_options
-    env = make_env
-    assert_equal 'proxy.com', env.request.proxy.host
-  end
```
somethingnew2-0: This test was removed because of where (and how) the proxy options are merged into the request object now.  The `make_env` method apart of this test class calls `connection.build_request` to build the request.  This PR changes the code so that t[he proxy options are merged later](https://github.com/somethingnew2-0/faraday/blob/6bc251833b5c837825aaa4055b19271d4d212d47/lib/faraday/connection.rb#L379) in the `connection.run_request` method instead because we need the destination url in order to compare to the `no_proxy` domain list.

```
def make_env(method = :get, connection = @conn, &block)
  request = connection.build_request(method, &block)
  request.to_env(connection)
end
```

iMacTia: Makes sense then.
I'd assume the coverage of this test have been moved to the new tests so happy to stay without it 👍 
Thanks for the clarification 😄

iMacTia:
Really happy to get this merged in.
Thanks @somethingnew2-0 for the implementation, but also thanks to @johnbellone and @olleolleolle for attempting the same in the past contributing with useful feedbacks 😄 !
If you want, feel free to join the new Gitter channel: https://gitter.im/lostisland/faraday

somethingnew2-0:
Thanked so much!

thojansen:
@iMacTia any timeline for releasing 0.12.0?

iMacTia:
There was an issue claiming that the `no_proxy` feature was not working, so I delayed the release of 1 week. Is currently scheduled for next Friday 😃

iMacTia:
Hello everyone,

just a quick update on the `no_proxy` implementation.
Even though I liked the work done by @somethingnew2-0 in this PR, I happily accepted an updated and alternative implementation provided by @authorNari in #670.
I would just like to share with you the main reasons for this decision:

1. Delegation to Ruby: delegating the issue to Ruby, we don't have to provide (and maintain in the future!) an ad-hoc implementation of proxy/no_proxy variables.
2. Simplified options: you probably noticed the no_proxy option introduced in this PR have been removed in #670. The reason is that if you manually set the `proxy` option then the `no_proxy` option is redundant.

I hope the above makes sense and I wanted to make it clear that we don't value contributions more than others, all we want is to find the easiest and flexible solution to problems :)

thojansen:
I do not fully get point 2, since usually you set `proxy` and `no_proxy` at the same time to have a default routing through the proxy and exclude certain domains matching `no_proxy` rules (network separation). But delegation of this functionality makes a lot of sense.

iMacTia:
Right, I was probably thinking in a too much simplistic way but when you initialise a Faraday Connection you specify the base_url so you basically already know if you'll need the proxy or not.

If you want all your connections to pass through the proxy (but for some reason you can't rely on the ENV variable) then you can use `default_connection_options`.

There are probably some shady edge cases that we're not properly covering but considering that we now fully support proxy/no_proxy env variables (most of cases) and we allow to manually set the proxy either per-connection or as a default option (some cases), I would say we're already covering enough (if not all) cases without going crazy with custom implementations.

The only case this would probably get harder to manage is when you can't use proxy/no_proxy and you need to set a default value for both proxy AND no_proxy. But is it worth having complex implementations and keeping maintaining them only for this case?

Of course we'll listen to the community and if we'll see this becoming a big issue then we'll eventually rethink it, but right now I don't see it as a worthy feature to have.

Issue #551, opened 2015-12-16
Improve documentation


YayConnolly:
Suggest improving the documentation to clarify the connection options and middleware API. For example, it seems like there are at least 3 ways to change default parameter encoding: `conn.request :url_encoded`, `conn.options.params_encoder = CustomeEncoder` and inline `Faraday.new params_encoder: CustomEncoder`. As a new user I found this difficult to piece together.

YayConnolly:
@GustavoCaso maybe you would be interested in helping on this one :)

GustavoCaso:
Of course @YayConnolly I'll have a look

mislav:
Thanks for the suggestion! It's important to distinguish that `conn.request :url_encoded` does not in fact change default parameter encoding, but mounts the UrlEncoded middleware that serializes raw hash POST data into `x-www-form-urlencoded`. The parameter encoder does affect the encoding that happens within, but primarily affects the handling of GET parameters throughout the request cycle.

GustavoCaso:
Thanks @mislav I was looking around the code to understand it better, before writing any documentation, so I don't mess up the `README`

iMacTia:
This should have been addressed in #659

PR #659, opened 2017-01-18
Issue #551 Revise Readme to be more accessible

bcgoss:
I have added some language to README.md which should make the document more accessible to people who are not familiar with Faraday. I am happy to make further adjustments based on your feedback. Thanks for taking a look!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ conn = Faraday.new(:url => 'http://www.example.com')
 response = conn.get '/users'                 # GET http://www.example.com/users' 
 ```
 
-Connections can also take an options hash as a parameter or be configured by using a block. Checkout the section called [#Advanced-middleware-usage] for more details about how to use this block for configurations. 
+Connections can also take an options hash as a parameter or be configured by using a block. Checkout the section called [Advanced middleware usage](#advanced-middleware-usage) for more details about how to use this block for configurations. 
 
 ```ruby
 conn = Faraday.new(:url => 'http://sushi.com') do |faraday|
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ conn = Faraday.new(:url => 'http://www.example.com')
 response = conn.get '/users'                 # GET http://www.example.com/users' 
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,23 @@ Available at [rubydoc.info](http://www.rubydoc.info/gems/faraday).
...
+response = Faraday.get 'http://sushi.com/nigiri/sake.json'
+```
+A simple `get` request can be performed by using the syntax described above. This works if you don't need to set up anything; you can roll with just the default middleware
+stack and default adapter (see [Faraday::RackBuilder#initialize](https://github.com/lostisland/faraday/blob/master/lib/faraday/rack_builder.rb)).
+
+A more flexible way to use Faraday is to start with a Connection object. If you want to keep the same defaults, you can use this syntax:
+
+```ruby
+conn = Faraday.new(:url => 'http://www.example.com') 
+resposne = conn.get '/users'                 # GET http://www.example.com/users'
```
iMacTia: small typo

README.md
```diff
@@ -23,6 +23,23 @@ Available at [rubydoc.info](http://www.rubydoc.info/gems/faraday).
...
+stack and default adapter (see [Faraday::RackBuilder#initialize](https://github.com/lostisland/faraday/blob/master/lib/faraday/rack_builder.rb)).
+
+A more flexible way to use Faraday is to start with a Connection object. If you want to keep the same defaults, you can use this syntax:
+
+```ruby
+conn = Faraday.new(:url => 'http://www.example.com') 
+resposne = conn.get '/users'                 # GET http://www.example.com/users' 
+```
+
+Connections can also take an options hash as a parameter or be configured by using a block. Checkout the section called Advanced middleware usage for more details about how to use this block for configurations.
```
iMacTia: Can you please insert an anchor link to the Advanced middleware usage section?

bcgoss:
Ok, I've made the suggested change. Let me know if you have any more, or other documentation you want another person looking at.

iMacTia:
Looks good to me 👍 
Thanks

PR #662, opened 2017-01-27
Options deep merge

Cohen-Carlisle:
As noted in https://github.com/lostisland/faraday/pull/629, deep merging of Options was not working.
This PR adds deep merging of Options.
Note that this does result in a new behavior in terms of deep merging nested nil options.
Previously, the nested instance options were completely overridden, allowing it to look like you could nil out an option on merge.

```ruby
    Faraday.default_connection_options.request.timeout = 10
    conn = Faraday.new 'http://sushi.com/foo', request: { timeout: nil }
    conn.options.timeout # => nil
```
But what was actually happening was the nested option was being completely replaced:

```ruby
    Faraday.default_connection_options.request.timeout = 10
    Faraday.default_connection_options.request.open_timeout = 1
    conn = Faraday.new 'http://sushi.com/foo', request: { timeout: nil }
    conn.options.timeout # => nil
    conn.options.open_timeout # => nil
```
Due to the underlying implementation of Options, merges will not replace any value with nil. Hence, the first code example now results in this:

```ruby
    Faraday.default_connection_options.request.timeout = 10
    conn = Faraday.new 'http://sushi.com/foo', request: { timeout: nil }
    conn.options.timeout # => 10
```

Note this is actually closer to the behavior of non-nested options currently:

```ruby
Foo = Faraday::Options.new(:a, :b)
Foo.new(1, nil).merge(Foo.new(nil, 2) # => #<Foo a=1, b=2>
```

### DIFF
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -67,11 +67,9 @@ module Faraday
     end
 
     # Public
-    def dup
+    def deep_dup
       self.class.from(self)
     end
-
-    alias clone dup
 
     # Public
     def fetch(key, *args)
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -18,17 +18,16 @@ module Faraday
     # Public
     def update(obj)
       obj.each do |key, value|
-        if sub_options = self.class.options_for(key)
-          value = sub_options.from(value) if value
-        elsif Hash === value
-          hash = {}
-          value.each do |hash_key, hash_value|
-            hash[hash_key] = hash_value
-          end
-          value = hash
+        sub_options = self.class.options_for(key)
+        if sub_options
+          new_value = sub_options.from(value) if value
+        elsif value.is_a?(Hash)
+          new_value = value.dup
+        else
+          new_value = value
         end
 
-        self.send("#{key}=", value) unless value.nil?
+        self.send("#{key}=", new_value) unless new_value.nil?
       end
       self
     end
@@ -51,14 +50,14 @@ module Faraday
     def merge(value)
       dup.update(value)
     end
-    
+
     # Public
     def dup
       self.class.from(self)
     end
 
     alias clone dup
-    
+
     # Public
     def fetch(key, *args)
       unless symbolized_key_set.include?(key.to_sym)
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -49,13 +49,7 @@ module Faraday
       other.each do |key, other_value|
         self_value = self.send(key)
         sub_options = self.class.options_for(key)
-        if sub_options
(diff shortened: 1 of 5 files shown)


iMacTia:
Thanks @Cohen-Carlisle for trying to fix this!
I think it has been there for too long therefore I'm happy you're spending some time to get it done properly.
As I said on other PRs, I'm a bit busy in this period, so can't promise to come back soon to you.
Thanks for your patience and your contribution

lib/faraday/options.rb
```diff
@@ -48,17 +45,32 @@ def clear
...
 
     # Public
-    def merge(value)
-      dup.update(value)
+    def merge!(other)
+      other.each do |key, other_value|
+        self_value = self.send(key)
+        sub_options = self.class.options_for(key)
+        if sub_options
+          self_value = sub_options.from(self_value)
```
iMacTia: Can't we assume self_value will always be a subclass of `Options` and thus doesn't require to be passed through `#from` class method?
If not, could it be an hash? If so, then we should manage also that case with an `elsif` as we do in `update`.

lib/faraday/options.rb
```diff
@@ -48,17 +45,32 @@ def clear
...
+
+    # Public
+    def merge(other)
+      dup.merge!(other)
     end
-    
+
     # Public
-    def dup
+    def deep_dup
```
iMacTia: Where is this used? Shouldn't we use this instead of `dup` in `merge`?

Cohen-Carlisle: The deep dup is no longer used after some of the refactoring I did, but I kept it in since it could be useful in the future. I can remove if having "dead" code is not desired.

As to why we can just use dup, it's because `merge!` (bang) recursively calls `merge` (non-bang). Hence, every level will be shallow `dup`ed, which is what a deep_dup would do anyway.

lib/faraday/options.rb
Cohen-Carlisle: Good point, it should always be a subclass of options at this point.
In fact, we don't even need to `from` the `other_value`, as `each` behaves the same for options and hashes.

Cohen-Carlisle:
@iMacTia I added the test and addressed your questions. Let me know if you'd like further changes or information! Thanks!

iMacTia:
Looks great to me 👍 !
Thanks for the changes @Cohen-Carlisle. There's only a little conflict now as I just merged another PR addressing the same portion of code. It should be really easy to fix, after that I'm happy to merge this in 😃

Cohen-Carlisle:
@iMacTia merged master. :tada:

iMacTia:
Great job, merged and ready to go with next release 😃

PR #664, opened 2017-02-09
Update README.md

v-kolesnikov:
- Add rubygems.org badge
 - Replace Travis CI PNG badge to SVG


iMacTia:
Thanks for the suggestions @v-kolesnikov!
I was still experimenting with badges as I just added them a couple of days ago.
I've added your changes directly so that I could have a preview of the result, so we can close this PR.
Feel free to open an issue if you have any other suggestion for some useful badge that you would like to see in our Readme 😃

PR #670, opened 2017-02-24
Support no_proxy via URI::Generic#find_proxy

authorNari:
Default proxy value is set via `URI::Generic#find_proxy` if it's defined.
Maintenance cost is lower using upstream implementation than a implementation is made by myself.

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -81,11 +81,20 @@ module Faraday
 
       @proxy = nil
       proxy(options.fetch(:proxy) {
-        uri = ENV['http_proxy']
-        if uri && !uri.empty?
-          uri = 'http://' + uri if uri !~ /^http/i
-          uri
+        uri = nil
+        if URI.parse("").respond_to?(:find_proxy)
+          case url
+          when String
+            uri = URI.parse(url).find_proxy
+          when URI
+            uri = url.find_proxy
+          when nil
+            uri = find_default_proxy
+          end
+        else
+          uri = find_default_proxy
         end
+        uri
       })
 
       yield(self) if block_given?
@@ -432,6 +441,17 @@ module Faraday
       header = Faraday::Request.lookup_middleware(header_type).
         header(*args)
       headers[Faraday::Request::Authorization::KEY] = header
+    end
+
+    def find_default_proxy
+      warn 'no_proxy is unsupported' if ENV['no_proxy'] || ENV['NO_PROXY']
+      uri = ENV['http_proxy']
+      if uri && !uri.empty?
+        uri = 'http://' + uri if uri !~ /^http/i
+        return uri
+      else
+        return nil
+      end
     end
   end
 end
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -448,9 +448,7 @@ module Faraday
       uri = ENV['http_proxy']
       if uri && !uri.empty?
         uri = 'http://' + uri if uri !~ /^http/i
-        return uri
-      else
-        return nil
+        uri
       end
     end
   end
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -79,15 +79,13 @@ module Faraday
       @params.update(options.params)   if options.params
       @headers.update(options.headers) if options.headers
 
-      @proxy = @no_proxy = nil
+      @proxy = nil
       proxy(options.fetch(:proxy) {
         uri = ENV['http_proxy']
         if uri && !uri.empty?
-          uri = 'http://' + uri unless uri.downcase.start_with?('http')
+          uri = 'http://' + uri if uri !~ /^http/i
           uri
         end
-      }, options.fetch(:no_proxy) {
-        ENV['no_proxy']
       })
 
       yield(self) if block_given?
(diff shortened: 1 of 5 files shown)


lib/faraday/connection.rb
```diff
@@ -464,5 +442,16 @@ def set_authorization_header(header_type, *args)
...
+
+    def find_default_proxy
+      warn 'no_proxy is unsupported' if ENV['no_proxy'] || ENV['NO_PROXY']
+      uri = ENV['http_proxy']
+      if uri && !uri.empty?
+        uri = 'http://' + uri if uri !~ /^http/i
+        return uri
+      else
+        return nil
+      end
```
olleolleolle: Code note: drop the `return` keyword on lines 451 and 453 (the `if`'s the last expression of the method).

iMacTia:
Hi @authorNari,

I can see many tests have been deleted in your PR.
Can I ask you why?

authorNari:
Hi @iMacTia 

511a5b4 is reverted on this pull request so these tests are deleted.
These tests are depends on `proxy_allowed?` method.

test/connection_test.rb
```diff
@@ -326,126 +319,71 @@ def test_proxy_accepts_hash
...
-      conn.proxy :uri => :bad_uri, :user => 'rick'
-    end
-  end
-
-  def test_proxy_allowed_when_url_in_no_proxy_list
-    with_env 'http_proxy' => 'http://proxy.com', 'no_proxy' => 'example.com' do
-      conn = Faraday::Connection.new
-      assert_equal conn.proxy_allowed?('http://example.com'), false
-    end
-  end
```
iMacTia: Please reintroduce all the deleted test as they're covering some important edge-cases.
This can be done without using the `proxy_allowed?` method:

- If the test asserts `assert_equal conn.proxy_allowed?('http://example.com'), false`, then you can change it into `assert_nil conn.proxy`
- If the test asserts `assert_equal conn.proxy_allowed?('http://prefixedexample.com'), true` than you can change it into `assert_equal 'proxy.com', conn.proxy.host`

lib/faraday/connection.rb
```diff
@@ -79,15 +79,22 @@ def initialize(url = nil, options = nil)
...
 
-      @proxy = @no_proxy = nil
+      @proxy = nil
       proxy(options.fetch(:proxy) {
-        uri = ENV['http_proxy']
-        if uri && !uri.empty?
-          uri = 'http://' + uri unless uri.downcase.start_with?('http')
-          uri
+        uri = nil
+        if URI.parse("").respond_to?(:find_proxy)
```
iMacTia: According to the documentation, `find_proxy` is supported in Ruby 1.9.3 (https://ruby-doc.org/stdlib-1.9.3/libdoc/open-uri/rdoc/URI/Generic.html). Since we dropped the support for older ruby version, I think this method should always be available? Do you get any failing test if you remove this check?

lib/faraday/connection.rb
iMacTia: I agree. Moreover, the `else` branch is completely redundant. so the `if` can be simplified into:

```ruby
if uri && !uri.empty?
  uri = 'http://' + uri if uri !~ /^http/i
  uri
end
```

test/connection_test.rb
authorNari: That make sense. done bbd89c1

lib/faraday/connection.rb
authorNari: >  Since we dropped the support for older ruby version, I think this method should always be available?

`open-uri` lib has `find_proxy` in >= Ruby 1.8.7. This method move to `uri/generic` lib in Ruby 2.0.0.
I think any library shouldn't require `open-uri` because this library overwrite Kernel#.open.
So this pull request doesn't support no_proxy in Ruby 1.9.3. 😢

lib/faraday/connection.rb
authorNari: I think If the return value of a method is meaningful, `return` should be explicitly written. It's easy to read :)

But I think that the consistency of this project should be respected.
I deleted `return` dd13327

lib/faraday/connection.rb
iMacTia: Thanks for the detailed response @authorNari that makes totally sense.
I have nothing against this as the plan is to move away from Ruby1.9.3 as soon as possible, so this can be a good incentive!

iMacTia:
Thank you @authorNari, outstanding work!
I'm grateful we could delegate to Ruby the management of the proxy/no_proxy ENV variables check.
Apologies if it took me some time to review and thanks again for the valuable contribution.

lib/faraday/connection.rb
```diff
@@ -464,5 +442,14 @@ def set_authorization_header(header_type, *args)
         header(*args)
       headers[Faraday::Request::Authorization::KEY] = header
     end
+
+    def find_default_proxy
+      warn 'no_proxy is unsupported' if ENV['no_proxy'] || ENV['NO_PROXY']
```
somethingnew2-0: So no_proxy is unsupported when the url is not given upfront in the initializer, but later on in the request?

iMacTia: You mean when you initialise the connection simply as `conn = Faraday.new` ?
That is actually true, I missed this detail as it's an unusual way to use it.
This could be easily fixed moving the #find_proxy call on the `run_request` or `build_request` methods.

somethingnew2-0: Yes, I agree it is a weird way to use it, but it is allowed.  The find_proxy method should be moved to when the request is created in `build_request`.

iMacTia: Totally agree with you, I'll try to squeeze this small fix in release 0.12.0 👍

PR #671, opened 2017-02-26
Prevent logger from dumping request and response headers if they are empty 

danila:
Adds additional checks for presence of collection of headers before attempting to iterate through it.

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -21,15 +21,23 @@ module Faraday
 
     def call(env)
       info "#{env.method} #{apply_filters(env.url.to_s)}"
-      debug('request') { apply_filters( dump_headers env.request_headers ) } if log_headers?(:request)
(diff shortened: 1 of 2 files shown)


iMacTia:
Hi @danila and thanks for your PR.
Is there any particular reason you're checking for headers presence? Did you get in the situation where the middleware failed because they were empty?

danila:
Hello, @iMacTia!
An error occurred when I was testing my app. In fact, I don't think that such case can be reproduced in real-life production mode because every HTTP request must contain a specific set of headers. So maybe this PR is a useless garbage.

iMacTia:
I wouldn't go that far as to call it "useless garbage" 😅, but my point was that I don't think under normal circumstances you'll ever have this issue. So I hope you'll understand I'm a bit reluctant in getting the code more complex to manage such an edge case (if not impossible).
I'm going to close this for now and eventually reconsider if anyone have real-life situations where this becomes an issue.

Thanks!

PR #673, opened 2017-03-01
Make multipart boundary unique

### LINKED ISSUE #672: Multipart uploads containing -------------RubyMultipartPost--\n\r fail
I'm not sure if the string `-------------RubyMultipartPost--` existed before faraday came along but it [can be observed](https://www.google.com/webhp?sourceid=chrome-instant&rlz=1C5CHFA_enUS690US690&ion=1&espv=2&ie=UTF-8#q=%22-------------RubyMultipartPost--%22&*) in nature today.

Because faraday's multipart middleware uses a static boundary string an occurrence of that string in the uploaded file will cause the server truncate the post body prematurely.

This happens more than you might expect, especially given the widespread adoption of faraday itself.

hawknewton:
Fixes #672.

I use `SecureRandom.hex` to generate a unique multipart mime boundary suffix.

### DIFF
--- a/lib/faraday/request/multipart.rb
+++ b/lib/faraday/request/multipart.rb
@@ -1,13 +1,14 @@
 require File.expand_path("../url_encoded", __FILE__)
+require 'securerandom'
 
 module Faraday
   class Request::Multipart < Request::UrlEncoded
     self.mime_type = 'multipart/form-data'.freeze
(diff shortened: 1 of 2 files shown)


iMacTia:
As far as I know boundary can be anything so I'm happy to merge this in knowing that it was giving some issues.
Thanks @hawknewton for the PR

Issue #608, opened 2016-09-20
Remove the legacy Typhoeus adapter


PragTob:
edit: see discussions, basically the adapter is old and you need to load the one from typhoeus

For all that I can tell Faraday has gotten incompatible with new typhoeus/ethon versions (our particular combination was 0.8 and 0.9, respectively)

The concrete error was:

```
ETHON: Libcurl initialized

[Ethon::Errors::InvalidOption] The option: disable_ssl_peer_verification is invalid.
Please try ssl_verifypeer instead of disable_ssl_peer_verification. ({:scheme=>"http", :user=>nil, :password=>nil, :host=>"localhost", :path=>"", :port=>9200, :protocol=>"http"})
Ethon::Errors::InvalidOption: The option: disable_ssl_peer_verification is invalid.
Please try ssl_verifypeer instead of disable_ssl_peer_verification.
from /some_direcctory/gems/ethon-0.9.0/lib/ethon/easy.rb:238:in `block in set_attributes'
```

The offending code is here in the [typhoeus adapter](https://github.com/lostisland/faraday/blob/3e5e1dd5af8b3c1e195ce04ea98630b221bb17c5/lib/faraday/adapter/typhoeus.rb#L39-L43):

``` ruby
        req = ::Typhoeus::Request.new env[:url].to_s,
          :method  => method,
          :body    => env[:body],
          :headers => env[:request_headers],
          :disable_ssl_peer_verification => (env[:ssl] && env[:ssl].disable?)
```

Thanks for making faraday!

mislav:
The Typhoeus adapter shipping with Faraday was only ever [compatible with ancient Typhoeus version 0.3](https://github.com/lostisland/faraday/blob/3e5e1dd5af8b3c1e195ce04ea98630b221bb17c5/Gemfile#L22). Subsequent versions of Typhoeus ship with their own Faraday adapter and it was their responsibility to ship a compatible adapter. That's why we stopped maintaining the Typhoeus adapter within Faraday core.

If there is a problem with the adapter, first make sure that the [adapter is loaded from Typhoeus own code](https://github.com/typhoeus/typhoeus/blob/181834e2483d392d0f7ab0cd17f544cd252c7b2f/lib/typhoeus/adapters/faraday.rb#L9-L15), and then if there's still a problem, report it to their repo.

PragTob:
@mislav thanks for the hint didn't know that, in this case I think it'd be good to remove the ancient adapter :)

[...]

PR #674, opened 2017-03-02
Rounds the result of casting timeout from seconds to ms

ebenoist:
* Prevents a warning in Typhoeus

Currently passing a float value to timeout in Farady (ie. 0.5) results in Typhoeus being passed 500.0 as the timeout in milliseconds. This forces Typhoeus to round the result so it can safely pass an integer to libCURL. 

This fix prevents this issue.

https://github.com/typhoeus/typhoeus/blob/master/lib/typhoeus/easy_factory.rb#L107-L112

Happy to write a spec as well if you can point me in the correct direction.

### DIFF
--- a/lib/faraday/adapter/typhoeus.rb
+++ b/lib/faraday/adapter/typhoeus.rb
@@ -101,8 +101,8 @@ module Faraday
 
       def configure_timeout(req, env)
         env_req = request_options(env)
-        req.timeout = req.connect_timeout = (env_req[:timeout] * 1000) if env_req[:timeout]
-        req.connect_timeout = (env_req[:open_timeout] * 1000)          if env_req[:open_timeout]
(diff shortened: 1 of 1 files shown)


iMacTia:
Hi @ebenoist,
I know this will sound confusing, but our Typhoeus adapter have not been updated in ages so it's not compatible with latest versions.
You should use the Faraday adapter provided in the gem itself (https://github.com/typhoeus/typhoeus/blob/master/lib/typhoeus/adapters/faraday.rb).

Apologies for the confusion, that adapter should not suffer the same issues and be compatible with the latest version of the gem.

ebenoist:
Got it, thank you for the explanation. If this adapter is no longer supported should it be removed?

iMacTia:
Yes indeed!
That's already in the pipeline for v1.0 of Faraday (and discussed in #608), so hopefully soon we'll get rid of it 😃

Issue #667, opened 2017-02-22
Bypass proxy using no_proxy not working 


nisharanij-cuelogic:
I have http_proxy, no_proxy set in my environment.
When I have mentioned my host inside no_proxy faraday unable to bypass that host, it routes through http_proxy which should not
I am using Faraday version 0.11.0. I observed that changes related to proxy are there in master branch but not there in latest faraday version. Is it possible to merge  latest code with version or is there any work around in current 0.11.0 version?

iMacTia:
Hi @nisharanij-cuelogic,

you're right, `no_proxy` option was just recently merged in and I initially planned to release it last Friday but I was delayed.
I will try to release it as soon as possible, until then I suggest you to use the master branch from Github as a temporary solution!
If you could give us a feedback on the `no_proxy` feature before I release it, that would be very appreciated!

In case you don't know how to do it, this is how you you use a gem from Github on bundler:

```ruby
gem 'faraday', github: 'lostisland/faraday'
```

nisharanij-cuelogic:
Added gem 'faraday', github: 'lostisland/faraday' to Gemfile and performed bundle, Still request goes to proxy

iMacTia:
Can I ask you the exact values of your `http_proxy` and `no_proxy` ENV variables?
Is this configuration working in any other application?

nisharanij-cuelogic:
export http_proxy=http://username:password@ip:port
export no_proxy=myhost, myhost:port

when I tried to request my host using curl http://myhost:port/ it's working fine but when tried using faraday it's not

iMacTia:
Ok thanks @nisharanij-cuelogic, I'll ask you to do a couple of additional tests:

1. Try to remove all the spaces from the `no_proxy` ENV variable, especially after each coma
2. Try to open the rails console and check what you see under `ENV['http_proxy']` and `ENV['no_proxy']` before running `Faraday.new(:url => 'http://api:81')`.

nisharanij-cuelogic:
In rails console
irb(main):003:0> ENV['http_proxy']   
=> "http://username:password@ip:port"
irb(main):004:0> ENV['no_proxy']  
=> "127.0.0.1,localhost,127.0.0.1:9506,api:81,api"

Still no luck

iMacTia:
Thanks for the details @nisharanij-cuelogic.
I'm on holiday until mid next week so not sure I can help before then.
[...]

PR #675, opened 2017-03-03
Update connection.rb

jadhavnisha25:
Proxy bypass fixes


iMacTia:
Hi @jadhavnisha25,
as I said in #667, we're not accepting PR against 0.8 or 0.9 branches for new features.
We kept those branches only to deliver critical hotfix updates.
Support for no_proxy have already been added to Faraday from version 0.11.0, so I strongly suggest you update your Faraday version if you need that feature.

PR #677, opened 2017-03-05
fix typo

yuzutas0:
change: from `ElasticSearch` to `Elasticsearch`.

refs. 
- Elasticsearch Reference: master
- https://www.elastic.co/guide/en/elasticsearch/reference/index.html

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -121,7 +121,7 @@ module Faraday
     #   conn.get '/items', {:page => 1}, :accept => 'application/json'
     #   conn.head '/items/1'
     #
-    #   # ElasticSearch example sending a body with GET.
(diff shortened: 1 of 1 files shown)


iMacTia:
Hi @yuzutas0, thank you for your PR.

When this file was first written (5 years ago), the correct wording was `ElasticSearch`, but the guys from Elastic went for a rebranding in 2014 (you can see it here: https://github.com/elastic/elasticsearch/commit/fa16969360ea43667ad11f827bbf86718c18fcc6).
I don't think a comment with the old naming convention is creating any issue, I'm sure you will find thousands of repositories on Github with the same "issue".

Therefore I'm closing this because we only want to accept contributions that are sensibly improving the usefulness of Faraday (e.g. new features or documentation updates).

PR #681, opened 2017-04-04
Parse headers from aggregated proxy requests/responses

ajroetker:
Prior to this commit when connecting to a proxy with auth, faraday
adapters would cause faraday to fail by first trying to connect without
auth and then connection again with auth, faraday could not handle the
response headers from the aggregate response of the two requests.
This commit addresses the above issue by changing faradays parse
function for headers to handle the case where there are multiple
repsonse headers by attempting to parse the last set of response
headers.

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -95,7 +95,13 @@ module Faraday
 
       def parse(header_string)
         return unless header_string && !header_string.empty?
-        header_string.split(/\r\n/).
+
+        headers = header_string.split(/\r\n/)
+
+        # Break headers into distinct responses, and only consider the last response.
+        last_response = headers.slice_before(/^HTTP\//).to_a.last
+
+        last_response.
           tap  { |a| a.shift if a.first.index('HTTP/') == 0 }. # drop the HTTP status line
           map  { |h| h.split(/:\s*/, 2) }.reject { |p| p[0].nil? }. # split key and value, ignore blank lines
           each { |key, value|
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -98,8 +98,9 @@ module Faraday
 
         headers = header_string.split(/\r\n/)
 
-        # Break headers into distinct responses, and only consider the last response.
-        last_response = headers.slice_before(/^HTTP\//).to_a.last
+        # Find the last set of response headers.
+        start_index = headers.rindex { |x| x.match(/^HTTP\//) }
+        last_response = headers.slice(start_index, headers.size)
 
         last_response.
           tap  { |a| a.shift if a.first.index('HTTP/') == 0 }. # drop the HTTP status line
--- a/test/utils_test.rb
+++ b/test/utils_test.rb
@@ -1,6 +1,25 @@
 require File.expand_path('../helper', __FILE__)
 
 class TestUtils < Faraday::TestCase
+
+  # Headers parsing
+
+  def test_headers
+    "HTTP/1.x 500 OK\r\nContent-Type: text/html; charset=UTF-8\r\n" \
+    "HTTP/1.x 200 OK\r\nContent-Type: application/json; charset=UTF-8\r\n\r\n"
+  end
+
+  def test_headers_parsing_for_aggregated_responses
+    headers = Faraday::Utils::Headers.new()
+    headers.parse(test_headers)
+
+    result = headers.to_hash
+
+    assert_equal result["Content-Type"], "application/json; charset=UTF-8"
+  end
+
+  # URI parsing
+
   def setup
     @url = "http://example.com/abc"
   end


ajroetker:
Not sure what's up with 1.9.3 tests, there don't appear to be any related warnings when I run the test script invocation locally.

iMacTia:
Hi @ajroetker and thanks for the PR.
I can see there are issue with ruby 1.9.3 related to the new `slice_before` method.
Unfortunately I'm not really sure of where the issue lies as the warning seems related to something outside Faraday.
I'll need more time to investigate, or maybe you can get the same result without using `slice_before`?

iMacTia:
Also, could you provide a test case to demonstrate the issue you described above?
Thanks

ajroetker:
@iMacTia Thanks for the review, I added a test case and removed the usage of slice_before

lib/faraday/utils.rb
```diff
@@ -95,7 +95,15 @@ def to_hash() ::Hash.new.update(self) end
...
       def parse(header_string)
         return unless header_string && !header_string.empty?
-        header_string.split(/\r\n/).
+
+        headers = header_string.split(/\r\n/)
+
+        # Find the last set of response headers.
+        last_response = headers.reverse.take(
+          headers.reverse.find_index { |x| x.match(/^HTTP\//) } + 1
+        ).reverse
```
iMacTia: The above is extremely inefficient.
Please change it to:

```ruby
start_index = headers.rindex { |x| x.match(/^HTTP\//) }
last_response = headers.slice(start_index, headers.size)
```

Or happy to review an even better proposal 😄

ajroetker:
@iMacTia Thanks for the review, I pushed up your suggested changes!

iMacTia:
Thanks @ajroetker.
This will probably go in v0.12.2

lib/faraday/utils.rb
```diff
@@ -95,7 +95,14 @@ def to_hash() ::Hash.new.update(self) end
 
       def parse(header_string)
         return unless header_string && !header_string.empty?
-        header_string.split(/\r\n/).
+
+        headers = header_string.split(/\r\n/)
+
+        # Find the last set of response headers.
+        start_index = headers.rindex { |x| x.match(/^HTTP\//) }
```
isaacseymour: I'm getting an issue where this returns `nil` when replaying a cassette from WebMock, because there's no `HTTP ...` line in the replayed headers. Is this a bug in Faraday or in WebMock?

iMacTia: That's really strange.
I haven't used Webmock cassettes but server responses should always start with the HTTP version. I suppose Webmock authors didn't think that was important and removed it from the cassette. I would suggest to open a ticket with them and double-check

isaacseymour: Just for reference, it was a bug in VCR. Someone has already fixed it in a PR though 😄 https://github.com/vcr/vcr/pull/657

bartoszkopinski: @iMacTia Hey, I ran into this as well. Agree this is a VCR bug but since the latest release (3.0.3) was last year I don't expect a quick fix. 
Would it be possible to add a `.to_i` cast here to keep it compatible?
For now the easiest solution would be to downgrade faraday to "0.12.1"

iMacTia: @bartoszkopinski can you please explain better where you would add the `.to_i`?
But please keep in mind that I'm not willing to add ugly patches to Faraday only to make it compatible with an outdated gem.
The solution in that case would be to either fix the gem (you can always fork it if it's dead), or monkey-patch tin your project

bartoszkopinski: @iMacTia I think `headers.rindex { |x| x.match(/^HTTP\//) }.to_i` should fix the problem (but I didn't test it).
If I understand correctly, this would simply cause it to start from the top, if `HTTP` header is missing. It's not VCR specific as I believe faraday shouldn't fail in this place anyway (the change is basically backwards incompatible by assuming the server response will always contain a certain string).
Looking at VCR code I now believe this was a conscious decision so it's even less likely to be fixed.

iMacTia: I see, I understand what you mean now.
I'll have a look at `rindex` implementation, I think what we're trying to add here is a fallback value in case there's no "HTTP" header. The maybe `.to_i` is not the best solution for that, but I get your point now!

iMacTia: According to this: https://ruby-doc.org/core-2.3.0/Array.html#method-i-rindex
I believe the best (or clearest, at least) implementation would then be:
```ruby
start_index = headers.rindex { |x| x.match(/^HTTP\//) } || 0
```

iMacTia: Fixed in #719

PR #686, opened 2017-04-17
expose wrapped_exception in all client errors

craig-day:
I have come across the desire to know specifically why my error timed out or why my connection failed, and providing access to the `wrapped_exception` is useful to have more granular control over my exception handling. I would like to be able do do something like this (simplified example).

```ruby
begin
  # do some stuff
rescue Faraday::ConnectionFailed => e
  raise unless e.wrapped_exception.is_a?(Net::OpenTimeout)
  retries += 1
  retry
end
```

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -3,7 +3,7 @@ module Faraday
   class MissingDependency < Error; end
 
   class ClientError < Error
-    attr_reader :response
+    attr_reader :response, :wrapped_exception
 
     def initialize(ex, response = nil)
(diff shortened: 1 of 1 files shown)


craig-day:
Turns out I can accomplish this with `e.cause` so take it or leave it I suppose.

iMacTia:
Hi @craig-day and thanks for your PR.
I wasn't even aware of the `cause` method, and from what I can see it's only available from Ruby 2.1.
For this reason, I think your PR would still be very useful and theoretically safe.
Happy to merge this is for the next release

PR #692, opened 2017-05-12
Fix compatibility with Net::HTTP::Persistent v3.

alexisbernard:
Since the version 3.0.0 of Net::HTTP::Persistent.new arguments have changed in a non compatible way. It expects named arguments. The change happened in the following commit: https://github.com/drbrain/net-http-persistent/commit/5d4b76c22dd38d29b6fbc1ed700a2a1c78c5abc4#diff-9da42550715aa09be7174a0babadcad3R512

This PR attempts to be compatible with both versions 2 and 3.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :test do
   gem 'httpclient', '>= 2.2'
   gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
   gem 'minitest', '>= 5.0.5'
-  gem 'net-http-persistent', '~> 3.0.0'
+  gem 'net-http-persistent', '~> 2.9.4'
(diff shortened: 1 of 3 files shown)


alexisbernard:
The build is failing because of code coverage to low for Ruby 2.4.0 only. It sounds logic that percentage is lower when `SSL=no` but I don't understand why it's even lower for 2.4.0 ...

iMacTia:
Hi @alexisbernard,

a similar approach has already been taken with #619.
I apologise if that branch have not been merged yet, but tests are passing with no issue in that case, so if I have to choose which PR to merge I would go for that one.
Currently we're waiting for an owner to review it as there are some important changes to the Gemfile.

Issue #693, opened 2017-05-17
Deletes with bodies


Carlbc18:
HTTP spec has changed in 2014 and allows for deletes to have bodies.  Is there a plan in place to resolve your delegator method in connection.rb to allow for a body with the delete method and not nil'ing it out?

Carlbc18:
A PR will be up shortly for your review.

iMacTia:
Hi @Carlbc18,

could you please provide a link to the above changes?
I guess that's only true for HTTP 2.0 specs, so we might need to consider this feature carefully.

Carlbc18:
https://tools.ietf.org/html/rfc7231#section-4.3.5

Carlbc18:
i have a code change with test that works (not sure if you want it). how do i create a pr for this project?

Carlbc18:
> I guess that's only true for HTTP 2.0 specs, so we might need to consider this feature carefully.

this is actually part of HTTP 1.1

iMacTia:
I would definitely consider it if you want to create a PR.
You can do it as with any other repo, simply fork `faraday` and apply the changes on your fork, then create a PR to our master branch.

Carlbc18:
Thanks @iMacTia. I'll put it up.  Whats your views on changing the version finally have a major version number with this?  This change could be breaking for many.  I'd like to follow semver pattern, and version faraday to be 1.0.0 at this point.  Let me know your thoughts.

iMacTia:
We're definitely trying to stick to semver and we've v1.0 scheduled with some breaking changes.
If this change is gonna break backward compatibility, then it will be released only in version 1.0 (which currently doesn't have a fixed release date).
However, my understanding is that you should now be able to OPTIONALLY add a body to GET, HEAD and DELETE requests, so shouldn't that be backward compatible? Am I missing something?

Carlbc18:
@iMacTia i versioned to 0.13.0 and the PR is up.  We can take any other conversation around this to the PR as i'm sure there will be comments.  Thanks again!

iMacTia:
There's an issue though, this might only work with ruby >= 2 as ruby 1.9 doesn't support keyword params 😞

Carlbc18:
[...]

PR #695, opened 2017-05-21
Bcarlson/http1 1 change to include body on deletes

Carlbc18:
These changes are to support HTTP 1.1 RFC https://tools.ietf.org/html/rfc7231 where HEAD, GET, DELETE all support a request body instead of merging request params into the url. For example on the delete (https://tools.ietf.org/html/rfc7231#section-4.3.5) the follow outlines the support for bodies: 

>  A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

The same verbiage is used for HEAD and GET.  Please let me know if there is anything else, or if i completely messed things up (totally possible).

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -14,7 +14,7 @@ require 'forwardable'
 #   conn.get '/'
 #
 module Faraday
-  VERSION = "0.12.1"
+  VERSION = "0.13.0"
 
   class << self
     # Public: Gets or sets the root path that Faraday is being loaded from.
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -116,44 +116,6 @@ module Faraday
 
     def_delegators :builder, :build, :use, :request, :response, :adapter, :app
 
-    # Public: Makes an HTTP request without a body.
-    #
-    # url     - The optional String base URL to use as a prefix for all
-    #           requests.  Can also be the options Hash.
-    # params  - Hash of URI query unencoded key/value pairs.
-    # headers - Hash of unencoded HTTP header key/value pairs.
-    #
-    # Examples
-    #
-    #   conn.get '/items', {:page => 1}, :accept => 'application/json'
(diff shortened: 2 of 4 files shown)


iMacTia:
Hi @Carlbc18 and thanks for this contribution.
Now I understand what you meant by "breaking changes" 😄!
The biggest issue I can see with this approach is that it removes the possibility to easily set url parameters on GET, HEAD and DELETE requests, even though the HTTP standard now allows you to have both.

Ideally, we would like Faraday to allow setting both as easily as providing an hash to the request.
I understand this is not easy since we have a generic `params` parameter on this methods, but it would be great to have something like this:

Option 1: When you provide parameters to Faraday `#get`, `#head` or `#delete`, they go on the URL by default (backward compatibility) but you can optionally set them in the body. This can be achieved with a connection configuration as it's just a switch (either one or the other).

Option 2: Faraday provides the possibility to easily setup both, keeping backward compatibility.

As you can see backward compatibility is quite important as we're talking about 3 of the most important functions in Faraday 😅, so I would really love to have it even for v1.0.
That said, Option 2 would definitely be my favourite, but even option 1 would be OK to begin with.
And of course both options can be released before v1.0, assuming we have backward compatibility 😃 

Open to suggestions on how to implement either of them

Carlbc18:
Hopefully it wasnt too scary of a change! I'm not very well versed in all of Faraday. I too like option 2 as a config option at object initialization. I might need some assistance in tracking down where everything is if you'd like me to make this change.

iMacTia:
Not scary at all, actually I like the idea of embracing the new standard.
I simply want to maintain backward compatibility in doing so and possibly leave it as an option for developers.
I'd suggest we go back to the issue and discuss in there how we want this to work before continuing with the PR

iMacTia:
Closing after discussion in https://github.com/lostisland/faraday/issues/693#issuecomment-305429333

PR #705, opened 2017-07-11
Fixes YAML serialization/deserialization for `Faraday::Utils::Headers`

### LINKED ISSUE #690: Issue with
Issue for bug reported in https://github.com/lostisland/faraday/pull/634#issuecomment-300050793 by @seanbjornsson.

iMacTia:
Fixes #690

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -10,10 +10,20 @@ module Faraday
         new(value)
       end
 
+      def self.allocate
+        new_self = super
+        new_self.initialize_names
+        new_self
+      end
+
       def initialize(hash = nil)
         super()
         @names = {}
         self.update(hash || {})
+      end
+
+      def initialize_names
+        @names = {}
       end
 
       # on dup/clone, we need to duplicate @names hash
@@ -113,14 +123,6 @@ module Faraday
               self[key] = value
             end
           }
-      end
-
-      def init_with(coder)
-        @names = coder['names']
-      end
-
-      def encode_with(coder)
-        coder['names'] = @names
       end
 
       protected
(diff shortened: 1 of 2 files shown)


iMacTia:
@olleolleolle @seanbjornsson @AvaelKross
I'm really sorry for the delay on this, but I finally got some time to spend investigating the issue.

**Issue with base functionality (before @AvaelKross / @olleolleolle fix):**
YAML.load calls `#allocate` on the object class (in this case `Faraday::Utils::Headers`) which skips the call to `#initialize`. In our case, this was causing `@names` to be uninitialised ("undefined method '[]' for nil:NilClass")

**Issue with encode_with/decode_with (after @AvaelKross / @olleolleolle fix):**
`encode_with` and `decode_with` are not extending the base behaviour of `#to_yaml`, but rather overriding it. This is causing all the key/value pairs to get lost in the process

**Final solution**
Unfortunately I had to revert @AvaelKross / @olleolleolle changes. Instead, I've overridden `#allocate` in `Faraday::Utils::Headers` initialising the `@names` instance variable.

Your input on the above will be much appreciated before I merge this in 😄
Moreover, tests have been upgraded with suggestions from @seanbjornsson, but if you have any personal project where you can test this special branch it would be even better.

olleolleolle:
@iMacTia I learned about `#allocate` _just yesterday_ and now, _now_, you use it in a crucial place. THANK YOU, and wow serendipity!

iMacTia:
Ahahah call them "coincidences" if you want 😄
Thanks for the feedback and I'm glad you liked the solution 😃

seanbjornsson:
Fantastic @iMacTia, thank you for coming back around to this! Sorry I couldn't respond sooner. 

[The spec additions](https://github.com/lostisland/faraday/pull/705/files#diff-836865c9709b47869e1960b80739db0aR94), however simple, capture my use case perfectly. 
I'll see about updating my project, and ripping out my monkey patch fix. 
Thanks again.

PR #712, opened 2017-07-14
Environment proxy refactoring

iMacTia:


iMacTia:
@JasonBarnabe I'm happy with what I got and there are now more tests.
Proxy support have been added to the `Faraday.get` feature and is also dynamically re-retrieved if you call an absolute url on a pre-initialised connector.
However, I had to do some changes to the code and delete one test in order for this to work.
It was a small and (IMO) insignificant test, but still I would prefer to avoid any issue on an hotfix release.
I'll move this change into the 0.13 relase. If you have some time to test this out in the meantime, that would be appreciated.

iMacTia:
0.12.2 has been released so I'm now merging this which will be packed in 0.13.0

PR #716, opened 2017-07-28
fix net_http_persistent adapter to be ready for >= v3.0.0

Overbryd:
Hi,

using `net_http_persistent` with version **3.0.0** crashes on initialisation, because the method signature of `Net::HTTP::Persistent::new` changed to named arguments. The backtrace is here:

```
ArgumentError:
       wrong number of arguments (given 1, expected 0)
net-http-persistent-3.0.0/lib/net/http/persistent.rb:505:in `initialize'
faraday-0.12.2/lib/faraday/adapter/net_http_persistent.rb:22:in `new'
faraday-0.12.2/lib/faraday/adapter/net_http_persistent.rb:22:in `net_http_connection'
faraday-0.12.2/lib/faraday/adapter/net_http.rb:85:in `with_net_http_connection'
faraday-0.12.2/lib/faraday/adapter/net_http.rb:33:in `call'
```

This commit should fix the issue for net_http_persistent >= 3.0.0.
Although, I followed the guidelines mentioned in CONTRIBUTING, I could not get the test suite to run. ~So please someone, run the tests for this adapter, to incorporate the fix.~ *(thanks Travis)

Kind regards,

Lukas Rieder

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :test do
   gem 'httpclient', '>= 2.2'
   gem 'mime-types', '~> 1.25', :platforms => [:jruby, :ruby_18]
   gem 'minitest', '>= 5.0.5'
-  gem 'net-http-persistent', '~> 2.9.4'
+  gem 'net-http-persistent', '~> 3.0.0'
   gem 'patron', '>= 0.4.2', :platforms => :ruby
   gem 'rack-test', '>= 0.6', :require => 'rack/test'
   gem 'rest-client', '~> 1.6.0', :platforms => [:jruby, :ruby_18]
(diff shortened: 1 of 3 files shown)


iMacTia:
Hi @Overbryd and thanks for your PR.
Unfortunately the issue with your approach is that if you ONLY support net_http_persistent >= 3, then you have to drop ruby < 2.1, or you'll get the following like Travis:

```ruby
Gem::InstallError: net-http-persistent requires Ruby version ~> 2.1.
```

We currently support ruby >= 1.9.3 so unfortunately there's no way we can drop net_http_persistent v2 support at the moment.
Good news is that someone else is already working on a PR that will add support for net_http_persistent v3 while keeping v2 support: #619 
It was a complicated PR but It's now in the final review phase so please subscribe to it to keep up to date and you should be able to use net_http_persistent v3 when Faraday 0.13 will ship :)

PR #731, opened 2017-09-11
do not set Excon's write_timeout to our open_timeout

apachelogger:
our timeout corresponds to all timeouts on the excon side.
our open_timeout exclusively corresponds to excon's connect_timeout (i.e.
timeout for opening a new connection).

As such we must only set the excon write_timeout from our timeout.
Previously we'd also set write_timeout from our open_timeout, overriding
the original timeout. Since open_timeout can be fairly short while
write_timeout may be fairly long (e.g. blocking upload of a large file)
the previous code made open_timeout effectively unusable as it would break
PUTing files.

### DIFF
--- a/lib/faraday/adapter/excon.rb
+++ b/lib/faraday/adapter/excon.rb
@@ -30,7 +30,6 @@ module Faraday
 
           if req[:open_timeout]
             opts[:connect_timeout]   = req[:open_timeout]
-            opts[:write_timeout]     = req[:open_timeout]
           end
 
           if req[:proxy]


iMacTia:
Hi @apachelogger, I'm having a look at those tests failing as it seems unrelated to your changes.
The change itself makes sense, as by the time you write the request the connection is already open.
I'll try to have a look at Travis

iMacTia:
Hi @apachelogger, apologies for the delay but the issue on Travis was tricky to solve and took me some time! Please pull changes from master into your branch and we should see green tests 👍

apachelogger:
It's a green! :tada: :wink:

iMacTia:
Great, thanks @apachelogger and sorry again for the delay!

PR #735, opened 2017-10-02
Fixed client error status code range.

borsothy:


borsothy:
I understand the idea behind raising the same exception class for 4xx _and_ 5xx responses, but the naming is really confusing. 5xx is server error, and it makes it really hard to filter for client and server errors separately.

This is more of a proposal than a real solution, tests are failing, and obviously more work needs to be done - please let me know what you think, and how this should be addressed.

iMacTia:
Hi @borsothy, I totally agree with you, I think here we should have a `Faraday::Error::ClientError` (4xx), but also a `Faraday::Error::ServerError` (5xx).
The change itself looks quite easy, but I'm really scared about backward compatibility.
Unless we can think of a backward compatible fix, I can only push this into v1.0

iMacTia:
Hi @borsothy,
I've added this to the v1.0 list and changed the base branch to v1.0.
The fix itself is ok, but as I said we need to introduce a new `Faraday::Error::ServerError` class that covers errors 5xx and tests will obviously need some work as well.

Would you still like to work on this?

bronislav:
As an option, we may make this configurable and allow users to explicitly agree with correct statuses and let others use old ones.

@borsothy, @iMacTia what do you think?

iMacTia:
I'd rather take advantage of v1.0 (backwards incompatible) to "force" everyone following the correct convention. The configuration would have been a good solution if we wanted to introduce this in 0.x versions as well

iMacTia:
Closing in favour of #841

PR #739, opened 2017-10-30
Handle nil headers in 0.9.x (F3ndot #728)

iMacTia:


iMacTia:
@f3ndot I'm sorry but after fixing versions for half of the Gemfile I'm still not able to make this working on all Ruby versions. Getting the correct gems to work on older rubies has become a mess after all this time.

> Heya @iMacTia, so I was able to sidestep the issue personally with some tweaking in my application. We're still using Faraday 0.9 for the application.

At this point I would give up and ask you to use your fix to keep working with Faraday 0.9.
If you wish to continue from here and keep fixing it, feel free to pull from this branch into yours otherwise please let me know and I'll delete it.

f3ndot:
I think it's fine to close this one out since I have my workaround in place. Should others come in with same issue in the future, I can look into trying again then.

Thanks for the effort!

Issue #358, opened 2014-03-26
Faraday::Adapter::Test can't stub by URL, only by path


Peeja:
I'm trying to write a spec for an object which will fetch content from a URL. (I'm abstracting away lots of domain-specific stuff that's not relevant to this issue here.) I want to stub the Faraday request that fetches the content. I want the spec to be something like so:

``` ruby
let(:http_connection) {
  Faraday.new do |builder|
    builder.adapter :test, http_stubs
  end
}

let(:http_stubs) { Faraday::Adapter::Test::Stubs.new }
let(:fetcher) { ThingFetcher.new(http_connection: http_connection) }

it "gets the right thing" do
  http_stubs.get("http://example.test/things/1") { [200, {}, 'a thing'] }

  expect {
    fetcher.fetch("http://example.test/things/1")
  }.to change { Thing.count }.by(1)

  expect(Thing.last.contents).to be('a thing')
end
```

But it turns out that doesn't work because

```
Faraday::Adapter::Test::Stubs::NotFound: no stubbed request for get /things/1
```

[`Faraday::Adapter::Test` normalizes the URL, stripping off the host information](https://github.com/lostisland/faraday/blob/3dc615a72eae5cec55d2c73dfdd8efea00f2c283/lib/faraday/adapter/test.rb#L146) before it looks it up in the stubs. But I _want_ to assert that it uses the right host. I want to stub the entire URL.

How can I do that? And if I can't, is there a path toward making it so one can?

erik-escobedo:
What the `any_volunteer?` tag means? I do wanna work in this and have an idea about how to fix it. Should I start working on the PR? 😮

iMacTia:
Hi @erik-escobedo, yes the `any_volunteer?` tag was introduced before the new official `help wanted` and `good first issue`. I suppose it would be better to update them. Anyway, I'm happy if you want to help on this one. I would suggest you expose your idea before working and opening the PR so we can discuss about it beforehand (and hopefully save some of your time)

erik-escobedo:
Well, as @Peeja pointed out, when the URL is provided the `host` is striped off by `Faraday::Utils.normalize_path(env[:url])` therefore, the `matched?` method completely ignores it. My plan is to store the host in a new variable and make sure the `host` provided on the stub matches the one provided on the `request`. If no host is provided, the variable stores `nil` and nothing changes. Even tests won't need to be updated. Then, only 2 new tests need to be written:

1. For when host is provided and matches
[...]

Issue #718, opened 2017-08-09
Distinguish TimeoutErrors for open and read timeouts


coberlin:
In faraday/adapter/rack.rb, TimeoutError is raised for either both open and read timeouts:

```
timeout  = env[:request][:timeout] || env[:request][:open_timeout]
response = if timeout
  Timer.timeout(timeout, Faraday::Error::TimeoutError) { execute_request(env, rack_env) }
else ... end
```
According to https://stackoverflow.com/questions/10322283/what-is-timeout-and-open-timeout-in-faraday, open_timeout is for the tcp connection and timeout is for the response read.

It would be nice to have separate exception types for these timeouts. Then we could determine whether or not to retry the request.  Does adding something like Faraday::Error::OpenTimeoutError and Faraday::Error::ResponseTimeoutError and using those here make sense?

iMacTia:
Hi @coberlin I believe this might be a nice addition, I'm just scared about backwards compatibility.
However, a possible solution for this might be to have `OpenTimeoutError` and `ResponseTimeoutError` to inherit from `TimeoutError`, so that existing `rescue`s will keep working as expected.

coberlin:
The rack_adapter might be the wrong place for this feature. I think Rack applications don't necessarily distinguish between open and read timeouts. Perhaps this feature would work in the HTTPClient adapter or other adapters? From adapter/httpclient.rb:
```  
    @app.call env
  rescue ::HTTPClient::TimeoutError, Errno::ETIMEDOUT
    raise Faraday::Error::TimeoutError, $!
  rescue ::HTTPClient::BadResponseError => err
    if err.message.include?('status 407')
      raise Faraday::Error::ConnectionFailed, %{407 "Proxy Authentication Required "}
    else
      raise Faraday::Error::ClientError, $!
    end
  rescue Errno::ECONNREFUSED, IOError, SocketError
    raise Faraday::Error::ConnectionFailed, $!
  rescue => err
    if defined?(OpenSSL) && OpenSSL::SSL::SSLError === err
      raise Faraday::SSLError, err
    else
      raise
    end
```
::HTTPClient::TimeoutError has 3 subclasses ConnectTimeoutError, ReceiveTimeoutError, SendTimeoutError, see e.g http://www.rubydoc.info/gems/httpclient/2.1.5.2/HTTPClient/TimeoutError

[...]

Issue #620, opened 2016-10-18
List of features for the next big releases


iMacTia:
## Faraday 1.0
- [x] Set minimum Ruby version to >= 2.2
- [x] Streaming (Net::HTTP) #604 
- [x] Change the way adapters are managed #47 #121 
- [ ] API improvements #241 #305 #462 #517 #693 #718 #735 
- [ ] Pull out adapters as separate gems. Faraday-Typheous, Faraday-Patron, etc. #112 
- [ ] Supports IPv6 #621 
- [ ] Streaming (others)
- [ ] Improve documentation/Readme #425 #575 
- [ ] Improve Codeclimate maintainability (https://codeclimate.com/github/lostisland/faraday)
- [ ] Pass the entire project through Rubocop and set it up on Github as an integration
- [ ] Improve Wiki adding more examples an how-to
- [ ] Convert tests to RSpec?
- [x] Setup Github integration for test coverage and code metrics
- [ ] Remove all deprecated methods/behaviours (with related warnings)

grosser:
I'm looking for streaming support ... is anyone working on 1.0 ? ... if not why was the other PR closed 😕

iMacTia:
Hi @grosser, reasons are expressed in this https://github.com/lostisland/faraday/pull/604#issuecomment-259125910.
Main reason was that the PR was only compatible with `Net::HTTP` adapter and the fact that streaming have been marked for v1.0.
There's no roadmap for v1.0 at the moment, so no one actively working on streaming for the time being.

jcoyne:
@iMacTia I'm a bit disappointed that you say "no one actively working on streaming" because it seems you squashed work on #604 (also #522, #461) by saying "Hold tight, this will be in the next release", but then not working on it.  Why not accept the communities changes considering there doesn't seem to be momentum by the core team?

iMacTia:
@jcoyne I didn't mean that streaming is not in yet BECAUSE "no one actively working on streaming". I'm fully aware that the fact no one is working on it is mainly my fault. However, I've explained why I pushed back on #604 and the problem was not the implementation.
In order to get 604 merged in one of the following needs to happen first:

1. Support for all the other adapters is added (right now only Net::HTTP is supported)
2. We wait for version 1.0 because we might remove direct support for other adapters and gamily them into external projects. That would make #604 mergeable as it is now, but unfortunately it has not been agreed internally yet.

I apologise for being slow and not having enough time to invest in working on any of the above solutions.
[...]

Issue #305, opened 2013-09-23
No method for OPTIONS HTTP verb


gjtorikian:
Shouldn't it be included in this array?

https://github.com/lostisland/faraday/blob/master/lib/faraday/connection.rb#L137

nhocki:
I am terribly sorry to _bump_ this, but any agreement on supporting `options` as a method for `OPTIONS` requests? I'd be more than happy to submit a pull request to make this happen.

mislav:
I'm still not sold on the idea. What would the current `options` method be called then?

Other methods that we support are verbs: get, post, put, delete. It makes it easy to understand that some action is happening when you call them. `options` wouldn't be a verb and as such I don't think it would make a nice shorthand method. If people are using `options` a ton in day-to-day code and can't bear to use `run_request` for some reason (I'd like to hear that reason!) then I would suggest naming the new method `http_options` to indicate that this is a HTTP request method.

nhocki:
The argument of "the other methods are verbs" should not matter. They are not methods in Faraday because they are verbs, those method exist because they are HTTP methods. The same should happen for `options` IMO.




If this is the reason for not having `options`, then `head` should not be defined, as "head" in HTTP is not about "going", the verb.




I totally get the concern about breaking the API with `options`, but it's a big surprise that it is not "supported". It's way more consistent to have all HTTP methods handled the same.




And about not using `run_request`, like it was pointed before, the code without it is much much cleaner.




It's still your call (obviously :-) ) to decide to support this or not, but I'd love to change your mind about it, and if it's not going to be supported, that it's for something else other than "_options_ is not a verb". 



--
Nicolás Hock Isaza

bithavoc:
+1 @nhocki

dwo:
This caught me by surprise today. All of the examples with `conn.get`, `conn.post`, `conn.delete` etc. lead me to believe the obvious way to do an OPTIONS request was `conn.options`. 

Perhaps we could document this inconsistency and its `run_request` workaround in the `Faraday::Connection` rdocs or the README? I'm happy to compose a pull request.

guyisra:
+1

Why isn't options the same as GET, POST, DELETE?

iMacTia:
Even though I personally agree with @mislav, I can't ignore @sferik opinion and the community feedback. I checked the code and noticed that `options` is simply an `attr_reader`, plus it's being used mainly by tests.
[...]

PR #741, opened 2017-11-02
Faraday::Adapter::Test stubs now support entire urls (with host)

erik-escobedo:
Addresses issue #358

### DIFF
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -53,17 +53,17 @@ module Faraday
           @stack.empty?
         end
 
-        def match(request_method, path, headers, body)
+        def match(request_method, host, path, headers, body)
           return false if !@stack.key?(request_method)
           stack = @stack[request_method]
           consumed = (@consumed[request_method] ||= [])
 
-          stub, meta = matches?(stack, path, headers, body)
+          stub, meta = matches?(stack, host, path, headers, body)
           if stub
             consumed << stack.delete(stub)
             return stub, meta
           end
-          matches?(consumed, path, headers, body)
+          matches?(consumed, host, path, headers, body)
         end
 
         def get(path, headers = {}, &block)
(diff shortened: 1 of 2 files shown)


lib/faraday/adapter/test.rb
```diff
@@ -110,37 +110,44 @@ def verify_stubbed_calls
...
+        def matches?(request_host, request_uri, request_headers, request_body)
           request_path, request_query = request_uri.split('?')
           request_params = request_query ?
             Faraday::Utils.parse_nested_query(request_query) :
             {}
           # meta is a hash use as carrier
           # that will be yielded to consumer block
           meta = {}
-          return path_match?(request_path, meta) &&
+          return host == request_host &&
```
iMacTia: Unless I'm missing something, this line should break backwards compatibility.
That's because we're now passing the `request_host` to be checked, but `host` could be nil (in case the stub is created with just a path).
That should cause this condition to fail

test/adapters/test_middleware_test.rb
```diff
@@ -81,7 +94,7 @@ def test_middleware_allow_different_outcomes_for_the_same_request
     end
 
     def test_yields_env_to_stubs
-      @stubs.get '/hello' do |env|
+      @stubs.get 'http://foo.com/hello' do |env|
```
iMacTia: See comment above regarding the host.
We still want to allow people to stub paths (hence URLs with no host).
We should then keep this tests unchanged and duplicate all of them to test that the stub works WITH and WITHOUT the host specified

test/authentication_middleware_test.rb
```diff
@@ -2,7 +2,7 @@
 
 class AuthenticationMiddlewareTest < Faraday::TestCase
   def conn
-    Faraday::Connection.new('http://example.net/') do |builder|
+    Faraday::Connection.new do |builder|
```
iMacTia: Why did you have to remove this?

erik-escobedo: Specifying a domain in the connection forces the stubs to only be available for that domain. Therefore, making `connection.get('/path')` (without the host) won't match. I removed this to make the connection domainless.

test/adapters/test_middleware_test.rb
erik-escobedo: Yeah, I'm actually kind of confused about this. The question is, what does it mean when user doesn't specify any host in the stub? Does it mean the stub should be available for all hosts? I mean, should `connection.get('http://domain.com/path')` look for stubs declared like `stubs.get('/path')` as well as `stubs.get('http://domain.com/path')`.

I think the use case mentioned in the issue (https://github.com/lostisland/faraday/issues/358#issue-30252173) is trying to make sure Faraday is hitting the right domain. With my changes is still possible to make stubs with no host, but they'll only match requests with no host as well. Is this wrong?

iMacTia: Apologies for not being more clear from the beginning.
The wanted behaviour is the following:

1. If a hostname is specified on the stub, then both the hostname AND the path should match.
2. If no hostname is specified on the stub (hence only the path), then any hostname should match.

Point 1 will "fix" #358, while point 2 will maintain backward compatibility, as the current behaviour is "stub this path on any host" and we should keep that working as well 👍

test/authentication_middleware_test.rb
iMacTia: See my new comment above on the expected behaviour.
I hope this is clear now, this test should definitely work with the domain in place (it will test point 2 of my comment).

Rather than changing existing tests, your approach should be to duplicate them in order to test both points (old and new behaviour).

iMacTia:
Hey @erik-escobedo, sorry I'm just touching base to know if there's any progress on this.
I have included it into v0.14.0 which will I'd like to ship before the end of the month.
Do you think you can find some time to work on this before then?

erik-escobedo:
Totally. I'm working on it right now

iMacTia:
Amazing job @erik-escobedo 👍 
Nice extension of the existing test adapter stubs.
Looks ready for merging to me, were you planning any additional work on it?

erik-escobedo:
Thanks! No, I'm not planning to make additional work on this, until you think it's necessary. I've been reading about this issue and I'm ready to help if you give the green light:
https://github.com/lostisland/faraday/issues/718

erik-escobedo:
Thanks! No, I'm not planning to make additional work on this, until you think it's necessary. I've been reading about this issue and I'm ready to help if you give the green light:
https://github.com/lostisland/faraday/issues/718

iMacTia:
Merged 👍!
Regarding #718, that's a bit tricky as I'm still deciding what to do.
I'll add a comment there with the latest details so you can leave your thoughts.
You can also contribute to v1.0 having a look at the list on #620 if you want 😄

erik-escobedo:
Of course! I'd like to work on this one: https://github.com/lostisland/faraday/issues/305

iMacTia:
Sounds great! There's still some discussion around what the current "options" method should be called so please join the discussion and provide a suggestion before working on the PR!
If you're out of ideas then let me know and I'll have a look and tell you my opinion 😉

PR #743, opened 2017-11-05
Warn about adapter being unset when making a request

antstorm:
Hi,

I was really surprised that you have to specify an adapter when using an initialiser with block. I would expect additional configuration abilities when passing in a block, but not a completely different behaviour. Especially when a `default_adapter` is publicly configurable, suggesting that it gets picked up by default.

I've added this simple warning to avoid confusion, because I was not sure how you guys feel about setting the defaults when initialising `Faraday::Connection` anyways and then exposing them for a reconfiguration within a passed block. If you're happy with it I'll open another PR.

As for this one — I'm less familiar with `MiniTest`. Can you suggest a good approach of testing this change?

### DIFF
--- a/lib/faraday/rack_builder.rb
+++ b/lib/faraday/rack_builder.rb
@@ -138,6 +138,8 @@ module Faraday
     #
     # Returns a Faraday::Response.
     def build_response(connection, request)
+      warn 'WARNING: No adapter was configured for this request' unless adapter_set?
+
       app.call(build_env(connection, request))
     end
 


iMacTia:
Hi @antstorm and thanks for contributing.

I agree this warning might help some devs to notice in case they're doing something wrong, so happy to merge this in.

Regarding the broader change you suggested above, I can reassure you we're already working on it in v1.0. You can see the discussion so far about it in #121.

antstorm:
@iMacTia thanks! I'm really glad it's being taken care of.

PR #744, opened 2017-11-12
Feature/v1.0/ruby versions

iMacTia:
## Description
JRuby and Rubinius tests have been red for too long.
There's no one actively supporting these languages and it doesn't make sense to keep supporting them this way. Moreover, Travis builds takes a lot of time to run the additional tests (for nothing).
This PR removes all jruby and rbx from Travis except for jruby-head (should be latest JRuby 9000). I worked a bit with JRuby so I'll try to make these tests green again but if that doesn't work then I'll drop it as well.

## Todos
- [x] ~Attempt to get green tests on JRuby Head (9xxx serie)~
- [x] Update README

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,36 +6,13 @@ script: bundle exec script/test
 cache: bundler
 
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1
   - 2.2
-  - 2.3.0
-  - 2.4.0
-  - jruby-19mode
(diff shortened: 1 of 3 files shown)


iMacTia:
Removing JRuby from the travis matrix "for now".
I'll attempt to reintroduce it again later on after some refactoring and see if it gets any better.
At the moment, the most important thing is to get tests running quickly

PR #750, opened 2017-11-19
Changes the way adapters are managed

### LINKED ISSUE #47: Adapters should be Rack apps not middleware
This is our current Rack app (i.e. the "inner app" wrapped by middleware):

```
app = lambda { |env|
  response = Response.new
  response.finish(env) unless env[:parallel_manager]
  env[:response] = response
}
```

It's really kind of a hack. I think our adapters should each perform this and stop acting like middleware when they really play a role of an endpoint. I can't think of a use case where you would want to mount middleware _after_ the adapter.

iMacTia:
## Description
Changes the way adapters are managed to gain the following:
* Adapter can be set anywhere in the stack and will always be the last
* We can easily check if the adapter is set or not
* Use default adapter if no adapter set

Fixes #47 and #121

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -233,20 +233,17 @@ module Faraday
       set_authorization_header(:authorization, type, token)
     end
 
-    # Internal: Traverse the middleware stack in search of a
-    # parallel-capable adapter.
+    # Internal: Check if the adapter is parallel-capable.
     #
-    # Yields in case of not found.
+    # Yields otherwise or if no adapter is set yet.
     #
-    # Returns a parallel manager or nil if not found.
+    # Returns a parallel manager or nil if yielded.
     def default_parallel_manager
       @default_parallel_manager ||= begin
-        handler = @builder.handlers.detect do |h|
-          h.klass.respond_to?(:supports_parallel?) and h.klass.supports_parallel?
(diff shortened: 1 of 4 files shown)


iMacTia:
@olleolleolle want to have a look 😄 ?

lib/faraday/connection.rb
```diff
@@ -233,20 +233,17 @@ def authorization(type, token)
...
     def default_parallel_manager
       @default_parallel_manager ||= begin
-        handler = @builder.handlers.detect do |h|
-          h.klass.respond_to?(:supports_parallel?) and h.klass.supports_parallel?
-        end
+        adapter = @builder.adapter.klass if @builder.adapter
 
-        if handler
-          handler.klass.setup_parallel_manager
+        if adapter and adapter.respond_to?(:supports_parallel?) and adapter.supports_parallel?
```
olleolleolle: You may wish to use `&&` over `and`?

iMacTia: Yeah that's the Rubocop convention and I've already planned to scan the entire gem through it before releasing 1.0. However in that file `and` is preferred to `&&` so I went for that.
It's just temporary, we'll eventually converge into the standard and fix all these occurrences later on

PR #758, opened 2018-01-20
[1.0] YARD docblocks annotation - a beginning in a single file

olleolleolle:
This PR adds a docblock to a single ~method~ file, to improve the API documentation in one place.

Just to see it work, and have it render at rubydoc.info later on. http://www.rubydoc.info/gems/faraday/Faraday

This early image is the rendered preview I had on my laptop:

The idea with this PR is to find a desired level of docblocks "documentation granularity" and exactness. What we're shooting for.

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -52,17 +52,18 @@ module Faraday
     # @option options [Hash] :proxy Hash of Proxy options.
     # @return [Faraday::Connection]
     #
-    # @example
-    #
+    # @example With an URL argument
     #   Faraday.new 'http://faraday.com'
-    #   # => http://faraday.com?page=1
+    #   # => Faraday::Connection to http://faraday.com
     #   
+    # @example With an URL argument and an options hash
     #   Faraday.new 'http://faraday.com', :params => {:page => 1}
-    #   # => http://faraday.com?page=1
+    #   # => Faraday::Connection to http://faraday.com?page=1
     #
+    # @example With everything in an options hash
     #   Faraday.new :url => 'http://faraday.com',
     #               :params => {:page => 1}
-    #   # => Faraday::Connection
+    #   # => Faraday::Connection to http://faraday.com?page=1
     def new(url = nil, options = nil)
       block = block_given? ? Proc.new : nil
       options = options ? default_connection_options.merge(options) : default_connection_options
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -3,13 +3,15 @@ require 'cgi'
 require 'set'
 require 'forwardable'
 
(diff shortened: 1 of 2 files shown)


lib/faraday.rb
```diff
@@ -37,33 +37,32 @@ class << self
...
-    #
-    #   # same
+    #   # => http://faraday.com?page=1
     #
     #   Faraday.new :url => 'http://faraday.com',
-    #     :params => {:page => 1}
-    #
-    # Returns a Faraday::Connection.
+    #               :params => {:page => 1}
+    #   # => Faraday::Connection
```
iMacTia: This just demonstrates how the old documentation was confusing.
I think results were put on top of the corresponding example, not below.
So the correct example -> results would be:
```ruby
Faraday.new 'http://faraday.com'
# => Faraday::Connection to http://faraday.com

Faraday.new 'http://faraday.com', :params => {:page => 1}
# => Faraday::Connection to http://faraday.com?page=1

Faraday.new :url => 'http://faraday.com',
            :params => {:page => 1}
# => Faraday::Connection to http://faraday.com?page=1
```

olleolleolle: Thanks for good review item!

olleolleolle: Using `@example` more than once, and using its title feature helps this method, I think:

image

iMacTia: That's looks even better 👍 !

lib/faraday.rb
```diff
@@ -6,7 +6,7 @@
 # Public: This is the main namespace for Faraday.  You can either use it to
 # create Faraday::Connection objects, or access it directly.
 #
-# Examples
+# @example
```
iMacTia: Would be great to have 2 separate examples here as well 😃

olleolleolle: Right, now we have a form, a style of explanation, and we can implement it throughout. 

I can let this PR be about docblocks in this file.

olleolleolle:
@iMacTia Now, I spent some more time adding docs to the whole file, and it makes me learn things about YARD, which is good.

I'm now using `@overload` to be able to get the docs about some attr_writer methods to render - this tip comes from YARD documentation.

I noted some default return values.

iMacTia:
Looks much better now 👍 good work!
We can use this as a demonstration on how we would like YARD to be introduced across the gem.
One more thing: I really would like v0.14.0 to be the last minor version before 1.0.
I've just aligned the 1.0 branch to master, please rebase your PR against that and switch the base branch on your PR

olleolleolle:
@iMacTia Yep, there. Thanks for keeping 1.0 aligned.

PR #767, opened 2018-02-03
Remove unused class error

edymerchk:
## Description

Since the JSON Request middleware was removed on https://github.com/lostisland/faraday/commit/db554b29260ff469e40e530af498b38433ede8ab this error class is not longer needed


iMacTia:
Thanks for taking the time to investigate and cleanup 👍

PR #768, opened 2018-02-06
[api docs] NestedParamsEncoder

olleolleolle:
This PR adds a bit of API documentation to an undocumented class.

Questions for reviewer:

- [x] Perhaps you know a narrative comment for the description of this class?

### DIFF
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -7,6 +7,11 @@ module Faraday
       def_delegators :'Faraday::Utils', :escape, :unescape
     end
 
+    # @param params [nil, Array, #to_hash] parameters to be encoded
+    #
+    # @return [String] the encoded params
+    #
+    # @raise [TypeError] if params can not be converted to a Hash
     def self.encode(params)
       return nil if params == nil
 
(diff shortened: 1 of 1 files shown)


lib/faraday/parameters.rb
```diff
@@ -63,6 +68,11 @@ def self.encode(params)
       return buffer.chop
     end
 
+    # @param query [nil, String]
+    #
+    # @raise [TypeError] if the nesting is incorrect
+    #
+    # @return [Array<Array, String>] the decoded params
```
iMacTia: I was looking for a convention on tags order, but I couldn't find anything.
At least I think we should try to be consistent. I prefer the order of the method above so can we align this one to it (`@param`, `@return`, `@raise`)?

olleolleolle: 👍

olleolleolle:
@iMacTia Thanks for the input and the class-level comment.

olleolleolle:
@iMacTia And I would like to thank the participants in the workshop at Malmo.rb where we did the investigation and markup work!

iMacTia:
Oh wow didn't know that was a team effort! Well done everyone 🥇 !

PR #770, opened 2018-02-13
Added retry block option to retry middleware.

borsothy:
## Description
Added `retry_blk` option to the retry middleware, so you can set a block which is executed after every retry.

## Todos
- [ ] Tests
- [ ] Documentation


borsothy:
Please let me know if you like the idea, and how it could be improved.

borsothy:
The usecase we want this for is sending a metric/report after every retry to an external service, so we can alert if there's a spike in retries.

Addressed your comments, thanks for the feedback :)

borsothy:
Absolutely, actually I just realised I only need `env`, but I can absolutely extend the block parameters.

iMacTia:
Cool, I noticed you've also updated the description to point out the 4 parameters.
Looks good now and all tests are green 👍 
LGTM

PR #773, opened 2018-02-19
Honor Retry-After header

bronislav:
## Description
Improve retry interval calculations to honor `Retry-After` header if present. This header may be set in case if a server does not respond or request being throttled. 

## Todos
- [x] Tests
- [ ] Documentation

## Additional Notes
I thought to create separate middleware for this purpose but then decided to just improve existing retry logic.

### DIFF
--- a/lib/faraday/request/retry.rb
+++ b/lib/faraday/request/retry.rb
@@ -23,7 +23,9 @@ module Faraday
     IDEMPOTENT_METHODS = [:delete, :get, :head, :options, :put]
 
     class Options < Faraday::Options.new(:max, :interval, :max_interval, :interval_randomness,
-                                         :backoff_factor, :exceptions, :methods, :retry_if, :retry_block)
+                                         :backoff_factor, :exceptions, :methods, :retry_if, :retry_block,
+                                         :retry_statuses)
+
       DEFAULT_CHECK = lambda { |env,exception| false }
 
       def self.from(value)
@@ -56,7 +58,8 @@ module Faraday
 
       def exceptions
         Array(self[:exceptions] ||= [Errno::ETIMEDOUT, 'Timeout::Error',
-                                     Error::TimeoutError])
+                                     Error::TimeoutError,
+                                     Faraday::Error::RetriableResponse])
       end
 
       def methods
@@ -71,6 +74,9 @@ module Faraday
         self[:retry_block] ||= Proc.new {}
       end
 
+      def retry_statuses
+        Array(self[:retry_statuses] ||= [])
+      end
     end
 
     # Public: Initialize middleware
@@ -106,12 +112,13 @@ module Faraday
       @errmatch = build_exception_matcher(@options.exceptions)
     end
 
-    def sleep_amount(retries)
-      retry_index = @options.max - retries
-      current_interval = @options.interval * (@options.backoff_factor ** retry_index)
-      current_interval = [current_interval, @options.max_interval].min
-      random_interval  = rand * @options.interval_randomness.to_f * @options.interval
-      current_interval + random_interval
+    def calculate_sleep_amount(retries, env)
+      retry_after     = calculate_retry_after(env)
+      retry_interval  = calculate_retry_interval(retries)
+
+      return if retry_after && retry_after > @options.max_interval
+
+      retry_after && retry_after >= retry_interval ? retry_after : retry_interval
     end
 
     def call(env)
@@ -119,16 +126,21 @@ module Faraday
       request_body = env[:body]
       begin
         env[:body] = request_body # after failure env[:body] is set to the response body
-        @app.call(env)
+        @app.call(env).tap do |resp|
+          raise Faraday::Error::RetriableResponse.new(nil, resp) if @options.retry_statuses.include?(resp.status)
+        end
       rescue @errmatch => exception
         if retries > 0 && retry_request?(env, exception)
           retries -= 1
           rewind_files(request_body)
           @options.retry_block.call(env, @options, retries, exception)
-          sleep sleep_amount(retries + 1)
-          retry
+          if (sleep_amount = calculate_sleep_amount(retries + 1, env))
+            sleep sleep_amount
+            retry
+          end
         end
-        raise
+
+        raise unless exception.is_a?(Faraday::Error::RetriableResponse)
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/faraday/request/retry.rb
```diff
@@ -106,26 +112,32 @@ def initialize(app, options = nil)
...
     end
 
     def call(env)
       retries = @options.max
       request_body = env[:body]
       begin
         env[:body] = request_body # after failure env[:body] is set to the response body
-        @app.call(env)
+        @app.call(env).tap do |resp|
+          raise Faraday::Error::RetriableResponse.new(nil, resp) if @options.retry_statuses.include?(resp.status)
```
iMacTia: I can see what you're trying to do with `Faraday::Error::RetriableResponse`, however I don't think it's a good idea to introduce this exception and use it this way.
The reason is that, after we reach the maximum number of retries, the new exception will be propagated outside and this might be an unexpected behaviour compared to what happens now.

I like the idea of allowing to define a list of statuses that you want to retry, but this should be tested internally without raising any exception.
My suggestion is to add this to `retry_request?`.

lib/faraday/request/retry.rb
```diff
@@ -106,26 +112,32 @@ def initialize(app, options = nil)
...
       retry_index = @options.max - retries
       current_interval = @options.interval * (@options.backoff_factor ** retry_index)
       current_interval = [current_interval, @options.max_interval].min
       random_interval  = rand * @options.interval_randomness.to_f * @options.interval
-      current_interval + random_interval
+
+      retry_after = calculate_retry_after(env)
+      interval    = current_interval + random_interval
+
+      retry_after && retry_after >= interval ? retry_after : interval
```
iMacTia: This method looks a bit too complicated now, so I would separate the logic to calculate the retry interval into a separate method (similarly to what you did with `calculate_retry_after`).
This new method could be called `calculate_retry_interval(retries)`.

test/middleware/retry_test.rb
```diff
@@ -95,30 +112,30 @@ def sleep_amount(retries)
 
     def test_exponential_backoff
       middleware = Faraday::Request::Retry.new(nil, :max => 5, :interval => 0.1, :backoff_factor => 2)
-      assert_equal middleware.sleep_amount(5), 0.1
-      assert_equal middleware.sleep_amount(4), 0.2
-      assert_equal middleware.sleep_amount(3), 0.4
+      assert_equal middleware.sleep_amount(5, {}), 0.1
```
iMacTia: We can use the new `calculate_retry_interval` for this and the following tests so we don't need to pass `{}` every time.

test/middleware/retry_test.rb
```diff
@@ -212,5 +229,42 @@ def test_should_rewind_files_on_retry
...
+
+    def test_interval_if_retry_after_present
+      started = Time.now
+
+      assert_raises(Faraday::Error::RetriableResponse) {
+        params = { :retry_after => 0.5 }
+        conn(:max => 1, :interval => 0.1, :retry_statuses => [429]).get("/throttled", params)
+      }
+
+      assert_in_delta 0.5, Time.now - started, 0.04
```
iMacTia: I'm a bit worried by this assertion, it might create false-positives if test execution is too slow.
Any reason why we don't simply test that `Time.now - started > 0.5`?

lib/faraday/request/retry.rb
bronislav: I've tried to do this without raising an exception but this will lead to implementing two places with retry logic which seems a bit overhead. It could be done using `do ... while` loop but it looks hacky for me.

And I'll move statuses testing to the better place.

test/middleware/retry_test.rb
bronislav: Agree. We don't care about delta in this case.

lib/faraday/request/retry.rb
```diff
@@ -106,29 +112,31 @@ def initialize(app, options = nil)
...
           retries -= 1
           rewind_files(request_body)
           @options.retry_block.call(env, @options, retries, exception)
-          sleep sleep_amount(retries + 1)
+          sleep sleep_amount(retries + 1, env)
           retry
         end
-        raise
+
+        raise unless exception.is_a?(Faraday::Error::RetriableResponse)
```
bronislav: @iMacTia I've silenced exception here, but not sure that this is acceptable and very good solution

iMacTia:
I agree this is not the most elegant solution, but it seems the easiest one without refactoring too much.
I'll try to think of a better way to do this.

bronislav:
I've added support for timestamp in the Retry-After header according to MDN https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

I've also stop retrying if `retry_after` interval is higher than `max_interval`

lib/faraday/request/retry.rb
bronislav: I've tried to move this line around but it seems that this is the best option.

bronislav:
@iMacTia Any chance to merge it soon?

bronislav:
@iMacTia Thank you.

iMacTia:
@bronislav Merged into master, will probably release next week as this weekend I'm unavailable in case of issues 😅
If you get a chance to test master branch in a real-world application than feedback would be welcome 😃

bronislav:
Yes. I'm going to use it from master and test toward real project tomorrow. Will let you know.

bronislav:
I found one configuration issue: if user enable `:raise_error` middleware or raise custom exceptions on particular responses, those exceptions should be included in the retry middleware configuration explicitly.

iMacTia:
@bronislav that should be in line with what happens currently and depends on how you set the middlewares in your stack, but thanks for raising this 👍

lib/faraday/request/retry.rb
```diff
@@ -106,29 +112,35 @@ def initialize(app, options = nil)
...
-          sleep sleep_amount(retries + 1)
-          retry
+          if (sleep_amount = calculate_sleep_amount(retries + 1, env))
+            sleep sleep_amount
+            retry
+          end
         end
-        raise
+
+        raise unless exception.is_a?(Faraday::Error::RetriableResponse)
```
fabiokr: @bronislav @iMacTia Hey all, I've tried to use `:retry_statuses`, but it seems that this portion makes it return a nil response and no exceptions are raised when it gives up retrying the `RetriableResponse`, which was unexpected for me. Maybe in that case it would make more sense to return the last `resp`, and then let the user handle the response as he sees fit?

bronislav: @fabiokr I see what you mean. Would you mind to create a pull request for this change?

iMacTia: You're right @fabiokr, we missed what to do in case we reach this point (too many retries) with a `Faraday::Error::RetriableResponse`. I agree in that case we should simply return the response, which should be available in the exception.

PR with demonstrating test and fix welcome!

fabiokr: @bronislav @iMacTia https://github.com/lostisland/faraday/pull/783

vvo:
Hi there, should we document the fact that retry-after headers are already handled? this was a surprise for me and I guess we could tell somewhere that it's already handled? thanks! (for example, here: https://www.rubydoc.info/gems/faraday/Faraday/Request/Retry)

iMacTia:
@vvo good point, #1109 takes care of that 👍

PR #777, opened 2018-03-02
Make frozen-string-literal safe

ankane:
Hi, thanks for this great gem!

This PR makes Faraday works with RUBYOPT="--enable-frozen-string-literal". Here's a good article explaining it: https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html

### DIFF
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -55,7 +55,7 @@ module Faraday
       end
 
       # The params have form [['key1', 'value1'], ['key2', 'value2']].
-      buffer = ''
+      buffer = String.new
       params.each do |parent, value|
         encoded_parent = escape(parent)
         buffer << "#{to_query.call(encoded_parent, value)}&"
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -55,7 +55,7 @@ module Faraday
       end
 
       # The params have form [['key1', 'value1'], ['key2', 'value2']].
-      buffer = String.new
+      buffer = String.new('')
       params.each do |parent, value|
         encoded_parent = escape(parent)
         buffer << "#{to_query.call(encoded_parent, value)}&"


iMacTia:
Hi @ankane,
thanks for raising this point.
I'm also trying to experiment with frozen string literals in my newest projects so appreciate proposing this for Faraday as well.
However, I would like to get this right if we're trying got make it "Frozen-String-Literal Friendly".

First of all, how did you identify that line of code that you updated? How do you know there aren't others out there?

I naively tried to execute `RUBYOPT="--enable-frozen-string-literal" script/test`, but got immediately stuck with SimpleCov:

```
/simplecov-0.14.1/lib/simplecov/version.rb:4:in `<module:SimpleCov>': can't define singleton (TypeError)
```

ankane:
Hey @iMacTia, I discovered it while making one of my projects (Searchkick) frozen string literal safe. Here's a minimal example to repro:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "faraday"
end

p Faraday.get "http://sushi.com/nigiri/sake.json", {test: "t"}
```

And run:

```sh
RUBYOPT="--enable-frozen-string-literal" ruby repro.rb
```

iMacTia:
Thanks for clarifying @ankane .
Unfortunately that's not what I expected though. This way we have no guarantee that Faraday will be frozen-string-literal safe, as your empirical test doesn't cover 100% of the code.

In order to test that properly, we'd need to run tests with the flag enabled as I suggested in my previous comment.
I commented out `require 'simplecov'` and its configuration from `/faraday/test/helper.rb` and the result is in fact much worse:

```
581 runs, 682 assertions, 5 failures, 236 errors, 0 skips
```

And as you can imagine, most of the errors are `RuntimeError: can't modify frozen String`.
Only fixing all these errors will really mean making Faraday frozen-string-literal safe

ankane:
From your last comment, think I misinterpreted what you were saying/asking for previously. I agree with everything you mention for long-term sustainability. This PR was an attempt to move that direction, but I understand wanting to do it together instead of piece-wise.

Issue #776, opened 2018-02-28
Persistent connections are not faster then regular connections ?


grosser:
```
require 'faraday'

def bench(name)
  x = Benchmark.realtime do
    100.times do
      yield
    end
  end
  puts "#{name}: #{x}"
end

bench("normal") do
  Faraday.get("https://api.github.com/rate_limit", {}, authorization: "token #{ENV["GITHUB_TOKEN"]}")
end

conn = Faraday.new("https://api.github.com") do |r|
  r.adapter :net_http_persistent
end
bench("persistent") do
  conn.get("/rate_limit", {}, authorization: "token #{ENV["GITHUB_TOKEN"]}")
end
http = Net::HTTP::Persistent.new
bench("raw") do
  http.request URI "https://api.github.com/rate_limit"
end

```

```
normal: 28.042679999955
persistent: 28.48966299998574
raw: 17.34108699997887
```

am I doing something wrong or is that expected ?

iMacTia:
Hi @grosser,

I believe someone raised at some point the issue that the current adapter for `Net::HTTP::Persistent` is re-creating the connection on every request, which of course doesn't make any sense...

I promised myself I'd look into this while moving it outside into a separate gem for v1.0, but if you want to try a quick-fix then it's more than welcome

iMacTia:
This was addressed in #778

PR #778, opened 2018-03-02
reuse persistent connections when using net-http-persistent

grosser:
https://github.com/lostisland/faraday/issues/776

/cc @iMacTia 

```
bundle console
require 'benchmark'

before:
>> Benchmark.realtime { 20.times { Faraday.get 'https://g.co/sddfsfds' } }
=> 4.300932000041939

after:
>> Benchmark.realtime { 20.times { Faraday.get 'https://g.co/sddfsfds' } }
=> 3.4228649999713525
```

con:
 - user could change ssl settings and would still use the old connection
 - memory leak when using tons of different urls/proxy settings

pro:
 - not a bunch of changes all over the place
 - "somewhat working" instead of "completely broken"

Moneky-patch:

```Ruby
Faraday::Adapter::NetHttpPersistent.prepend(Module.new do
  def net_http_connection(env)
    (@cached_connection ||= {})[env[:url].host] ||= super
  end
end)
```

### DIFF
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -1,13 +1,23 @@
-# Rely on autoloading instead of explicit require; helps avoid the "already
-# initialized constant" warning on Ruby 1.8.7 when NetHttp is refereced below.
-# require 'faraday/adapter/net_http'
-
 module Faraday
   class Adapter
     class NetHttpPersistent < NetHttp
       dependency 'net/http/persistent'
 
+      private
+
       def net_http_connection(env)
+        proxy_uri = proxy_uri(env)
+
+        cached_connection env[:url], proxy_uri do
+          if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
+            Net::HTTP::Persistent.new(name: 'Faraday', proxy: proxy_uri)
+          else
+            Net::HTTP::Persistent.new('Faraday', proxy_uri)
+          end
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


iMacTia:
Hi @grosser and thanks for taking the time to work on this.
I agree, it doesn't look like the most elegant of the solutions, but at least solves the issue.
I think this will need to be properly addressed once the adapter moves to it's own separate gem for v1.0, but this is a good workaround.

I just can't understand why did you post the monkey-patch at the end.
Is that needed as well to make the fix work? Or is it just a temporary solution you put in place?

grosser:
temporary solution until this is released and also for others to verify that this actually improves the speed without having to go through installing from github :)

lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -1,29 +1,37 @@
...
         proxy_uri = nil
         if (proxy = env[:request][:proxy])
           proxy_uri = ::URI::HTTP === proxy[:uri] ? proxy[:uri].dup : ::URI.parse(proxy[:uri].to_s)
           proxy_uri.user = proxy_uri.password = nil
           # awful patch for net-http-persistent 2.8 not unescaping user/password
-          (class << proxy_uri; self; end).class_eval do
+          (
+          class << proxy_uri;
+            self;
+          end).class_eval do
```
iMacTia: This looks like an automatic indentation fix.
If you do indent the code this way, then you don't need the `;` and `()` anymore

grosser: fixed

PR #784, opened 2018-03-24
Improve response logger middleware output

### LINKED ISSUE #782: Improve logger middleware output
## Basic Info

* Faraday Version: 0.14.0
* Ruby Version: 2.5.0

## Issue description
I am currently using the Faraday response logger middleware in some projects, however the output is a bit confusing because it does not follow a similar structure for request/response log lines. Right now here is what you get in your logs if you make a request with this middleware enabled:

```
get https://api.example.com 
Status: 200
```
This might get even more confusing depending on you log formatter settings. I propose to follow the same structure for request/response logs, in a way that each line can be clearly inspected later. It would then look like this:

```
 request: GET https://api.example.com 
response: Status 200
```

Where the `request` and `response` strings are set as the `progname` of the INFO log calls accordingly, just like the debug lines currently work.

I've already made a branch with this implementation and added the required tests, if you like it I can open a PR.

erichmachado:
## Description

Currently, log output looks like this with the firmware enabled:

```
get https://api.example.com
Status: 200
```

This PR changes the log output to be formatted like this instead:

```
 request: GET https://api.example.com
response: Status 200
```

Where the `request` and `response` strings are set as the progname of the `INFO` log calls accordingly, just like the debug lines currently work.

Fixes #782

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -20,7 +20,7 @@ module Faraday
     def_delegators :@logger, :debug, :info, :warn, :error, :fatal
 
     def call(env)
-      info('request')  { "#{env.method} #{apply_filters(env.url.to_s)}" }
+      info('request')  { "#{env.method.upcase} #{apply_filters(env.url.to_s)}" }
       debug('request') { apply_filters( dump_headers env.request_headers ) } if log_headers?(:request)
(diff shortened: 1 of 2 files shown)


huangyq23:
This looks great! Do you plan to release a new gem version?

iMacTia:
Yeah plan was to release tomorrow, but it may actually be easier today.
I'm creating the tag now, so expect to see the release later today 👍

PR #792, opened 2018-04-24
Allow setting min/max SSL version for a Net::HTTP connection on Ruby 2.5

bdewater:
The functionality in the OpenSSL gem was introduced in ruby/openssl#142 and supported by Net::HTTP in Ruby 2.5: https://github.com/ruby/ruby/commit/dcea9198a9d80bdf4eeacd9d9e9d883850a4a8d2

An example why this might be useful; for payment data the PCI DSS [mandates](https://blog.pcisecuritystandards.org/are-you-ready-for-30-june-2018-sayin-goodbye-to-ssl-early-tls) that TLS 1.1 or newer is used after June 30. Using `ssl_version` would disallow the client negotiating TLS 1.2 (or 1.3 in the near future) if both sides support it, `min_version` doesn't have this problem.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -106,6 +106,8 @@ module Faraday
         http.ca_path      = ssl[:ca_path]      if ssl[:ca_path]
         http.verify_depth = ssl[:verify_depth] if ssl[:verify_depth]
         http.ssl_version  = ssl[:version]      if ssl[:version]
+        http.min_version  = ssl[:min_version]  if ssl[:min_version]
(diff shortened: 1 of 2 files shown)


iMacTia:
Hi @bdewater, this makes sense, but unfortunately it won't work unless you add the new fields to the [SSLOptions](https://github.com/lostisland/faraday/blob/e6311bd286de8df1a585b2d5a1eeb150f6d067bf/lib/faraday/options.rb#L216) class.

I'd also ask you to add some tests to ensure we're not missing anything else

bdewater:
Hey @iMacTia thanks for the pointer, I didn't have the time to dig deeper yesterday but CI is green now :)

PR #793, opened 2018-04-26
NetHttpPersistent adapter reuse SSL connections

### LINKED ISSUE #791: Faraday::Adapter::NetHttpPersistent does not reuse requests for SSL with Net::HTTP::Persistent < 3.0.0
## Basic Info

* Faraday Version: 1.15.0, 1.13.1, 1.12.2
* Net::HTTP::Persistent version: 2.9.4
* Ruby Version: 2.2.3

## Issue description
Faraday will use new connections for every request with the `NetHttpPersistent` adapter.

This is because `Faraday::Adapter::NetHttp#call` will call `configure_ssl` for every request made https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http.rb#L34.

`Faraday::Adapter::NetHttpPersistent#configure_ssl` calls multiple methods on the `Net::HTTP::Persistent` instance that internally call `Net::HTTP::Persistent#reconnect_ssl` https://github.com/lostisland/faraday/blob/v0.15.0/lib/faraday/adapter/net_http_persistent.rb#L49.

`reconnect_ssl` increments a counter called `@ssl_generation` which is used as a key for the hash of available connections https://github.com/drbrain/net-http-persistent/blob/v2.9.4/lib/net/http/persistent.rb#L604.

Because this counter is incremented for every request, connections cannot be reused.

## Steps to reproduce
```ruby
require "net/http/persistent"
require "faraday"

connection = Faraday.new("https://www.google.com") do |f|
  f.adapter :net_http_persistent
end

module Hack
  def request(*args)
    puts "request #{object_id}"
    super
  end

  def initialize(*args)
    puts "initialize #{object_id}"
    super
  end
end

module PersistentHack
  def reconnect_ssl(*args)
    puts "reconnect_ssl #{@ssl_generation}"
    super
  end
end

Net::HTTP.send(:prepend, Hack)
Net::HTTP::Persistent.send(:prepend, PersistentHack)

puts "request 1"
connection.get("")

puts "request 2"
connection.get("")
```

Output
```
request 1
reconnect_ssl 0
reconnect_ssl 1
initialize 7789240
request 7789240
request 2
reconnect_ssl 2
reconnect_ssl 3
initialize 7744280
request 7744280
```
If we force `Net::HTTP::Persistent#reconnect_ssl` to no-op instead of incrementing the counter instead we get
```
request 1
initialize 13027780
request 13027780
request 2
request 13027780
```
which clearly shows the connection being reused.

katsuya94:
## Description
Fixes #791 by only setting SSL attributes on the `Net::HTTP::Persistent` instance when params are changed.

@iMacTia this is a (not working yet) first stab at fixing the issue. The problem I'm running into is in order to determine when to actually update SSL params I need to compare the current and new values. Unfortunately some of these values like `cert_store` are of non-comparable types like `OpenSSL::X509::Store`. Any ideas?

Update: fixed issues with object comparison by caching the blank cert store.

### DIFF
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -6,15 +6,16 @@ module Faraday
       private
 
       def net_http_connection(env)
-        proxy_uri = proxy_uri(env)
-
         cached_connection do
           if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
-            Net::HTTP::Persistent.new(name: 'Faraday', proxy: proxy_uri)
+            Net::HTTP::Persistent.new(name: 'Faraday')
           else
-            Net::HTTP::Persistent.new('Faraday', proxy_uri)
+            Net::HTTP::Persistent.new('Faraday')
           end
         end
+        proxy_uri = proxy_uri(env)
+        cached_connection.proxy = proxy_uri if cached_connection.proxy_uri != proxy_uri
+        cached_connection
       end
 
       def proxy_uri(env)
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -8,7 +8,7 @@ module Faraday
       def net_http_connection(env)
         proxy_uri = proxy_uri(env)
 
-        cached_connection env[:url], proxy_uri do
+        cached_connection do
           if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
             Net::HTTP::Persistent.new(name: 'Faraday', proxy: proxy_uri)
           else
@@ -46,17 +46,23 @@ module Faraday
       end
 
       def configure_ssl(http, ssl)
-        http.verify_mode  = ssl_verify_mode(ssl)
-        http.cert_store   = ssl_cert_store(ssl)
+        http_set(http, :verify_mode, ssl_verify_mode(ssl))
+        http_set(http, :cert_store,  ssl_cert_store(ssl))
 
-        http.certificate  = ssl[:client_cert]  if ssl[:client_cert]
-        http.private_key  = ssl[:client_key]   if ssl[:client_key]
-        http.ca_file      = ssl[:ca_file]      if ssl[:ca_file]
-        http.ssl_version  = ssl[:version]      if ssl[:version]
+        http_set(http, :certificate, ssl[:client_cert]) if ssl[:client_cert]
+        http_set(http, :private_key, ssl[:client_key])  if ssl[:client_key]
+        http_set(http, :ca_file,     ssl[:ca_file])     if ssl[:ca_file]
+        http_set(http, :ssl_version, ssl[:version])     if ssl[:version]
       end
 
-      def cached_connection(url, proxy_uri)
-        (@cached_connection ||= {})[[url.scheme, url.host, url.port, proxy_uri]] ||= yield
+      def http_set(http, attr, value)
+        if http.sent(attr) != value
+          http.send("#{attr}=", value)
+        end
+      end
+
+      def cached_connection
+        @cached_connection ||= yield
       end
     end
   end
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -56,7 +56,7 @@ module Faraday
       end
 
       def http_set(http, attr, value)
-        if http.sent(attr) != value
+        if http.send(attr) != value
           http.send("#{attr}=", value)
         end
       end
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -28,6 +28,11 @@ module Faraday
       NET_HTTP_EXCEPTIONS << OpenSSL::SSL::SSLError if defined?(OpenSSL)
       NET_HTTP_EXCEPTIONS << Net::OpenTimeout if defined?(Net::OpenTimeout)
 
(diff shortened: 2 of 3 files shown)


lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -46,17 +46,23 @@ def perform_request(http, env)
...
+        http_set(http, :certificate, ssl[:client_cert]) if ssl[:client_cert]
+        http_set(http, :private_key, ssl[:client_key])  if ssl[:client_key]
+        http_set(http, :ca_file,     ssl[:ca_file])     if ssl[:ca_file]
+        http_set(http, :ssl_version, ssl[:version])     if ssl[:version]
       end
 
-      def cached_connection(url, proxy_uri)
-        (@cached_connection ||= {})[[url.scheme, url.host, url.port, proxy_uri]] ||= yield
+      def http_set(http, attr, value)
+        if http.sent(attr) != value
```
iMacTia: This should probably be `send` ?

lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -46,17 +46,23 @@ def perform_request(http, env)
...
+          http.send("#{attr}=", value)
+        end
+      end
+
+      def cached_connection
+        @cached_connection ||= yield
```
iMacTia: Removing the hash caching, we're going to reuse the same connection even though the url changes. Or am I missing something?

katsuya94: `Net::HTTP::Persistent` has logic for using different connections for different URI bases (including differences in proxy) internally.

katsuya94: in 2.9.4 https://github.com/drbrain/net-http-persistent/blob/v2.9.4/lib/net/http/persistent.rb#L613-L619
in 3.0.0 https://github.com/drbrain/net-http-persistent/blob/v3.0.0/lib/net/http/persistent.rb#L608-L611

katsuya94:
@iMacTia Any thoughts on the updates?

iMacTia:
I’m really sorry @katsuya94, really wanted to have another look but I still haven’t got the time :(
Will do my best to review them reasonably soon as I would like to include this into the next release!

katsuya94:
No worries! Take your time.

iMacTia:
@katsuya94 I had a quick look at changes and they look good to me!

The only thing I'm not completely sure about is the comeback on relying on `Net::HTTP::Persistent` caching of connection.
In #778 a fix was introduced that sensibly improved performances by managing the cache on our side.

I would like to re-run the same tests @grosser did on that PR to see if we're regressing on that side.
If you have some time, would you like to run it? Otherwise I'll do as soon as I can.

katsuya94:
net-http-persistent 2.9.4
```
faraday/ (net_http_persistent_fix|✔) % git checkout net_http_persistent_fix
faraday/ (net_http_persistent_fix|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
3.403531900999951
faraday/ (net_http_persistent_fix|✔) % git checkout origin/master
faraday/ (:4381c2d|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
4.028428414000018
faraday/ (:4381c2d|✔) % git checkout 6ddf9c3b80d2eee6673c72d82d181a810746088c
faraday/ (:6ddf9c3|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
3.9857757019999553
```

net-http-persistent 3.0.0
```
faraday/ (:6ddf9c3|✔) % git checkout net_http_persistent_fix
faraday/ (net_http_persistent_fix|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
3.5376455999999052
faraday/ (net_http_persistent_fix|✔) % git checkout origin/master
faraday/ (:4381c2d|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
5.190602235999904
faraday/ (:4381c2d|✔) % git checkout 6ddf9c3b80d2eee6673c72d82d181a810746088c
faraday/ (:6ddf9c3|✔) % ruby -Ilib -e 'require "faraday"; require "benchmark"; puts Benchmark.realtime { 20.times { Faraday.get "https://g.co/sddfsfds" } }'
3.96499722700014
```

6ddf9c3b80d2eee6673c72d82d181a810746088c was the commit before #778 was merged

iMacTia:
@katsuya94 Thanks for running the tests!
I'm surprised master is actually the slowest (even slower than before the fix!).
Anyway, the important thing is that your branch yields the best results.
And that tests are all green, of course 😃

lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -46,17 +47,23 @@ def perform_request(http, env)
...
+
+      def http_set(http, attr, value)
+        if http.send(attr) != value
+          http.send("#{attr}=", value)
+        end
       end
 
-      def cached_connection(url, proxy_uri)
-        (@cached_connection ||= {})[[url.scheme, url.host, url.port, proxy_uri]] ||= yield
+      def cached_connection
```
grosser: might as well inline it at that point :)

PR #794, opened 2018-04-27
Logger middleware: allow redefined STDOUT

Aupajo:
`STDOUT` is a constant that represents the standard output. `$stdout` is a global constant that is _usually_ set to `STDOUT`, but can be redefined if desirable.

This change continues to support logging to the standard output, but will also play nicely with users who chose to redefine `$stdout`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Since the default middleware stack uses url\_encoded middleware and default adap
 ```ruby
 conn = Faraday.new(:url => 'http://sushi.com') do |faraday|
   faraday.request  :url_encoded             # form-encode POST params
-  faraday.response :logger                  # log requests to STDOUT
+  faraday.response :logger                  # log requests to $stdout
   faraday.adapter  Faraday.default_adapter  # make requests with Net::HTTP
 end
 
@@ -79,7 +79,7 @@ response.body
 
 conn.get '/nigiri', { :name => 'Maguro' }   # GET http://sushi.com/nigiri?name=Maguro
 
-conn.get do |req|                           # GET http://sushi.com/search?page=2&limit=100  
+conn.get do |req|                           # GET http://sushi.com/search?page=2&limit=100
   req.url '/search', :page => 2
   req.params['limit'] = 100
 end
@@ -118,7 +118,7 @@ conn.get do |req|
   req.options.context = {
       foo: 'foo',
       bar: 'bar'
-  }          
+  }
 end
 ```
 
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -10,7 +10,7 @@ module Faraday
       super(app)
       @logger = logger || begin
         require 'logger'
-        ::Logger.new(STDOUT)
+        ::Logger.new($stdout)
       end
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ Since the default middleware stack uses url\_encoded middleware and default adap
 ```ruby
 conn = Faraday.new(:url => 'http://sushi.com') do |faraday|
   faraday.request  :url_encoded             # form-encode POST params
-  faraday.response :logger                  # log requests to STDOUT
+  faraday.response :logger                  # log requests to $stdout
```
Aupajo: Also: I find this example confusing. `faraday.response` logs _requests_ to standard out? Was this a mistake?

iMacTia:
Hi @Aupajo, thanks for raising this.
To be honest, I've always used `STDOUT` and usually been told to stay away from the `$...` variables in Ruby. That's not a justification for rejecting this PR, I just want to understand things better first.

If I look at the official Ruby Doc for [Logger](https://ruby-doc.org/stdlib-2.5.0/libdoc/logger/rdoc/Logger.html), all examples uses `STDOUT` instead of `$stdout`.

I understand using the latter will make the gem more "friendly" and give you the possibility to change the stdout, but are we sure this is actually a good practice? Do you have any supporting link/documentation I could look at?

Aupajo:
@iMacTia Sure. From [this StackOverflow answer](https://stackoverflow.com/a/6671770):

>  `$stdout` is a global variable that represents the current standard output. `STDOUT` is a constant representing standard output and is typically the default value of `$stdout`.
>
> With `STDOUT` being a constant, you shouldn't re-define it, however, you can re-define `$stdout` without errors/warnings (re-defining `STDOUT` will raise a warning). for example, you can do:
>
>      $stdout = STDERR
>
> Same goes for `$stderr` and `STDERR`.

The advice to stay away from the `$stdout` variable comes from the hazard that redefining it can lead to some surprising outcomes (“Where did my `puts` go?”).

However, there are some cases where you may want to redefine `$stdout`. For instance, [RSpec redefines `$stdout`](https://github.com/rspec/rspec-expectations/blob/add9b271ecb1d65f7da5bc8a9dd8c64d81d92303/lib/rspec/matchers/built_in/output.rb#L132-L149) in order to make the `expect { ... }.to output(/something/).to_stdout` matcher work.

In the application I'm working on, we similarly redefine `$stdout` for the purposes of setting expectations on log messages.

All in all: _redefining_ `$stdout` is something that should be handled with care, but _using_ it should be the default for all libraries. If a user wants to alter their `$stdout`, they should be allowed to do so, and libraries should respect that choice.

README.md
iMacTia: That's a fair point. Reality is that logger middleware logs both requests and responses, but for some reason has been historically created as a response middleware.
The example description can probably be changed into `# log requests and responses to $stdout`.
However we're going to do some major documentation refactoring with v1.0, so happy to hold this for now and address it later

iMacTia:
Cool, it's seems the Logger is the only place where we use `STDOUT`.
LGTM

PR #795, opened 2018-05-01
Allow setting min/max ssl version for Excon

bdewater:
After https://github.com/excon/excon/pull/672 I figured I might as well create this companion PR to https://github.com/lostisland/faraday/pull/792 😄


lib/faraday/adapter/excon.rb
```diff
@@ -15,6 +15,9 @@ def call(env)
           opts[:client_key]  = ssl[:client_key]  if ssl[:client_key]
           opts[:certificate] = ssl[:certificate] if ssl[:certificate]
           opts[:private_key] = ssl[:private_key] if ssl[:private_key]
+          opts[:ssl_version] = ssl[:ssl_version] if ssl[:ssl_version]
```
iMacTia: There's probably a typo here, should be `ssl[:version]`.

bdewater: Woops thanks!

bdewater:
Hey @iMacTia the upstream changes in Excon and Net::HTTP::Persistent (PR outstanding) were easy since to since it was so similar to what was done in Net::HTTP. 

But I don't have the time to fix them all. Typhoeus looks not too hard but no promises - I was kind of hoping support in Faraday might inspire gem authors to add the feature itself to keep up with the competition 😄 

Anybody reading along: feel free to create the appropriate PRs yourself.

PR #797, opened 2018-05-07
Refactor: inline private method NetHttpPersistent#cached_connection

olleolleolle:
## Description

In a comment in #793, @grosser noted an opportunity to inline away a method.

This PR inlines that method.


katsuya94:
Thanks! That was an easy thing I could have done 😅

PR #799, opened 2018-05-17
Set max_retries to 0 if available.

### LINKED ISSUE #774: Net::HTTP adapter retries requests automatically
## Basic Info

* Faraday Version: any
* Ruby Version: 2.5+

## Issue description
When using `Net::HTTP` as adapter, users get surprised by the retry strategy provided by `Net::HTTP`. This brings to different "issues" (e.g. #612 and #771).
There's not much we can do for Ruby <= 2.4, however Ruby 2.5 provides a new `max_retries` config for `Net::HTTP` so Faraday should set that automatically when available.

## Steps to reproduce
See issues #612 and #771.

michielboekhoff:
## Description
This PR sets the `max_retries` on `Net::Http` to zero where supported (Ruby 2.5+). This fixes #774. 

## Todos
- [ ] Documentation - maybe? We might need to document this is only since Ruby 2.5+.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -118,6 +118,8 @@ module Faraday
       def configure_request(http, req)
         http.read_timeout = http.open_timeout = req[:timeout] if req[:timeout]
         http.open_timeout = req[:open_timeout]                if req[:open_timeout]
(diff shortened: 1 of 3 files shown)


michielboekhoff:
Just checked the Travis results - JRuby seems to be failing, but the failing tests seem to have nothing to do with the code in question. Is this a different issue or related?

PR #801, opened 2018-05-23
#800 correct handling parameter value empty array

### LINKED ISSUE #800: empty array parameter value encodes incorrectly
## Basic Info

parameters like `{ foo: { bar: [] } }` and `{ foo: [] }` encoded incorrectly

it's just being skipped. I don't think this is right behavior

related to https://github.com/JsonApiClient/json_api_client/issues/278

* Faraday Version:
0.15.1 and master github branch
* Ruby Version:
2.2.4

## Steps to reproduce
If possible, please provide the steps to reproduce the issue.
```ruby
>> Faraday::NestedParamsEncoder.encode(foo: { bar: [] }, baz: [])
=> "&"
```

I think correct behavior will be to encode it to `"baz%5B%5D&foo%5Bbar%5D%5B%5D"` or `"baz%5B%5D=&foo%5Bbar%5D%5B%5D="` or `"baz&foo%5Bbar%5D"`, `"baz=&foo%5Bbar%5D="`

senid231:
## Description
correct handling of parameters when value is an empty array
fix #800

## Todos
- [x] Tests
- [ ] Documentation *(not sure where it should be documented)*

## Additional Notes
Optional section
will fix https://github.com/JsonApiClient/json_api_client/issues/278

### DIFF
--- a/lib/faraday/parameters.rb
+++ b/lib/faraday/parameters.rb
@@ -40,9 +40,10 @@ module Faraday
           end
           return buffer.chop
         elsif value.is_a?(Array)
+          new_parent = "#{parent}%5B%5D"
+          return new_parent if value.empty?
           buffer = ""
           value.each_with_index do |val, i|
-            new_parent = "#{parent}%5B%5D"
             buffer << "#{to_query.call(new_parent, val)}&"
           end
           return buffer.chop
--- a/test/parameters_test.rb
+++ b/test/parameters_test.rb
@@ -138,4 +138,20 @@ class TestParameters < Faraday::TestCase
     assert_equal expected.split("&").sort,
       Faraday::Utils.unescape(Faraday::NestedParamsEncoder.encode(params)).split("&").sort
   end
+
+  def test_encode_empty_string_array_value
+    expected = 'baz=&foo%5Bbar%5D='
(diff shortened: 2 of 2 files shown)


lib/faraday/parameters.rb
```diff
@@ -39,13 +39,15 @@ def self.encode(params)
...
           return buffer.chop
-        elsif value.is_a?(Array)
+        elsif value.is_a?(Array) && value.size != 0
           buffer = ""
           value.each_with_index do |val, i|
             new_parent = "#{parent}%5B%5D"
             buffer << "#{to_query.call(new_parent, val)}&"
           end
           return buffer.chop
+        elsif value.is_a?(Array) && value.size == 0
```
olleolleolle: Perhaps the `Array#empty?` method is even clearer to read?

iMacTia: I'd go a step further and merge this 2 branches:

```ruby
elsif value.is_a?(Array)
  return "#{parent}%5B%5D" if value.empty? # This is the only new line
  buffer = ""
  value.each_with_index do |val, i|
    new_parent = "#{parent}%5B%5D"
     buffer << "#{to_query.call(new_parent, val)}&"
  end
  return buffer.chop
```

senid231: nice catch
fixed

iMacTia:
Nice optimisation, looks good now.

PR #803, opened 2018-05-26
Make Faraday::Request serialisable with Marshal.

### LINKED ISSUE #535: Requests can no longer be converted to simple JSON
`fba7ce503d2be7bd88229f95107559a14c98821c` causes a stack too deep to occur when attempting to call `to_json`, due to the nested `env`/`request` nature of the objects.

Some context: We have been logging the json version of faraday response objects for verification purposes when performing sensitive data changes through faraday requests; This is no longer possible.

michielboekhoff:
## Description
This partially aids #535, or rather the over-arching need for serialisation of `Faraday::Request`. I didn't include any JSON serialisation in this on purpose, as I believe it needs more discussion.

### DIFF
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -69,6 +69,26 @@ module Faraday
       headers[key] = value
     end
 
+    def marshal_dump
+      {
+        :method  => method,
+        :body    => body,
+        :headers => headers,
+        :path    => path,
+        :params  => params,
+        :options => options
+      }
+    end
+
+    def marshal_load(serialised)
+      self.method  = serialised[:method]
+      self.body    = serialised[:body]
+      self.headers = serialised[:headers]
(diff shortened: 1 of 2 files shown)


iMacTia:
Thanks @michielboekhoff for taking a shot at this.
No worries for JSON support, we can do one thing at a time 👍
Could you just please rename the existing `test_marshal` into `test_marshal_response` for consistency?

michielboekhoff:
Thanks for the quick reply, @iMacTia. I've changed the test name to `test_marshal_response`. Concerning JSON serialisation, could we have a chat about that in #535?

iMacTia:
@michielboekhoff thanks, merged 👍 
And yes we can talk in #535 about JSON serialisation, although I doubt we can do something reversible like YAML and Marshal. I'd also like to understand the use-case for that before reviewing an implementation

Issue #813, opened 2018-08-21
Faraday::Request::Retry default exceptions as a constant


krmannix:
Not reporting a bug or issue, so eschewing the template per instructions.

We want to add a few extra exceptions to the default list of exceptions provided by `Request::Retry`. However, this leads us to using `Faraday::Request::Retry::Options.new.exceptions` to fetch the default list of exceptions to work with. Alternatively, creating a custom list of retriable methods in addition to the default list seems quite natural (`Faraday::Request::Retry::IDEMPOTENT_METHODS`).

I'd like to make the list of default exceptions a constant on `Faraday::Request::Retry`, but per contribution instructions wanted to create an issue here first in case this has already been discussed.

My implementation would result in custom exceptions arguments like below:

```
Faraday::Connection.new(url, options) do |conn|
  conn.request :retry, max: 1, interval: 0.5,
                       exceptions: Faraday::Request::Retry::EXCEPTIONS + [Faraday::ConnectionFailed]
end
```

iMacTia:
Hi @krmannix, makes sense to me, thanks for the contribution!

The only thing I would change is the name of the constant into `DEFAULT_EXCEPTIONS` to make it very clear it's just the default value, as we do for `DEFAULT_CHECK`.

iMacTia:
Closed by #814

PR #814, opened 2018-08-22
Add DEFAULT_EXCEPTIONS constant to Request::Retry

krmannix:
## Description
Adds a `DEFAULT_EXCEPTIONS` constant to `Request::Retry`, which allows easy access much like the `IDEMPOTENT_METHODS` constant

## Todos
None

## Additional Notes
Initial request in [Issue 813](https://github.com/lostisland/faraday/issues/813)

### DIFF
--- a/lib/faraday/request/retry.rb
+++ b/lib/faraday/request/retry.rb
@@ -20,6 +20,7 @@ module Faraday
   #
   class Request::Retry < Faraday::Middleware
 
+    DEFAULT_EXCEPTIONS = [Errno::ETIMEDOUT, 'Timeout::Error', Error::TimeoutError, Faraday::Error::RetriableResponse]
     IDEMPOTENT_METHODS = [:delete, :get, :head, :options, :put]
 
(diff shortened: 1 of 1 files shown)


lib/faraday/request/retry.rb
```diff
@@ -20,6 +20,7 @@ module Faraday
   #
   class Request::Retry < Faraday::Middleware
 
+    DEFAULT_EXCEPTIONS = [Errno::ETIMEDOUT, 'Timeout::Error', Error::TimeoutError, Faraday::Error::RetriableResponse]
```
krmannix: One potential addition is calling `.freeze` on this array, as exposing this list rather than constructing it on each call to `exceptions` means it's now modifiable by users. That may be okay though.

iMacTia: Good point @krmannix.
However being this a constant, I would rather `.freeze` it and make it immutable as you suggested.
If we want to improve the way you simply add new exceptions to the list, we should probably revisit the API and provide a better way to do it

krmannix: great! I'll update. Thanks for the quick response!

Issue #833, opened 2018-11-23
Expose `pool_size` as a config option for net_http_persistent


col:
## Basic Info

* Faraday Version: 0.15.3
* Ruby Version: jruby-9.1.17.0

## Issue description
The constructor for Net::HTTP::Persistent contains a `pool_size` param. There is currently no way to set this when using Faraday.

Net::HTTP::Persistent constructor definition here:
https://github.com/drbrain/net-http-persistent/blob/0e5a8fb8a27deff37247ddb3bc5e6c9e390face5/lib/net/http/persistent.rb#L505 
```ruby
def initialize name: nil, proxy: nil, pool_size: DEFAULT_POOL_SIZE
  ...
end
```

Usage of Net::HTTP::Persistent in the Faraday::Adapter::NetHttpPersistent:
https://github.com/lostisland/faraday/blob/88de44ebe3c40f9050553ce7c0018990b92c2f38/lib/faraday/adapter/net_http_persistent.rb#L11

## Steps to reproduce
N/A

## Solution ?

I don't have a solution for this yet but I'm happy to submit a PR if someone can suggest the right way to approach this. 

An example usage is here:
https://github.com/lostisland/faraday#nethttppersistent
```ruby
conn = Faraday.new(...) do |f|
  f.adapter :net_http_persistent do |http| # yields Net::HTTP::Persistent
    http.idle_timeout = 100
    http.retry_change_requests = true
  end
end
```

The problem with this is that it yields an already initialized Net::HTTP::Persistent object so it's too late to provide the additional constructor param. 

Faraday::Adapter::NetHttpPersistent does have access to `env` when it's creating the instance of Net::HTTP::Persistent though. Is there an appropriate way to add a `pool_size` config to the `env`?

iMacTia:
Hi @col, thanks for raising this.
No need to stick this information into the `env`, we can use `connection_options`.
This instance variable is already supported by some of the adapters and can easily be added to `Net::HTTP::Persistent` too. You can see how it works by running a quick search: https://github.com/lostisland/faraday/search?q=connection_options&unscoped_q=connection_options

Usage would be something like:

```ruby
conn = Faraday.new(...) do |f|
  f.adapter :net_http_persistent, pool_size: 100
end
```

If you manage to understand how it works, feel free to open a PR otherwise I'll try to find some time over the weekend, it should be a 5 mins job to add it in

col:
Great. That’s exactly what I was looking for. I’ll get it done in the next
[...]

PR #834, opened 2018-11-23
Expose `pool_size` as a option for the NetHttpPersistent adapter

col:
Expose `pool_size` as a connection option for the NetHttpPersistent adapter

Fixes: https://github.com/lostisland/faraday/issues/833

## Description
- Note: connection pool was only added to `net-http-persistent` from version 3.0
- The `pool_size` constructor param in Net::HTTP::Persistent has a default value so I've been careful to ensure this will still use the default behaviour when the `pool_size` adapter option is not provided.

## Todos
- [x] Tests
- [x] Documentation

## Additional Notes
...

### DIFF
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -8,7 +8,9 @@ module Faraday
       def net_http_connection(env)
         @cached_connection ||=
           if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
-            Net::HTTP::Persistent.new(name: 'Faraday')
+            options = {}
+            options[:pool_size] = @connection_options[:pool_size] if @connection_options.key?(:pool_size)
+            Net::HTTP::Persistent.new(name: 'Faraday', **options)
           else
             Net::HTTP::Persistent.new('Faraday')
           end
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -8,9 +8,9 @@ module Faraday
       def net_http_connection(env)
         @cached_connection ||=
           if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
-            options = {}
+            options = {name: 'Faraday'}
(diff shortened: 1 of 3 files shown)


lib/faraday/adapter/net_http_persistent.rb
```diff
@@ -8,7 +8,9 @@ class NetHttpPersistent < NetHttp
       def net_http_connection(env)
         @cached_connection ||=
           if Net::HTTP::Persistent.instance_method(:initialize).parameters.first == [:key, :name]
-            Net::HTTP::Persistent.new(name: 'Faraday')
+            options = {}
+            options[:pool_size] = @connection_options[:pool_size] if @connection_options.key?(:pool_size)
+            Net::HTTP::Persistent.new(name: 'Faraday', **options)
```
iMacTia: The above can be rewritten as:

```ruby
options = {name: 'Faraday'}.merge(@connection_options.slice(:name, :pool_size)
Net::HTTP::Persistent.new(options)
```

This way we:
* Simplify the filtering using `slice`
* Allow to override the connection name

`proxy` would also be a valid key, but I'd not slice it in as Faraday manages the proxy separately.

col:
Yeah that looks good. Think about making name an option as well. Will
update shortly.

Want me to update the example usage in the readme to include the pool_size
option?

iMacTia:
Why not, sounds like a great idea 😃

lib/faraday/adapter/net_http_persistent.rb
col: I believe `slice` is only available in rails.

I've updated the the PR but only with a small improvement to fix the 1.9.3 issue. 
Out of time now but feel free to refactor if you have any other ideas.

iMacTia: I tried it in the console to be sure and it worked, but I just found out that's because it was also added to Ruby 2.5 🤦‍♂️.
You're right, it's not available in Ruby 2.4 or before unless using activesupport.
The new implementation doesn't allow to customise `name`, but was never part of the issue you opened so I guess we can leave that out for now 👍

PR #835, opened 2018-11-28
Inject connection class as a dependency

adsteel:
## Description
The `Faraday` module is a handy little builder for instances of `Faraday::Connection`, but it makes it hard to create a `Faraday::Connection` wrapper for custom behavior. By injecting a connection class the gem becomes more easily extended.

Case in point: In my example, I want to be able to log details of each request, in a specific format, to Datadog. If I could subclass `Faraday::Connection`, wrap the methods I care about with our custom code, and have the builder user our connection wrapper instead, we'd be pretty set.

## Todos
- [ ] Documentation

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -14,7 +14,7 @@ require 'forwardable'
 #   conn.get '/'
 #
 module Faraday
-  VERSION = "0.15.4"
+  VERSION = "0.16.0"
 
   class << self
     # Public: Gets or sets the root path that Faraday is being loaded from.
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -63,11 +63,11 @@ module Faraday
     #   Faraday.new :url => 'http://faraday.com',
     #     :params => {:page => 1}
     #
-    # Returns a Faraday::Connection.
+    # Returns an instance of the connection class (Faraday::Connection by default)
     def new(url = nil, options = nil)
       block = block_given? ? Proc.new : nil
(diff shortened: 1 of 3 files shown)


iMacTia:
Hi @adsteel and thanks for using Faraday!
This looks like a pretty standard case for a custom middleware. Have you already considered going down that route?

adsteel:
@iMacTia I have not, but I will, thanks! Closing while I explore existing options.

iMacTia:
Not sure if this will allow you to do everything you want to do, but my suggestion would be to subclass the [Logger](https://github.com/lostisland/faraday/blob/master/lib/faraday/response/logger.rb) middleware or to create a similar one and add it to your stack. That's much easier than creating a subclass of `Connection`. Plus middlewares are the favourite tool to extend Faraday functionalities :)
Feel free to open an issue if you need any help

adsteel:
@iMacTia I see in the docs how to write custom middleware and how to modify existing middleware, but how would I add the custom middleware that I wrote? Just set an arbitrary prop on the connection object?

```
conn = Faraday.new(:url => 'http://sushi.com') do |faraday|
  faraday.request  :url_encoded
  faraday.my_middleware MyMiddleware
  faraday.response :logger
  faraday.adapter  Faraday.default_adapter
end
```

I can open this as an issue, too.

iMacTia:
Almost!

That would be `faraday.use MyMiddleware` in the connection setup

adsteel:
Works a treat, thank you!

PR #839, opened 2018-12-28
Travis: add 2.6.0 to CI matrix, remove sudo: false

olleolleolle:
## Description

Updates **master branch** CI matrix, using aliases in rvm (`2.6` for latest in 2.6.x series).

Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: ruby
 script: bundle exec script/test
 cache: bundler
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
 language: ruby
 script: bundle exec script/test
 cache: bundler
```
iMacTia: I noticed all failing tests have some issues with bundler.
Maybe removing the cache for it will fix them up?

olleolleolle: It’s worth a try!

iMacTia:
Don't sweat too much on it @olleolleolle if you can't get it working.
v1.0 will replace this anyway so Ruby 2.6 support is not too much important in a legacy version.

olleolleolle:
More like 🐢 !

PR #843, opened 2019-01-10
Allow adapter to work with Net::HTTP::Persistent 3.0 breaking changes

booleanbetrayal:
## Description

See also v3 release notes: https://github.com/drbrain/net-http-persistent/blob/master/History.txt
> Net::HTTP::Persistent::new now uses keyword arguments for +name+ and +proxy+.


iMacTia:
Hi @booleanbetrayal this has already been implemented as you can see from the code, you just mixed the condition branches.
The condition checks if the initialiser expects keyword arguments (i.e. if you're using v3.0) and passes the `{ name: 'Faraday' }` options.
The `if` branch you modified was the one for v2.0

booleanbetrayal:
Thanks! Had been grep'ing and missed the fallthrough in my haste.  =]

PR #517, opened 2015-09-01
don't stomp on request body

lanej:
- store request_body in env
- allow response to contain all the information of the request

PR #847, opened 2019-02-19
Separate Request and Response bodies

iMacTia:
## Description
Separates request and response bodies in `Faraday::Env`.
Adds methods and hacks for backwards compatibility.
Fixes #517

### DIFF
--- a/spec/faraday/options/env_spec.rb
+++ b/spec/faraday/options/env_spec.rb
@@ -1,20 +1,22 @@
 RSpec.describe Faraday::Env do
+  subject(:env) { described_class.new }
+
   it 'allows to access members' do
-    expect(subject.method).to be_nil
-    subject.method = :get
-    expect(subject.method).to eq(:get)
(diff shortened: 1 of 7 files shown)


spec/faraday/options/env_spec.rb
```diff
@@ -34,4 +33,34 @@
     expect(env2[:bar]).to eq(:custom_2)
     expect(subject[:baz]).to be_nil
   end
+
+  describe '#body' do
+    subject { Faraday::Env.from(body: { foo: 'bar' }) }
```
olleolleolle: Spec Style note - which I just thought of now, but is not critical: When you're going to directly refer to the `subject`, name it.

Suggested standard name: `env` (Provided it's not taken by something else.) So, as definition:

```ruby
subject(:env) { ... }
```
 
That would make it possible to use it like:

```ruby
expect(env.body).to eq({ foo: 'bar' })
```

PR #857, opened 2019-02-20
implement #options.

technoweenie:
This implements `Faraday::Connection#options` _almost_ as expected, fixing https://github.com/lostisland/faraday/issues/305.

We can't rename [the current `:options attr_reader`](https://github.com/lostisland/faraday/blob/63918ee8f00ab92f2c2ff7a434eb67e6ae151b76/lib/faraday/connection.rb#L30-L31) right now for backwards compatibility reasons. Instead, this overloads the method to support both use cases. Merging the two risks some clashing, but I can only think of one issue: the other request helpers like `#get` can be called with no arguments:

```ruby
conn = Faraday.new 'http://example.com'
conn.get # GET /
conn.options # oops!
```

You can get around this easily with `conn.options(nil)`. I think it's worth introducing this regardless of the risk, assuming we'll rip it out when v1.0 ships.

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -163,12 +163,35 @@ module Faraday
     METHODS_WITH_QUERY.each do |method|
       class_eval <<-RUBY, __FILE__, __LINE__ + 1
         def #{method}(url = nil, params = nil, headers = nil)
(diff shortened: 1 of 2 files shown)


iMacTia:
Ignoring CodeClimate issues, not sure they actually make sense in this case

PR #861, opened 2019-02-21
Connection methods

technoweenie:
This adds generic tests for all of the supported request method convenience methods on `Faraday::Connection`, and adds support for [TRACE](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/TRACE) and [CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT). Both methods were put in the `METHODS_WITH_QUERY` const based on the linked MDN doc pages for each method. The MDN pages state that requests using either method will not have a request body. However, you can override this with Faraday's block syntax.

```ruby
res = conn.trace("/foo") do |req|
  req.body = "t r a c e"
end
```


technoweenie:
These shared tests uncovered a slight difference in behavior between `#options` and the other convenience methods that don't typically have a body (`#get`, `#head`, etc).

```
 1) making requests with #options behaves like method with query makes request with path
     Failure/Error: expect(res.env.request_body).to be_nil
     
       expected: nil
            got: ""
     Shared Example Group: "method with query" called from ./spec/faraday/connection_methods_spec.rb:70
     # ./spec/faraday/connection_methods_spec.rb:15:in `block (3 levels) in <top (required)>'
```

This is because `Faraday::Env::MethodsWithBodies` includes `:options` as of https://github.com/lostisland/faraday/commit/2343a4e98fdbce3f95e0968083f7bbe92c52b9f6. It turns out that line moved files and classes several times, before ending up on `Faraday::Env`. A new `Faraday::METHODS_WITH_BODY` const was added to define the convenience helpers without `options`. 

MDN says that [OPTIONS requests have no body](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS), so I'm comfortable with removing that from `Faraday::Env::MethodsWithBodies`.

technoweenie:
I'm going to merge the added tests to the existing shared examples in `spec/support/shared_examples/adapter.rb` so everything is together.

iMacTia:
The presence of multiple sources for methods with bodies might be due to the parallel work on 0.x and 1.0 branches. We should indeed consolidate them and have a single source of truth (my suggestion would be on the main Faraday module, as they will be available everywhere this way.

technoweenie:
@iMacTia I had to tweak my change because I discovered the constants are used for _slightly_ different purposes.

* `Faraday::METHODS_WITH_QUERY` and `Faraday::METHODS_WITH_BODY` determine which convenience helpers are created on `Faraday::Connection`.
* `Faraday::Env::MethodsWithBodies` determines which HTTP verbs expect to have a request body.

They're different, but both `Faraday::Env::MethodsWithBodies` and `Faraday::METHODS_WITH_BODY` should be the same (except one is a String Array, and the other is a Set of Symbols). However, If there was a `Faraday::Env::MethodsWithQueries` constant, it would have to include `:options`. But, we specifically do not want `"options"` in `Faraday::METHODS_WITH_QUERY`, because its implementation is custom.

It's a difference worth noting, but for now I will link `Faraday::Env::MethodsWithBodies` to `Faraday::METHODS_WITH_BODY`.

spec/faraday/adapter/excon_spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe Faraday::Adapter::Excon do
-  features :body_on_get, :reason_phrase_parse
+  features :body_on_get, :reason_phrase_parse, :trace_method, :connect_method
```
iMacTia: Is this because not all adapters support these verbs?

spec/support/shared_examples/request_method.rb
```diff
@@ -44,10 +44,29 @@
   #   end
   # end
 
+  on_feature :body_on_get do
```
iMacTia: nit: This now is tested for all verbs, should we rename the feature to something more generic?

spec/faraday/adapter/excon_spec.rb
technoweenie: Yup, Patron :)

```
  51) Faraday::Adapter::Patron behaves like an adapter with SSL disabled #connect behaves like a request method handles headers with multiple values
      Got 1 failure and 1 other error:
      Shared Example Group: "a request method" called from ./spec/support/shared_examples/adapter.rb:56
      Shared Example Group: "adapter examples" called from ./spec/support/shared_examples/adapter.rb:11
      Shared Example Group: "an adapter" called from ./spec/faraday/adapter/patron_spec.rb:4

      51.1) Failure/Error: session.request(env[:method], env[:url].to_s, env[:request_headers], :data => data)

            ArgumentError:
              Action must be one of GET, PUT, POST, DELETE, HEAD, COPY, PATCH, OPTIONS
            # ./lib/faraday/adapter/patron.rb:30:in `call'
            # ./lib/faraday/response.rb:8:in `call'
            # ./lib/faraday/request/url_encoded.rb:20:in `call'
            # ./lib/faraday/request/multipart.rb:19:in `call'
            # ./lib/faraday/rack_builder.rb:151:in `build_response'
            # ./lib/faraday/connection.rb:485:in `run_request'
            # ./lib/faraday/connection.rb:196:in `connect'
            # ./spec/support/shared_examples/request_method.rb:3:in `public_send'
            # ./spec/support/shared_examples/request_method.rb:3:in `block (2 levels) in <top (required)>'
            # ./spec/support/shared_examples/request_method.rb:13:in `block (2 levels) in <top (required)>'
```

It may be easy to fix upstream, but then Faraday would have to depend on a more specific version of Patron.

spec/support/shared_examples/request_method.rb
technoweenie: How about `:request_body_on_query_methods`?

iMacTia: Sounds good!

spec/faraday/adapter/excon_spec.rb
iMacTia: Ah, shame on Patron! Fair enough, thanks for confirming :)

iMacTia:
I've changed the only outstanding comment so it's good to go now 👍

technoweenie:
So clearly I've never implemented a Proxy or Websocket server (or fully read the MDN links above), but now I think that the `CONNECT` method should be removed from Faraday. I added it because I initially thought it was some other obscure method like `TRACE` or `OPTIONS`. A Faraday user isn't going to be implementing a proxy (each adapter has its own proxy support) or websocket (out of scope) client.

PR #868, opened 2019-02-24
Support Frozen String Literals

iMacTia:
## Description
Support Frozen String Literals


olleolleolle:
@iMacTia There were issues re: CodeClimate still.

PR #869, opened 2019-02-24
chore: RuboCop: Enable with_fixed_indentation cop for Layout/AlignParameters

olleolleolle:
## Description

This is about `Layout/AlignParameters` - a RuboCop rule.

This PR configures that rule, to try to keep most code as-is.

## Alternative routes

I was reviewing #867 - which used the default setting in that Cop from RuboCop and applied auto-correct.  I went the other way – decided to add a setting to that cop – and then ran auto-correct.

That PR follows RuboCop default recommendations, this PR follows (my preconceptions of) the existing code-base.

The amount of code changes are _about the same_.

**Problem with this PR**: It has to change more files than the other PR.


olleolleolle:
Closing - I think the other PR is better.

Issue #854, opened 2019-02-20
RuboCop Quest Issue


olleolleolle:
## Basic Info

We are linting the whole codebase using RuboCop.

This is a list of all the rules in the `.rubocop_todo.yml` when we started. The checkmarks mean "this is fixed and merged to master".

If you create a PR to fix one of these, use the `--only` option in RuboCop, to focus the effort on 1 kind of fix. It's so easy to get overwhelmed when reviewing code like this. (If you fix slightly more than that, no biggie. This is all about keeping the changes reviewable.)

<details>

```bash
# edit .rubocop.yml, commenting out the first line -
#   this removes the "ignore the TODOs" setting

# Auto-correct with only 1 rule
rubocop -a --only Name/OfTheRuleYouAreFixing --auto-correct

# Re-generate the config file
rubocop --auto-gen-config

# Revert line 3 in .rubocop_todo.yml
# This will avoid PR conflicts

# Inspect to see if it looks OK
git diff

# Re-set the TODO list
git checkout .rubocop.yml

# Add and commit the change
git add .
git commit -m"chore: RuboCop lint Name/OfTheRuleYouAreFixing"
```
<summary>A workflow</summary>

</details>

## Issue description

 - [x] Bundler/OrderedGems #856
 - [x] Layout/AccessModifierIndentation #862 
 - [x] Layout/AlignHash #863 
 - [x] Layout/AlignParameters #867 
 - [x] Layout/BlockEndNewline
 - [x] Layout/CaseIndentation #864 
 - [x] Layout/CommentIndentation #864
 - [x] Layout/DotPosition #880
 - [x] Layout/ElseAlignment #871
 - [x] Layout/EmptyLineAfterGuardClause #860
 - [x] Layout/EmptyLineBetweenDefs #860
 - [x] Layout/EmptyLines #860
 - [x] Layout/EmptyLinesAroundAccessModifier #881 
 - [x] Layout/EmptyLinesAroundClassBody #882
 - [x] Layout/EndAlignment #887 
 - [x] Layout/ExtraSpacing #884
 - [x] Layout/IndentArray #890
 - [x] Layout/IndentHash #890
 - [x] Layout/IndentationWidth #864
 - [x] Layout/LeadingCommentSpace #860
 - [x] Layout/MultilineBlockLayout
 - [x] Layout/MultilineMethodCallIndentation #864
 - [x] Layout/MultilineOperationIndentation #864
 - [x] Layout/RescueEnsureAlignment #890
[...]

PR #870, opened 2019-02-25
Fix rubocop issues from #861

technoweenie:
Goal is to fix the rubocop failures in master that came from #861:

```
spec/support/shared_examples/request_method.rb:53:9: C: Layout/IndentationWidth: Use 2 (not -4) spaces for indentation. (https://github.com/rubocop-hq/ruby-style-guide#spaces-indentation)
        conn.public_send(http_method, '/', 'test')
        ^^^^
spec/support/shared_examples/request_method.rb:54:7: C: Layout/ElseAlignment: Align else with if.
      else
      ^^^^
spec/support/shared_examples/request_method.rb:58:7: W: Layout/EndAlignment: end at 58, 6 is not aligned with if at 52, 12.
      end
      ^^^
spec/support/shared_examples/adapter.rb:82:3: C: Layout/IndentationWidth: Use 2 (not 1) spaces for indentation. (https://github.com/rubocop-hq/ruby-style-guide#spaces-indentation)
   let(:http_method) { :options }
  ^
spec/support/shared_examples/adapter.rb:84:5: C: Layout/IndentationConsistency: Inconsistent indentation detected. (https://github.com/rubocop-hq/ruby-style-guide#spaces-indentation)
    it_behaves_like 'a request method', :options
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/support/shared_examples/adapter.rb:107:5: C: Layout/IndentationWidth: Use 2 (not 1) spaces for indentation. (https://github.com/rubocop-hq/ruby-style-guide#spaces-indentation)
     let(:http_method) { :trace }
    ^
spec/support/shared_examples/adapter.rb:109:7: C: Layout/IndentationConsistency: Inconsistent indentation detected. (https://github.com/rubocop-hq/ruby-style-guide#spaces-indentation)
      it_behaves_like 'a request method', :trace
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #871, opened 2019-02-25
Resolve Layout/ElseAlignment Violations

jherdman:
## Description

Tick another box on #854

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -22,14 +22,6 @@ Layout/DotPosition:
     - 'lib/faraday/autoload.rb'
     - 'lib/faraday/connection.rb'
 
-# Offense count: 3
-# Cop supports --auto-correct.
-Layout/ElseAlignment:
-  Exclude:
-    - 'lib/faraday/adapter/rack.rb'
-    - 'lib/faraday/options.rb'
-    - 'lib/faraday/utils.rb'
-
 # Offense count: 1
 # Cop supports --auto-correct.
 Layout/EmptyLinesAroundAccessModifier:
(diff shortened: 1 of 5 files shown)


technoweenie:
Hi @jherdman, thanks for taking a look at this! This PR doesn't actually fix all of the `Layout/ElseAlignment` offenses though, just the ones not excused by the [rubocop todo file](https://github.com/lostisland/faraday/blob/3a11286c8746e4e086729e60cc3cc1893060db8d/.rubocop_todo.yml#L27-L31).

This specific change fails on master because it looks like our CI process is not preventing merges properly. This PR fixes half of it, and #870 fixes the rest. I'd like to get our CI issues resolved before merging these though.

jherdman:
Ah, snap. I didn't realize that file was there. I'll take a look at those too.

technoweenie:
@jherdman I'm new to rubocop, so I'm trying to come up with a general workflow. I think it's:

1. Remove exceptions from `.rubocop_todo.yml`
2. Fix code
3. Run `bundle exec rubocop --auto-gen-config`
4. Commit, PR, Merge, Profit :metal:

Do you know if there's some way to run `rubocop --only=Layout/BlockEndNewline` while ignoring the todo file. I'm just trying to document the rubocop fix workflow.

jherdman:
> Do you know if there's some way to run `rubocop --only=Layout/BlockEndNewline` while ignoring the todo file. I'm just trying to document the rubocop fix workflow

Sadly, no. I'm in the same boat as you w.r.t RuboCop.

I went back, though, and resolved the remaining ElseAlignment problems.

PR #873, opened 2019-02-25
CI: RuboCop as a Build Stage before tests run

olleolleolle:
## Description

~This PR frontloads RuboCop checks onto the default Rake task.~

This PR adds a TravisCI Build Stage `linting`, which installs `rubocop` and runs it.

If that stage fails, no virtual machines with the RSpec tests are even booted.

We use this way, in order to keep up with latest RuboCop releases, since CodeClimate's built-in is a good few versions behind.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,5 +1,4 @@
 version: "2"
 plugins:
   rubocop:
-    enabled: true
-
+    enabled: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,24 @@
 language: ruby
 cache: bundler
+
 rvm:
   - 2.3
   - 2.4
   - 2.5
   - 2.6
   - ruby-head
+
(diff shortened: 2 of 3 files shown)


olleolleolle:
See https://github.com/lostisland/faraday/pull/870 for comparison.

olleolleolle:
Full disclosure: CodeClimate runs "channels" of versions of RuboCop:

https://github.com/codeclimate/codeclimate-rubocop/branches/all?utf8=%E2%9C%93&query=channel%2Frubocop

This allows one to pick a version family, using the `channel` setting in the CodeClimate configuration.

olleolleolle:
The furious pace of [RuboCop releases](https://github.com/rubocop-hq/rubocop/releases) is hard to keep up with – also for CodeClimate. 0.64 is the latest of their Channels. 0.65.0 is the latest RuboCop release.

PR #878, opened 2019-02-26
chore: Fix RuboCop Style/HashSyntax

iMacTia:
## Description

See #854 - RuboCop linting

Fixes: 

- Style/HashSyntax


lib/faraday/adapter/em_http.rb
```diff
@@ -19,8 +19,8 @@ def connection_config(env)
 
         def request_config(env)
           options = {
-            :body => read_body(env),
-            :head => env[:request_headers]
+            body: read_body(env),
+            head: env[:request_headers]
             # :keepalive => true,
             # :file => 'path/to/file', # stream data off disk
```
olleolleolle: You may wish to change these in the comment, too.

iMacTia: Wow congrats for noticing in between all the changes!!

PR #889, opened 2019-02-27
Add the log formatter that is easy to override and safe to inherit

### LINKED ISSUE #888: Suggestion to improve logging support
## Basic Info

I am concerned about the usability of request/response logging. At the moment one should override the logging middleware like that:

```
Faraday::Response.register_middleware logger: -> { FaradayMiddlewares::MyLogger }
```

And then you have to deal with the inheritance. For example you want to change the request logging - then you have to skip using `super` and bring code from `Faraday::Response::Middleware#call` to it(To avoid duplicate logging for example). As of `super` would call a method from `Faraday::Response::Logger` and then it would propagate to `Faraday::Response::Middleware`.

Dealing with logging requests/response in `faraday` is non-transparent(for me at least). I have a suggestion(backed up by some code: https://github.com/lostisland/faraday/pull/889) of separating log formatter from the logging middleware.

E.g. I would be happy to provide my formatter to the connection configuration:

```rb
conn = Faraday.new(:url => 'http://sushi.com/api_key=s3cr3t') do |faraday|
  faraday.response :logger, logger,   formatter: MyFormatter
end
```

That could be inherited from the base formatter with something like 2 methods to override: `request` and `response`.

## Steps to reproduce
To see the issue I would recommend an exercise: integrate a structured logger(like `SemanticLogger`) or at least convert a multiline logging to a one-line-per-request(supporting bodies logging)

prikha:
## Description
Extract Formatter from logging middleware and enable it to be configurable.

Fixes #888

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -1,82 +1,29 @@
 # frozen_string_literal: true
 
 require 'forwardable'
+require 'faraday/logging/formatter'
 
 module Faraday
   class Response
     class Logger < Middleware
-      extend Forwardable
-
-      DEFAULT_OPTIONS = { headers: true, bodies: false }.freeze
-
       def initialize(app, logger = nil, options = {})
         super(app)
-        @logger = logger || begin
+        logger ||= begin
           require 'logger'
           ::Logger.new($stdout)
         end
-        @filter = []
-        @options = DEFAULT_OPTIONS.merge(options)
-        yield self if block_given?
+        formatter_class = options.delete(:formatter) || Logging::Formatter
+        @formatter = formatter_class.new(logger: logger, options: options)
+        yield @formatter if block_given?
       end
 
-      def_delegators :@logger, :debug, :info, :warn, :error, :fatal
-
       def call(env)
-        info('request')  { "#{env.method.upcase} #{apply_filters(env.url.to_s)}" }
-        debug('request') { apply_filters(dump_headers(env.request_headers)) } if log_headers?(:request)
-        debug('request') { apply_filters(dump_body(env[:body])) } if env[:body] && log_body?(:request)
+        @formatter.request(env)
         super
       end
 
       def on_complete(env)
-        info('response')  { "Status #{env.status}" }
-        debug('response') { apply_filters(dump_headers(env.response_headers)) } if log_headers?(:response)
-        debug('response') { apply_filters(dump_body(env[:body])) } if env[:body] && log_body?(:response)
-      end
-
-      def filter(filter_word, filter_replacement)
-        @filter.push([filter_word, filter_replacement])
-      end
-
-      private
-
-      def dump_headers(headers)
-        headers.map { |k, v| "#{k}: #{v.inspect}" }.join("\n")
-      end
-
-      def dump_body(body)
-        if body.respond_to?(:to_str)
-          body.to_str
-        else
-          pretty_inspect(body)
-        end
-      end
-
-      def pretty_inspect(body)
-        require 'pp' unless body.respond_to?(:pretty_inspect)
-        body.pretty_inspect
-      end
-
-      def log_headers?(type)
-        case @options[:headers]
-        when Hash then @options[:headers][type]
-        else @options[:headers]
-        end
-      end
-
-      def log_body?(type)
-        case @options[:bodies]
-        when Hash then @options[:bodies][type]
-        else @options[:bodies]
-        end
-      end
-
-      def apply_filters(output)
-        @filter.each do |pattern, replacement|
-          output = output.to_s.gsub(pattern, replacement)
-        end
-        output
+        @formatter.response(env)
       end
     end
   end
--- a/spec/faraday/response/logger_spec.rb
+++ b/spec/faraday/response/logger_spec.rb
@@ -38,6 +38,61 @@ RSpec.describe Faraday::Response::Logger do
     expect(resp.body).to eq('hello')
   end
 
+  context 'without configuration' do
+    let(:conn) do
+      Faraday.new do |b|
+        b.response :logger
+        b.adapter :test do |stubs|
+          stubs.get('/hello') { [200, { 'Content-Type' => 'text/html' }, 'hello'] }
+        end
+      end
+    end
+
+    it 'defaults to stdout' do
+      expect(Logger).to receive(:new).with($stdout).and_return(Logger.new(nil))
+      conn.get('/hello')
+    end
+  end
+
+  context 'with default formatter' do
+    let(:formatter) { instance_double(Faraday::Logging::Formatter, request: true, response: true, filter: []) }
+
+    before { allow(Faraday::Logging::Formatter).to receive(:new).and_return(formatter) }
+
+    it 'delegates logging to the formatter' do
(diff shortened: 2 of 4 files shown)


technoweenie:
@prikha: Thank you for making your suggestion AND filing a PR with an implementation. I agree with the pain points you stated in #888, however I'm not sold on the implementation. It leaves the Logger middleware class so bare that I wonder why it's even there.

```ruby
class Faraday::Response::Logger < Response::Middleware 
  def call(env)
    @formatter.request(env)
    super
  end

  def on_complete(env)
    @formatter.response(env)
  end
end
```

Did you consider implementing `Formatter#request` and `Formatter#response` in the `Logger` class itself? I think you'd get most of the benefits, without having to deal with a `Logger` and a `Formatter` class.

I suppose passing a custom Logger would require something like this (as you suggested in #888):

```ruby
Faraday::Response.register_middleware logger: -> { FaradayMiddlewares::MyLogger }

conn = Faraday.new(:url => 'http://sushi.com/api_key=s3cr3t') do |faraday|
  faraday.response :logger, logger
end
```

That would at least let you replace all of the loggers in an application at once, but could obscure the fact that you're using a custom logger if you're just looking at a single instance of Faraday usage.

Is there some big reason that you feel the Formatter should be extracted out to a separate class that I'm missing? I think refactoring the `Logger` class itself would give you the same benefits. The only difference is _how_ you specify your custom logger.

prikha:
Hey @technoweenie thanks for taking a look. Strictly speaking there is nothing a decent programmer can't overcome(well at least in Ruby 😆). One can invent custom logging middleware and then do whatever I want there. That is true. However I would like to safely reuse some of the code from the `faraday` since it is already there and is documented to work in a certain way.

I see no issue in having a logging middleware just for the purpose of exposing the logging capabilities. 

With the approach I suggest one can safely inherit from the current formatter. However inheriting from logging middleware implies that every change to parents `call` method would break it.

On the other hand custom middleware accepts a logger which is expected to have more or less trivial behaviour, like expose some methods having some signatures. Well ruby isn't a typed language, but when I see `logger` I set myself some expectations.

Maybe it would make sense to have `request` and `response` log formatters? That would allow code reuse and clearly separate the responsibility.

In general it would be cool to have logging extracted out of `Request` namespace which is way misleading. But it is way out of current discussion. 

To sum up, I see #888 as an opportunity to improve the integration points of `faraday`, maybe we can thing of a better implementation, but this one solves `faraday` and `SemanticLogger`. Which would be a bit more painful otherwise.

iMacTia:
To add on top of what @technoweenie already said, I'm personally happy with both solutions (a more modular logger contract, or a separate formatter). However I'd like to point out that the current implementation doesn't allow to reuse some useful methods even if you sub-class `Formatter`: I'm talking about methods like `log_headers?`, `log_bodies?` and `apply_filters`, which are currently `private`.
Whatever solution we go for, I'd like these helpers to be available on sub-classes so that creating my custom logger/formatter will be easier

prikha:
@iMacTia thanks for participating. Maybe I'm not the best seller(I appreciate some help here), I am happy to tune the code to the point where we get a clean point of integration, however. 

Can you clarify about the `private` methods? You mean it might be useful for them to be acceptable in the middleware itself? We could delegate those to the formatter from the logging middleware. Because when I subclass from the `Formatter` I surely get them available inside of my own implementation.

iMacTia:
@prikha My fault, I was sure you couldn't access `private` method in subclasses 🙈. Please ignore me

iMacTia:
@technoweenie Although both solutions are perfectly fine, I personally like the separation of concerns introduced with this PR as it clearly identifies the different actors in the logging process:

* **Logger Middleware**: Implements the middleware contract (`call`) and delegates the implementation to the formatter.
* **Formatter**: Takes request or response and format them to be printed.
* **Logger**: Takes the formatted request/response and logs it.

It's quite similar to how `lograge` works and allows to easily change one piece to change the behaviour.
Do you need to log as JSON? Change the logger.
Do you need to log only one line per request/response? Change formatter.

technoweenie:
Great, lets go with this PR's approach :metal: 

My only concern now is whether it breaks any existing APIs. Just looking at [the current Logger class](https://github.com/lostisland/faraday/blob/4c7e0a7717bf20bebbdb5cec70680432bee3477c/lib/faraday/response/logger.rb), we may have to keep `#filter` and the other private methods around in case anyone is using them. We're building up to Faraday v1.0 with compatibility with the existing community being a major consideration.

lib/faraday/response/logger.rb
```diff
@@ -4,78 +4,93 @@
...
+      def dump_body(body)
+        if body.respond_to?(:to_str)
+          body.to_str
+        else
+          pretty_inspect(body)
+        end
+      end
+
+      def pretty_inspect(body)
+        require 'pp' unless body.respond_to?(:pretty_inspect)
```
technoweenie: Any thoughts on moving this? This could be required at the top instead. The Faraday Middleware Registry should prevent this from being loaded _until_ someone uses the logger middleware for the first time.

iMacTia: makes sense to me

lib/faraday/response/logger.rb
```diff
@@ -4,78 +4,93 @@
...
-      info('response')  { "Status #{env.status}" }
-      debug('response') { apply_filters(dump_headers(env.response_headers)) } if log_headers?(:response)
-      debug('response') { apply_filters(dump_body(env[:body])) } if env[:body] && log_body?(:response)
+      @formatter.response(env)
     end
 
-    def filter(filter_word, filter_replacement)
-      @filter.push([filter_word, filter_replacement])
-    end
+    class Formatter
```
iMacTia: Although this class is directly linked to this logger, we're trying to isolate all classes into separate files do make the codebase a little cleaner.
@technoweenie what do you think of moving the formatter on a separate file?

prikha: Lets move it to `faraday/logging/formatter.rb` this was it wouldn't have a misleading `Request` namespace. I'll add a commit for that.

prikha:
@technoweenie comments addressed, when we agree on the changes - I'll rebase and fix conflicts.

spec/faraday/response/logger_spec.rb
```diff
@@ -38,6 +38,65 @@
...
+
+        def response(_env)
+          info 'Custom log formatter response'
+        end
+      end
+    end
+
+    after do
+      Object.send(:remove_const, :CustomFormatter)
+    end
```
technoweenie: Setting the class to a local variable should keep it from modifying the global ruby module namespace. What do you think about something like this?

```ruby
  let :formatter_class do
    Class.new(Faraday::Logging::Formatter) do
      # impl
    end
  end
```

technoweenie:
@prikha Thanks for addressing the comments. I commented on a nit pick in the tests, but I won't let that block the merging of this PR. We should be able to merge this once you fix the conflict.

prikha:
@iMacTia @technoweenie Thanks for the tip. Addressed. Give it a quick look and it's good to go.

olleolleolle:
🌞 😍 🌞 Thank you so much for taking this all the way!

PR #890, opened 2019-02-27
Resolve Remaining Layout Cops

jherdman:
Going one at a time was taking a bit too long. Each commit resolves an individual layout cop for #854


jherdman:
@olleolleolle is this approach okay? If so I might resolve another group of cops.

olleolleolle:
Ah, ha - I got carried away!

olleolleolle:
(Now updated the worklist. Thanks again!)

PR #896, opened 2019-02-27
fix Lint/* offenses

technoweenie:
This fixes some more rubocop offsenses:

* Lint/Void
* Lint/UnusedMethodArgument
* Lint/UnusedBlockArgument
* Lint/UnneededRequireStatement
* Lint/ReturnInVoidContext
* Lint/HandleExceptions


lib/faraday/connection.rb
```diff
@@ -416,8 +414,6 @@ def url_prefix=(url, encoder = nil)
         basic_auth user, password
         uri.user = uri.password = nil
       end
-
-      uri
```
technoweenie: `Lint/Void` looks for property setters that explicitly return some value. I really hope someone isn't doing something like `foo = conn.url_prefix = "bar"`, because this would break that. And, it's probably not doing what the user thinks it's doing... :)

lib/faraday/adapter.rb
```diff
@@ -35,7 +35,7 @@ def inherited(subclass)
     extend Parallelism
     self.supports_parallel = false
 
-    def initialize(app = nil, opts = {}, &block)
```
technoweenie: This is weird. Why is that argument unused???

script/generate_certs
```diff
@@ -11,7 +11,7 @@ $shell = ARGV.include? '-s'
 
 # Adapted from WEBrick::Utils. Skips cert extensions so it
 # can be used as a CA bundle
-def create_self_signed_cert(bits, cn, comment)
+def create_self_signed_cert(bits, cn)
```
olleolleolle: Change needed: Get rid of the `comment` passed in to this method in the usages (in this same file). I think 2 usages, maybe.

lib/faraday/adapter.rb
iMacTia: Adapters have now become endpoints for the stack, they are always put at the end of it.
So `app` becomes useless. However, to ensure backwards-compatibility, we need to accept it

olleolleolle: This is perhaps worth a documentation comment (YARD format comment for the parameters).

PR #899, opened 2019-02-28
Improves `.travis.yml`

iMacTia:
## Description
* Test Coverage Report is sent to CodeClimate
* Gem release happens on a separate stage

I'm not a Travis CI expert so appreciate your feedback on these changes 😄.
The idea was that I wanted the code climate script to be downloaded and executed only one, not for all `rvm`s.
Also, deployment is currently limited to a singular version thanks to the `on` option, but moving that into a stage allows to use the stage `rvm` option to do the same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,6 @@ rvm:
   - 2.6
   - ruby-head
 
-before_script:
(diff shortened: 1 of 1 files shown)


iMacTia:
Still need quite some work! Closing and reopening as a draft

PR #913, opened 2019-03-03
chore: RuboCop lint Style/MultilineTernaryOperator

olleolleolle:
## Description

This PR lints a file with complex code.

The rule `Style/MultilineTernaryOperator` is now all fixed.

## Details

I fixed the rule. Then, with that fix, there were follow-on issues. I took them on, too.

I was much helped by the Workflow described in the #854 - thanks @iMacTia for adding that.


.rubocop_todo.yml
```diff
@@ -6,6 +6,12 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 1
+# Cop supports --auto-correct.
+Lint/UnneededCopDisableDirective:
```
olleolleolle: Hah, will regen!

.rubocop_todo.yml
```diff
@@ -25,7 +31,7 @@ Metrics/ClassLength:
 Metrics/CyclomaticComplexity:
   Max: 20
 
-# Offense count: 44
+# Offense count: 45
```
olleolleolle: This will be harder to reduce right now.

olleolleolle:
@iMacTia In order to verify that I'd got rid of the new sorts of issues, I regenerated the TODO.

PR #914, opened 2019-03-03
chore: Fix RuboCop Style/FormatStringToken

htwroclau:
See #854 - this PR fixes Style/FormatStringToken


.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2019-03-03 10:30:09 +0100 using RuboCop version 0.65.0.
+# on 2019-03-03 19:25:18 +0900 using RuboCop version 0.65.0.
```
olleolleolle: Ah, one of the new tricks is to revert JUST this.

```suggestion
# on 2019-03-03 10:30:09 +0100 using RuboCop version 0.65.0.
```

lib/faraday/encoders/nested_params_encoder.rb
```diff
@@ -99,7 +99,8 @@ def self.decode(query)
           if !last_subkey || is_array
             value_type = is_array ? Array : Hash
             if context[subkey] && !context[subkey].is_a?(value_type)
-              raise TypeError, format("expected %s (got %s) for param `%s'", value_type.name, context[subkey].class.name, subkey)
+              raise TypeError, format("expected %<value_type_name>s (got %<class_name>s) for param `%<subkey>s'",
+                                      value_type_name: value_type.name, class_name: context[subkey].class.name, subkey: subkey)
```
olleolleolle: If this would go on more lines, each key-value pair in the hash on its own line, we could avoid adding new Metrics/LineLength issues.

.rubocop_todo.yml
olleolleolle: See the #854 under Workflow.

lib/faraday/encoders/nested_params_encoder.rb
htwroclau: Thank you for your comments!
I try do it.

.rubocop_todo.yml
htwroclau: Thank you for your advise!

Does that mean that the time(19:25:18 +0900) has reverted?

PR #917, opened 2019-03-03
Set the max of Metrics/LineLength

htwroclau:
I think that it is too small to default value(80) of 'Style/LineLength'.

How about setting it between 100 and 120? (I set it to 120 in this PR)
Otherwise, for example `Style/IfUnlessModifier` modification is tough.

See #854


htwroclau:
Sorry, 
It was a halfway fix, so close this PR.

PR #921, opened 2019-03-03
chore: RuboCop lint Style/N*

iMacTia:
## Description

See #854 - RuboCop linting

Fixes: 

* Style/NumericPredicate
* Style/NilComparison

### DIFF
--- a/lib/faraday/encoders/flat_params_encoder.rb
+++ b/lib/faraday/encoders/flat_params_encoder.rb
@@ -46,8 +46,8 @@ module Faraday
     end
 
     def self.decode(query)
-      empty_accumulator = {}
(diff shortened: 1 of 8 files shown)


lib/faraday/encoders/flat_params_encoder.rb
```diff
@@ -47,7 +47,7 @@ def self.encode(params)
 
     def self.decode(query)
       empty_accumulator = {}
-      return nil if query == nil
+      return nil if query.nil?
```
olleolleolle: Could this be the first line of this method?

iMacTia: yes, agree

PR #923, opened 2019-03-04
Chore: RuboCop Style/MultipleComparison

Drenmi:
## Description

Part of [RuboCop Quest](https://github.com/lostisland/faraday/issues/854).

This fixes the single instance of the offense for `MultipleComparison`. It turned out that the code wasn't needed at all, since `Faraday::Utils.escape` already correctly stringifies booleans.

## Additional Notes

I also added a unit test that seemed to be missing for this code.


lib/faraday/encoders/flat_params_encoder.rb
```diff
@@ -29,7 +29,6 @@ def self.encode(params)
       buffer = +''
       params.each do |key, value|
         encoded_key = escape(key)
-        value = value.to_s if value == true || value == false
```
Drenmi: This is handled by `Faraday::Utils.escape` already:

```ruby
Faraday::Utils.escape(true)
#=> "true"
```

olleolleolle: I got curious about the test for this, where a `value` is a Boolean:

`.encode([{ 'a' => true, 'b' => false}])`

Perhaps we want that test, too?

iMacTia: Agree with @olleolleolle, we need a test that tests `encode` method with boolean values, if there isn't one already.

Drenmi: I might have been a bit too quick. Will have a look at it tonight. 🙂

iMacTia: After adding some more tests, it looks like this line really is redundant.

PR #924, opened 2019-03-04
Chore: RuboCop Style/MultilineIfModifier

Drenmi:
## Description

Part of [RuboCop Quest](https://github.com/lostisland/faraday/issues/854).

This change fixes all instances of `MultilineIfModifier`.

## Additional Notes

This introduces two more instances of `MethodLength` and `LineLength`, but since those are not addressed yet, I chose to focus on the linter at hand.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -44,10 +44,6 @@ module Faraday
     extend LiveServerConfig
     self.live_server = ENV['LIVE']
 
-    def test_default
-      assert true
-    end unless defined? ::MiniTest
-
     def capture_warnings
       old = $stderr
       $stderr = StringIO.new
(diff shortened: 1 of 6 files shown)


test/helper.rb
```diff
@@ -44,10 +44,6 @@ class TestCase < MiniTest::Test
     extend LiveServerConfig
     self.live_server = ENV['LIVE']
 
-    def test_default
```
Drenmi: I couldn't find a single reference to this, and the trailing modifier can not be used to prevent method definition, so it was actually not doing anything.

olleolleolle:
@Drenmi Hello, and thanks for joining this Quest! Shoutout to [Stockholm Ruby](http://www.stockholmruby.com/)

PR #928, opened 2019-03-04
chore: RuboCop lint Style/MethodMissingSuper

Insti:
## Description
Fix for Rubocop lint Style/MethodMissingSuper
This is part of the RuboCop Quest: #854 

## Additional Notes
Adds unit test coverage for the `Faraday#method_missing` method.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -83,11 +83,6 @@ Style/GuardClause:
     - 'lib/faraday/utils/headers.rb'
 
 # Offense count: 1
-Style/MethodMissingSuper:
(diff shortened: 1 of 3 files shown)


Insti:
The assignment to the mock connection 

`Faraday.default_connection = mock_conection`

is persisting throughout the tests.

If the random number generator is unlucky the following error occurs: 
```
       #<Double "Connection"> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
```

I'm working on a fix for this.

PR #932, opened 2019-03-05
Adds em-http adapter to RSpec tests

iMacTia:
## Description
Covers `em-http` with Rspec tests like other adapters.

## Additional Notes
Excluded `em_http_ssl_patch.rb` from the Rspec coverage as there's no way to unit-test that patch.
As discussed in pvt with other maintainers, these kind of things will be tested separately by an integration tests suite.

### DIFF
--- a/lib/faraday/adapter/em_http.rb
+++ b/lib/faraday/adapter/em_http.rb
@@ -166,17 +166,17 @@ module Faraday
       end
 
       def raise_error(msg)
-        errklass = Faraday::ClientError
-        if msg == Errno::ETIMEDOUT
-          errklass = Faraday::TimeoutError
+        error_class = Faraday::ClientError
+        if msg == Errno::ETIMEDOUT || (msg.is_a?(String) && msg.include?('timeout error'))
+          error_class = Faraday::TimeoutError
           msg = 'request timed out'
         elsif msg == Errno::ECONNREFUSED
-          errklass = Faraday::ConnectionFailed
+          error_class = Faraday::ConnectionFailed
           msg = 'connection refused'
         elsif msg == 'connection closed by server'
-          errklass = Faraday::ConnectionFailed
+          error_class = Faraday::ConnectionFailed
         end
-        raise errklass, msg
+        raise error_class, msg
       end
 
       # @return [Boolean]
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -25,6 +25,7 @@ SimpleCov.formatters = [SimpleCov::Formatter::HTMLFormatter, Coveralls::SimpleCo
 
 SimpleCov.start do
   add_filter '/spec/'
+  add_filter '/lib/faraday/adapter/em_http_ssl_patch'
   minimum_coverage 90
   minimum_coverage_by_file 70
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -31,7 +31,7 @@ end
 
 # Ensure all /lib files are loaded
 # so they will be included in the test coverage report.
-Dir['./lib/**/*.rb'].each { |file| require file }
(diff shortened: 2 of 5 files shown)


lib/faraday/adapter/em_http.rb
```diff
@@ -166,17 +166,17 @@ def error_message(client)
       end
 
       def raise_error(msg)
-        errklass = Faraday::ClientError
-        if msg == Errno::ETIMEDOUT
-          errklass = Faraday::TimeoutError
+        error_class = Faraday::ClientError
+        if msg == Errno::ETIMEDOUT || (msg.is_a?(String) && msg.include?('timeout error'))
```
technoweenie: Why was that extra `'timeout error'` check removed?

spec/support/helper_methods.rb
```diff
@@ -54,11 +58,7 @@ def with_env(new_env)
         yield
       ensure
         old_env.each do |key, value|
-          if value == false
-            ENV.delete key
-          else
-            ENV[key] = value
-          end
+          value == false ? ENV.delete(key) : ENV[key] = value
```
technoweenie: A little shorter. But is it _too_ short???

```suggestion
          value ? ENV[key] = value : ENV.delete(key)
```

spec/support/helper_methods.rb
```diff
@@ -11,11 +11,15 @@ def features(*features)
         @features = features
       end
 
-      def on_feature(name, &block)
+      def on_feature(name)
+        yield if block_given? && feature?(name)
+      end
+
+      def feature?(name)
```
technoweenie: Neato.

spec/spec_helper.rb
```diff
@@ -25,6 +25,7 @@
 
 SimpleCov.start do
   add_filter '/spec/'
+  add_filter '/lib/faraday/adapter/em_http_ssl_patch'
```
Insti: I really don't think this is a good idea.
If you've decided you don't want to test it, fine, but don't hide the fact that it's not tested.

Insti: After doing some research into what this is, I feel strongly that it should be tested. 
So I'll add it to my list of things to do.

lib/faraday/adapter/em_http.rb
iMacTia: Not sure I follow. I just added an extra check on `msg`, there was no removal

spec/support/helper_methods.rb
iMacTia: I thought about that, but then I wondered: why did whoever wrote this tested specifically against `false`?
Since `value` can be anything, changing the implementation to the shorter one you suggested will cause the condition to be extended to all `falsy` values.
For example, if `value` is `0` then it will be deleted, but under the current implementation it would remain

spec/spec_helper.rb
iMacTia: I've already agreed with the other maintainers that the SSL patch will be deprecated in v1.0 and dropped in future versions, hence my decision not to spend any time test-covering it, also because it would be hard to do it with unit-tests.

The current plan is to eventually cover it in some sort of integration tests, which we can add later. If you want to help in that direction then you're more than welcome 😃.

That said, I saw your other PR so it seems like there are other files that were ignored, so if we agree on merging that one, it makes sense to keep the file in

iMacTia:
**UPDATE:** After merging @Insti PR total coverage went down to 84%, so this PR now increases that to 88% 🎉

iMacTia:
@technoweenie we can get this merged and I can start working on `em-synchrony` adapter if you're happy with answers your comments 👍

spec/support/helper_methods.rb
technoweenie: `0` is truthy in ruby, but your point still stands because of `nil`.

PR #933, opened 2019-03-05
Rubocop: style/classvars

technoweenie:
This fixes the Style/ClassVars rubocop offenses. I replaced the `@@constants` stuff with a separate registry class.

TODO:

* [x] test `AdapterRegistry`


lib/faraday/rack_builder.rb
```diff
@@ -20,23 +22,19 @@ class StackLocked < RuntimeError; end
     # borrowed from ActiveSupport::Dependencies::Reference &
     # ActionDispatch::MiddlewareStack::Middleware
     class Handler
-      @@constants_mutex = Mutex.new
-      @@constants = Hash.new do |h, k|
-        value = k.respond_to?(:constantize) ? k.constantize : Object.const_get(k)
-        @@constants_mutex.synchronize { h[k] = value }
-      end
+      REGISTRY = Faraday::AdapterRegistry.new
```
technoweenie: I set the `AdapterRegistry` instance into a const so it is run during the initial process initialization (`require 'faraday'`), and not inside some code that could potentially be multi-threaded. 

I'm open to other approaches, as my ruby concurrency game is not strong.

technoweenie:
Turns out when you write tests, the code coverage improves :metal:

I'd appreciate a look at the spec, I haven't used rspec in ages :)

spec/faraday/adapter_registry_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+RSpec.describe Faraday::AdapterRegistry do
+  describe '.initialize' do
+    subject { described_class.new }
```
olleolleolle:
```suggestion
    subject(:registry) { described_class.new }
```

spec/faraday/adapter_registry_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    it { expect { subject.get(:FinFangFoom) }.to raise_error(NameError) }
```
olleolleolle:
```suggestion
    it { expect { registry.get(:FinFangFoom) }.to raise_error(NameError) }
```

olleolleolle: The suggestion is: name the subject when we use it explicitly. 

Would that feel OK?

[Longer description](http://www.betterspecs.org/#subject)

spec/faraday/adapter_registry_spec.rb
technoweenie: Good idea. There should be a cop for that :)

spec/faraday/adapter_registry_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+RSpec.describe Faraday::AdapterRegistry do
+  describe '.initialize' do
```
iMacTia: nit: although called during initialisation, `initialize` is still an instance method, so it should be prefixed with `#` instead of `.` (the latter is for class methods)

spec/faraday/adapter_registry_spec.rb
olleolleolle: `rubocop_rspec` is a bag of such cops. They're pretty good at making one refactor tests.

lib/faraday/adapter_registry.rb
```diff
@@ -0,0 +1,38 @@
...
+
+require 'monitor'
+
+module Faraday
+  # AdapterRegistry registers adapter class names so they can be looked up by a
+  # String or Symbol name.
+  class AdapterRegistry
+    def initialize
+      @lock = Monitor.new
+      @constants = nil
```
Insti: If we initialize `@constants` to `Hash.new` here, a few other lines can disappear.

lib/faraday/adapter_registry.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def get(name)
+      klass = @constants && @constants[name]
+      return klass if klass
+
+      klass =
+        if name.respond_to?(:constantize)
+          name.constantize
```
Insti: This line is untested by the existing unit tests.

Is this a useful feature?
I would think that `Object.const_get(name)` should be enough, but this may just be me not appreciating the benefits of `constantize`

iMacTia: Good point here, they should do the same thing but according to `benchmark`, Rails `#constantize` is around 4 times slower than `const_get`. Plus the former is only available under `active_support`, while the latter is always available. 

Is there a particular reason for not simply using `const_get`?

lib/faraday/adapter_registry.rb
iMacTia: Good catch, initializing it to `{}` or `Hash.new` here, we can avoid testing its initialization on `get` and `set`

lib/faraday/adapter_registry.rb
technoweenie: It was used because `Object.const_get` was not able to traverse module namespaces (ex: `Object.const_get('Faraday::Connection')`. Apparently that was changed some time in the last decade, so let's 🔥 constantize down!

technoweenie:
Thank you for the reviews, everyone :metal:

Insti:
It looks much nicer now, thanks :+1:

iMacTia:
Thanks @Insti for the additional feedbacks 👍

PR #936, opened 2019-03-07
Adds em-synchrony adapter to RSpec tests

iMacTia:
## Description
Covers em-synchrony adapter with RSpec
No support for `trace` and `connect` methods, unfortunately 😢.

## Additional Notes
Back to over 90% test coverage 💪 🎉 !

### DIFF
--- a/spec/faraday/adapter/em_synchrony_spec.rb
+++ b/spec/faraday/adapter/em_synchrony_spec.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+RSpec.describe Faraday::Adapter::EMSynchrony do
+  features :request_body_on_query_methods, :reason_phrase_parse, :skip_response_body_on_head, :parallel
+
+  it_behaves_like 'an adapter'
+
+  it 'allows to provide adapter specific configs' do
+    url = URI('https://example.com:1234')
+    adapter = Faraday::Adapter::EMSynchrony.new nil, inactivity_timeout: 20
+    req = adapter.create_request(url: url, request: {})
+
+    expect(req.connopts.inactivity_timeout).to eq(20)
+  end
+end
--- a/spec/faraday/adapter/em_synchrony_spec.rb
+++ b/spec/faraday/adapter/em_synchrony_spec.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 RSpec.describe Faraday::Adapter::EMSynchrony do
-  features :request_body_on_query_methods, :reason_phrase_parse, :skip_response_body_on_head, :parallel
+  features :request_body_on_query_methods, :reason_phrase_parse,
+           :skip_response_body_on_head, :parallel, :local_socket_binding
 
   it_behaves_like 'an adapter'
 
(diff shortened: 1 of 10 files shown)


spec/faraday/adapter/em_synchrony_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+RSpec.describe Faraday::Adapter::EMSynchrony do
+  features :request_body_on_query_methods, :reason_phrase_parse, :skip_response_body_on_head, :parallel
+
+  it_behaves_like 'an adapter'
+
+  it 'allows to provide adapter specific configs' do
+    url = URI('https://example.com:1234')
+    adapter = Faraday::Adapter::EMSynchrony.new nil, inactivity_timeout: 20
```
olleolleolle:
```suggestion
    adapter = described_class.new nil, inactivity_timeout: 20
```

spec/faraday/adapter/em_synchrony_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    req = adapter.create_request(url: url, request: {})
+
+    expect(req.connopts.inactivity_timeout).to eq(20)
```
olleolleolle: Q: Is `#connopts` shared by all kinds of adapters, or only by EmSynchrony? That is: Do we have an oppportunity for more shared test setup code?

(This is me being curious, not a blocker or a criticism.)

spec/faraday/adapter/em_synchrony_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+RSpec.describe Faraday::Adapter::EMSynchrony do
+  features :request_body_on_query_methods, :reason_phrase_parse, :skip_response_body_on_head, :parallel
```
olleolleolle: Perhaps this is a too-long line?

spec/faraday/adapter/em_synchrony_spec.rb
iMacTia: `create_request` returns an `EventMachine::HttpRequest` so it's only for `em-http` and `em-synchrony`

spec/faraday/adapter/em_synchrony_spec.rb
iMacTia: Good spot, other adapters tests are affected as well, so I'll change it everywhere for consistency 😃

spec/faraday/adapter/em_synchrony_spec.rb
iMacTia: Decided no to address it here as we still have to decide the project-wise max line length.
I personally think the default 80 characters is a bit too restrictive and a bit 80's, but we'll see once we decide to take down that cop.

spec/faraday/adapter/em_synchrony_spec.rb
Insti: If you're going to change it everywhere it should it be a separate pull request since it's not at all related to adding the em-synchrony test?

(For the record I prefer using the full class name rather than `described_class` but don't care enough to try and change anyone's mind about it. :man_shrugging: )

iMacTia: I admit I could have done that in a separate PR 😄.
It's going to be only one line per test file, so if @olleolleolle don't mind I'd keep the change here rather than moving it into another PR.

PR #938, opened 2019-03-07
chore: rubocop lint Metrics/LineLength

technoweenie:
* [x] finish


lib/faraday/dependency_loader.rb
```diff
@@ -4,15 +4,18 @@ module Faraday
...
+    # libraries
     def dependency(lib = nil)
       lib ? require(lib) : yield
     rescue LoadError, NameError => error
       self.load_error = error
     end
 
     def new(*)
-      raise "missing dependency for #{self}: #{load_error.message}" unless loaded?
+      if !loaded?
```
Insti: Is there a style decision to prefer `if !` over `unless`?

iMacTia:
My suggestion to whoever will end up reviewing this: review one commit at a time 😄

lib/faraday/adapter/em_synchrony.rb
```diff
@@ -126,6 +117,23 @@ def execute_single_request(env, request, http_method)
           end
         end
       end
+
+      def call_block(block)
+        client = nil
+
+        if !EM.reactor_running?
```
olleolleolle: This would read easier if conditional was flipped to be positive.

iMacTia: Good spot, I'm surprised RuboCop didn't catch that

lib/faraday.rb
```diff
@@ -78,8 +79,8 @@ class << self
     #   Faraday.new url: 'http://faraday.com',
     #               params: { page: 1 }
     #   # => Faraday::Connection to http://faraday.com?page=1
-    def new(url = nil, options = nil, &block)
-      options = options ? default_connection_options.merge(options) : default_connection_options
+    def new(url = nil, options = {}, &block)
+      options = default_connection_options.merge(options)
```
technoweenie: FWIW this will raise an exception if someone passes nil to `options`. I don't know why anyone would do that though. It also needlessly creates a new hash object for an optional argument. Probably not a big deal though.

lib/faraday/rack_builder.rb
```diff
@@ -198,19 +202,26 @@ def dup
...
+                request.path, request.params,
+                request.options.params_encoder
+              ),
               request.options, request.headers, connection.ssl,
               connection.parallel_manager)
     end
 
     private
 
+    LOCK_ERR = "can't modify middleware stack after making a request".freeze
```
technoweenie: nit: double quotes aren't needed here, and single quotes seem to be the preference in the rest of the code.

lib/faraday.rb
iMacTia: It's a good, I can fix the first point by defaulting to `nil` and doing `default_connection_options.merge(options || {})`, but an empty hash will still be initialised every time.

lib/faraday/rack_builder.rb
iMacTia: It's because of the `'` in `can't`, the other solution is to use single quotes but escape that single quote

technoweenie:
Looks great :+1: I can't approve since I started the PR :)

PR #941, opened 2019-03-09
Support standalone adapters

iMacTia:
## Description
This PR exposes a spec support file that external adapter gems can require in order to run all the relevant `shared_examples` to test an adapter.
You can see this tested (against this branch) here: https://github.com/lostisland/faraday-http/pull/1


technoweenie:
Great work enabling Faraday to live its best life :metal:

PR #958, opened 2019-03-29
CI: yard junk

technoweenie:
This PR **fails CI builds if `yard-junk` returns any issues**.

Some questions though:

* Do we really care about _all_ issues that yard-junk reports?
* Is there a way to ignore certain issues?
* Are there things that yard-junk misses that we would want when verifying code docs?

My approach is to add a new step to the `linting` job that runs yard-junk. It currently fails on Faraday with:

```
lib/faraday/connection.rb:219: [UnknownParam] @param tag has unknown parameter name: url
lib/faraday/connection.rb:221: [UnknownParam] @param tag has unknown parameter name: params
lib/faraday/connection.rb:222: [UnknownParam] @param tag has unknown parameter name: headers
```

However, it doesn't seem to complain if new code is added without documentation. That should be caught by rubocop though. Locally, it looks like this:

```sh
$ bundle exec rubocop
Inspecting 97 files
.................................................C...............................................

Offenses:

lib/faraday/foo.rb:1:1: C: Style/FrozenStringLiteralComment: Missing magic comment # frozen_string_literal: true.
module Faraday
^
lib/faraday/foo.rb:2:3: C: Style/Documentation: Missing top-level class documentation comment.
  class Foo
  ^^^^^
lib/faraday/foo.rb:3:5: C: Style/EmptyMethod: Put empty method definitions on a single line. (https://github.com/rubocop-hq/ruby-style-guide#no-single-line-methods)
    def initialize(a, b = nil) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/faraday/foo.rb:3:20: C: Naming/UncommunicativeMethodParamName: Method parameter must be at least 3 characters long.
    def initialize(a, b = nil)
                   ^
lib/faraday/foo.rb:3:23: C: Naming/UncommunicativeMethodParamName: Method parameter must be at least 3 characters long.
    def initialize(a, b = nil)
                      ^

97 files inspected, 5 offenses detected

$ yard-junk --path=lib
Running YardJunk janitor (version 0.0.7)...


Problems
--------
mistyped tags or other typos in documentation

lib/faraday/connection.rb:219: [UnknownParam] @param tag has unknown parameter name: url
lib/faraday/connection.rb:221: [UnknownParam] @param tag has unknown parameter name: params
lib/faraday/connection.rb:222: [UnknownParam] @param tag has unknown parameter name: headers

0 failures, 3 problems (2 seconds to run)
```

TODO

* [x] Squash commit those bad messages away

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
       - checkout
       - run:
           name: Create test-results dir
-          command: mkdir -p ~/test-results/rspec
+          command: mkdir -p ~/test-results/rspec && mkdir -p ~/test-results/linting
           when: always
       - persist_to_workspace:
           root: .
@@ -69,11 +69,16 @@ jobs:
       - run:
           name: RuboCop
           command: |
+            echo "starting rubocop..."
+            ls -al ~/test-results
             gem install rubocop rubocop-junit_formatter  --no-document
             rubocop --require rubocop/formatter/junit_formatter \
                     --format progress \
                     --format RuboCop::Formatter::JUnitFormatter \
                     --out ~/test-results/linting/rubocop.xml
+            echo "rubocop finished..."
+            ls -al ~/test-results
+            ls -al ~/test-results/linting
           when: always
       - run:
           name: Yard-Junk
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,17 +67,11 @@ jobs:
       - run:
           name: RuboCop
           command: |
-            echo "starting rubocop..."
-            ls -al ~
             gem install rubocop rubocop-junit_formatter  --no-document
             rubocop --require rubocop/formatter/junit_formatter \
                     --format progress \
                     --format RuboCop::Formatter::JUnitFormatter \
                     --out ~/test-results/linting/rubocop.xml
-            echo "rubocop finished..."
-            ls -al ~
-            ls -al ~/test-results
-            ls -al ~/test-results/linting
           when: always
       - run:
           name: Yard-Junk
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -51,7 +51,10 @@ jobs:
       - image: circleci/ruby:2.6
     steps:
       - checkout
-      - run: "mkdir -p ~/test-results/rspec && mkdir -p ~/test-results/rubocop"
+      - run:
+          name: Create test-results dir
+          command: mkdir -p ~/test-results/rspec
```
iMacTia: Not sure if this is actually an issue or not, but should we also create the `~/test-results/linting` folder that we use in row 76?

technoweenie: My initial attempt added yard-junk as its own job separate from the rubocop one. However, it [failed](https://circleci.com/gh/lostisland/faraday/142?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) because `~/test-results/yard` doesn't exist, even though I [explicitly created it](https://github.com/lostisland/faraday/pull/958/commits/e503819a38d1f6ff8870369e9cd1df5d549e283b).

From further testing, `~/test-results/rubocop` doesn't exist either for the currrent `linting` job. It looks like Rubocop takes the initiative in creating the empty directories.

Though, perhaps instead of relying on Rubocop to take the initiative, the `linting` job should handle its own directory creation as an initial `run` step?

technoweenie: That's fine, except they're not being persisted outside of the `checkout_code` job's container. It's only persisting `.` (which is `~/project`), so `~/test-results` is skipped. 

I've updated this to create the test-results dirs in the same container that they're used.

PR #962, opened 2019-03-31
Fix Memoized w/o block

### LINKED ISSUE #961: .rvm/gems/ruby-head/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: tried to create Proc object without a block
## Basic Info

* Faraday Version: 0.15.4
* Ruby Version: ruby 2.7.0dev (2019-03-18 trunk 67296) [x86_64-darwin17]

## Issue description

I get many warnings like this:

```
.rvm/gems/ruby-head/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: tried to create Proc object without a block
```

## Steps to reproduce

- Using latest ruby-head (e.g. 2.7.0dev)
- I'm using faraday through discordrb but I think this might be for any faraday usage.

* Are you using the latest Faraday version? yes

technoweenie:
This fixes `Faraday::Options.memozied` so it stops warning on Ruby 2.7.0. `Proc.new` without an explicit block is now considered a warning: https://docs.ruby-lang.org/en/trunk/NEWS.html

Fixes #961

## Todos

* [ ] squash merge

### DIFF
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -171,11 +171,11 @@ module Faraday
       @attribute_options ||= {}
     end
 
-    def self.memoized(key)
+    def self.memoized(key, &block)
       if !block_given?
         raise ArgumentError, "#memoized must be called with a block"
       end
-      memoized_attributes[key.to_sym] = Proc.new
+      memoized_attributes[key.to_sym] = block
       class_eval <<-RUBY, __FILE__, __LINE__ + 1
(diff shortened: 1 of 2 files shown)


technoweenie:
@olleolleolle To be fair, there isn't any known code that would raise that `ArgumentError`. That was just the path I went down at first because I forgot to run `rbenv local 2.7.0-dev` after installing it. I left that check in `#memoized` though just in case ;)

PR #969, opened 2019-04-08
Skip tests that do not run on JRuby

olleolleolle:
Add RSpec skips to test cases that can not run on JRuby.

## Description

There are adapters that do not support JRuby.

This is about investigating a JRuby build for the CI.

We do not have a maintainer which can work focused with JRuby, right now, so we want to keep it in there, but not as a _required_ job.

<details>

## Todos

- [x] Be sure that the desired behavior is achieved.
- [x] Investigate and write up `Invalid header type`
- [x] Turn this into GitHub Action format (future)

</details>

### DIFF
--- a/spec/faraday/adapter/em_http_spec.rb
+++ b/spec/faraday/adapter/em_http_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe Faraday::Adapter::EMHttp do
+RSpec.describe Faraday::Adapter::EMHttp, unless: defined?(JRUBY_VERSION) do
   features :request_body_on_query_methods, :reason_phrase_parse, :trace_method,
            :skip_response_body_on_head, :parallel, :local_socket_binding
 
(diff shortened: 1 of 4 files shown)


olleolleolle:
https://circleci.com/gh/lostisland/faraday/355

> Invalid header type

in the RubyParser.java. I have seen this failure mode before!

technoweenie:
Ah, good ole `header_value_type`. I'm not sure what else it would be... `:mixed` looks like a fine value.

https://github.com/tmm1/http_parser.rb/blob/ec53181a4c944f8c308830b84b08fc8f71075ba6/ext/ruby_http_parser/org/ruby_http_parser/RubyHttpParser.java#L477-L479

olleolleolle:
OK, the failing CI run on jruby92 is upload to Coverage website. Hm!

iMacTia:
@olleolleolle I guess next step here is to add JRuby to the GitHub Actions matrix?

olleolleolle:
One thing about GH actions, they don't take Allow Failure easily

PR #971, opened 2019-04-10
expose the gemfile.lock file in CircleCI

technoweenie:
This exposes the Gemfile.lock in CI so we can see the _exact_ ruby gems its using while debugging failures [like this one](https://circleci.com/gh/lostisland/faraday/355).


technoweenie:
Continuous_Integration_and_Deployment

The above should be visible on any of the ruby ci jobs for the most recent commit. Here's an [example Gemfile.lock](https://374-417389-gh.circle-artifacts.com/0/home/circleci/project/Gemfile.lock).

olleolleolle:
Thanks for adding the screenshot

PR #972, opened 2019-04-13
Allow setting min/max SSL version for a Net::HTTP::Persistent connection

bdewater:
Basically the same as https://github.com/lostisland/faraday/pull/792, possible now https://github.com/drbrain/net-http-persistent/pull/94 is merged 😄

PR #973, opened 2019-04-14
Refactor: Extract repeated calls to a shared map

olleolleolle:
## Description

Follows up #972 by extracting some code, to make it pass CodeClimate checks.

**Update**: It fixed the issue it came to fix.

### DIFF
--- a/spec/faraday/adapter/net_http_persistent_spec.rb
+++ b/spec/faraday/adapter/net_http_persistent_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe Faraday::Adapter::NetHttpPersistent do
       {
         headers: { 'X-Faraday-Adapter' => adapter },
         ssl: {
-          min_version: '1.0'
+          min_version: :TLS1_2
         }
       }
     end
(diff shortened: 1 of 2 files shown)


spec/faraday/adapter/net_http_persistent_spec.rb
```diff
@@ -40,4 +40,26 @@
     # `pool` is only present in net_http_persistent >= 3.0
     expect(http.pool.size).to eq(5) if http.respond_to?(:pool)
   end
+
+  context 'min_version' do
+    let(:conn_options) do
+      {
+        headers: { 'X-Faraday-Adapter' => adapter },
+        ssl: {
+          min_version: '1.0'
```
bdewater: IMO it's nice that tests can serve as documentation, a value like `:TLS1_2` (which is [mapped](https://github.com/ruby/openssl/blob/01b23fa8eee2ff9cc4ef5f6fabca6d999e0979c2/ext/openssl/ossl_ssl.c#L120-L127) to `OpenSSL::SSL::TLS1_2_VERSION`) would be a real-life example that can be used here.

olleolleolle: Thanks for chiming in! Alright, let's use a realistic value here, it wont hurt readability, or fool anyone.

PR #979, opened 2019-05-16
Add exception for conflict response code

lucasmoreno:
## Description
I wanted to rescue an exception for 409 response code and I noticed that it was not implemented yet. So, here it is 😄

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -61,6 +61,10 @@ module Faraday
   class ProxyAuthError < ClientError
   end
 
+  # Raised by Faraday::Response::RaiseError in case of a 409 response.
+  class ConflictError < ClientError
+  end
+
   # Raised by Faraday::Response::RaiseError in case of a 422 response.
   class UnprocessableEntityError < ClientError
   end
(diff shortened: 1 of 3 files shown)


iMacTia:
Hi @lucasmoreno and thanks for working on this.
I'd just like to clarify that we didn't create a custom error class for every http status code because that would be just too much 😅, so we created them for Rails' most common ones and the one related to proxy authentication.
All other http status codes can be rescues with the generic `ClientError` and then you can access the status through the exception `response` property.

I personally don't have anything against this PR per-se, but I don't think that 409 is such a common http status to justify it's inclusion in the custom set of exceptions.

Would like to hear the opinion of the other mantainers though 😄

lucasmoreno:
Yeah, it makes sense! We should not have all the status codes mapped or the code would get huge 😄 

I think 409 should have its specific exception because some APIs use it to indicate that some entity was already created and, in some scenarios, it can be understood as an indication that "the job was already done".

iMacTia:
Discussed with the other maintainers offline and decided to give it the green light 👍.
Thanks for contributing!

PR #982, opened 2019-05-23
Faraday Website

iMacTia:
## Description
Introduces a GitHub-pages-compatible website based on Jekyll and the Type theme.
The website will also be used to reorganise and extend the README content with tutorials, mini-guides and general documentation.
This will NOT replace the YARD-generated API Documentation.
Supersedes #970.

## Todos
List any remaining work that needs to be done, i.e:
- [x] Complete moving README content into separate pages
- [x] Add explanation in README on how to contribute to the website (Jekyll instructions)
- [x] Add more documentation pages
  - [x] Team
  - [x] Introduction
  - [x] Finish Middleware pages
    - [x] Multipart
    - [x] UrlEncoded
    - [x] Retry
    - [x] Instrumentation
    - [ ] List and links to popular 3rd party middleware
  - [ ] SSL
  - [ ] Parallel Requests
- [x] Add some navigation between documentation pages (next/prev/menu)
- [x] Add some branding/styles to website
- [ ] First Blog page (article on v1.0 or new website?)
- [x] Correct links to Faraday website (currently https://???)

PR #986, opened 2019-05-25
cleanup readme a bit so it looks less dated

grosser:


iMacTia:
Hi @grosser, thanks for attempting to cleaning up the README, indeed it's looking old and it need some refreshing! The good news is that we're going much further than that, and we're also splitting it into separate, more detailed pages, plus moving it onto the new Faraday website 😎

You can see what we've done so far on by looking at #982, still a WIP but close to completion 👍
Most of the points you raised were already addressed on that PR, plus the content has been moved onto other files, so I won't be able to merge your PR unfortunately as it will cause lots of conflicts otherwise.

Hope you like the redesign of the README and the new website!

PR #988, opened 2019-05-28
Replace nil values with blank strings in application/x-www-form-urlencoded body

jeremiahmegel:
From what I can tell, params in `application/x-www-form-urlencoded` [must have](https://url.spec.whatwg.org/#urlencoded-serializing) `=` characters in order to be valid, even if their values are `nil`. We initially discovered this in a project of ours that was `POST`-ing data to a Microsoft IIS 8.5 server - it couldn't parse the request body without the `=` characters.

This PR fixes the issue by replacing any `nil` values in the request body with blank strings - which causes the `to_query` method to include `=` characters.

### DIFF
--- a/lib/faraday/request/url_encoded.rb
+++ b/lib/faraday/request/url_encoded.rb
@@ -17,7 +17,7 @@ module Faraday
       # @param env [Faraday::Env]
       def call(env)
         match_content_type(env) do |data|
-          params = Faraday::Utils::ParamsHash[data]
+          params = replace_nil_with_blank_str(Faraday::Utils::ParamsHash[data])
           env.body = params.to_query(env.params_encoder)
         end
         @app.call env
@@ -30,6 +30,24 @@ module Faraday
 
         env.request_headers[CONTENT_TYPE] ||= self.class.mime_type
         yield(env.body) unless env.body.respond_to?(:to_str)
+      end
+
+      def replace_nil_with_blank_str(data)
+        if data.nil?
+          ''
+        elsif data.is_a? Array
+          data.map(&method(:replace_nil_with_blank_str))
+        elsif data.is_a? Hash
+          new_data = data.class.new
+
+          data.each do |k, v|
+            new_data[k] = replace_nil_with_blank_str(v)
+          end
+
+          new_data
+        else
+          data
+        end
       end
 
       # @param env [Faraday::Env]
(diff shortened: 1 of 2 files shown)


jeremiahmegel:
Thanks @iMacTia!

If I change it in the encoders, that would also affect how URL query strings are built - and it looks like that behavior was [intentionally added](https://github.com/lostisland/faraday/pull/531/commits/f78d51da17b5af4ff8460d142f843e31aa0875c0) in #531 for `NestedParamsEncoder` (haven't checked when it was added for `FlatParamsEncoder`). Plus I haven't been able to find anything that would suggest omitting the `=` is invalid for a query string (just the `x-www-form-urlencoded` request body).

What do you think of passing an `options` hash as an arg into [`NestedParamsEncoder#encode`](https://github.com/lostisland/faraday/blob/d76489a0629beb9070edfb2a69730cea1eb21a9c/lib/faraday/encoders/nested_params_encoder.rb#L11) (same for `FlatParamsEncoder`) via [`Utils::ParamsHash#to_query`](https://github.com/lostisland/faraday/blob/d76489a0629beb9070edfb2a69730cea1eb21a9c/lib/faraday/utils/params_hash.rb#L50)? That way `Request::UrlEncoded` could pass some option that would tell it to keep the `=`.

iMacTia:
@jeremiahmegel I've just realised (remembered actually, not sure why it didn't come to me sooner) something important. Have a look at #345.

As you can see from that discussion, there's a subtle but important difference between `?a=` and `?a`, with the former being parsed as `a: ''` and the latter as `a: nil`.

Now, I understand this is probably Ruby-like, and IIS doesn't entirely like it, but this is going to be the same for both query-string params and body.
If we change the standard behaviour now, it will surely cause issues for anyone using Faraday to connect to a Ruby server (e.g. Rails or Sinatra).

At this point, the option you suggested would make sense, although I'm a bit conflicted now on how much useful this might be for the community. I've been using IIS for some time and I remember it being particularly fussy on some things, so I'd be inclined not to add this to the Faraday codebase since it doesn't seem to be a very diffused problem. An extra option on the codebase is an extra burden on maintenance for us, with a very small users-pool.

I'd rather suggest you to consider sub-classing `NestedParamsEncoder` into your custom class and provide that class to the Faraday connection calling the IIS server. That's the main reason why we made that configurable. This way we won't over-complicate the param-encoders to support all the different server types, and instead we focus on the Ruby world, while providing customisation options.

Appreciate this is not the answer you were expecting, but this should be equally simple for you to implement. Should that not be the case, please feel free to open an Issue and I'll happily help out

jeremiahmegel:
@iMacTia That makes sense. Thanks for taking a look at this anyway - and happy coding!

PR #994, opened 2019-07-08
Website Navigation Fix

iMacTia:
## Description
Fixes website navigation by adding trailing slashes to /index.md files.
See https://stackoverflow.com/questions/54727643/trailing-slashes-in-jekyll-github-pages-site-cause-404

## Additional Notes
Tested locally and works (but it was working before the change as well.
Hard to say if it will work on GitHub Pages, but we can't really test before merging

### DIFF
--- a/docs/adapters/index.md
+++ b/docs/adapters/index.md
@@ -1,7 +1,7 @@
 ---
 layout: documentation
 title: "Adapters"
-permalink: /adapters
+permalink: /adapters/
 order: 2
 ---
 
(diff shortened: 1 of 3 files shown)


technoweenie:
That looks like it should fix the links in the top right have, from `/faraday/{section}` to `/faraday/{section}/`. That _should_ work, unless GH Pages won't support a trailing slash in the URL like that. Considering that the site is only soft launched at an unused url, I think we can just ship this and cross our fingers.

PR #1002, opened 2019-07-13
Reuse Patron session with mutex

### LINKED ISSUE #1001: Patron adapter does not support keep-alive
## Basic Info

* Faraday Version: master
* Ruby Version: 2.5.5

## Issue description

The pull-request #796 removed the cached patron session to avoid issues with multi-threading (patron sessions are not thread-safe). To use keep-alive, the session has to be reused and this is now impossible... Can we revert the changes introduced in #796 and use e.g. a [Mutex](https://ruby-doc.org/core-2.5.5/Mutex.html) to made it thread-safe in faraday?

## Steps to reproduce

Test Server:

    require 'webrick'
    
    server = WEBrick::HTTPServer.new Port: 4567
    server.mount_proc '/' do |req, res|
      res.body = "peerport=#{req.peeraddr[1]}"
    end
    server.start

Test Client:

    require 'patron'

    s1 = Patron::Session.new
    puts "First request on s1: #{s1.get('http://localhost:4567').body}"
    puts "Second request on s1: #{s1.get('http://localhost:4567').body}"
    s2 = Patron::Session.new
    puts "Third request on s2: #{s2.get('http://localhost:4567').body}"

Output:

    First request on s1: peerport=39218
    Second request on s1: peerport=39218
    Third request on s2: peerport=39220

The first and second requests reuse a session and using keep-alive on the same connection. The third request use a new session, creates a new connection and does not use keep-alive.

derSascha:
Reuse Patron session to support keep-alive/connection reuse and make
them thread-safe using a mutex.

## Description

This reverts the changes from PR #796 and reuse the patron session for keep-alive/connection reuse. To be thread safe, this adds a mutex around the patron session. See [persisted connections](https://github.com/toland/patron#persistent-connections]) and [Threads](https://github.com/toland/patron#persistent-connections) in the patron documentation for more details.

Fixes #1001 

## Todos

- [x] Tests
- [x] Documentation

### DIFF
--- a/spec/faraday/adapter/patron_spec.rb
+++ b/spec/faraday/adapter/patron_spec.rb
@@ -15,4 +15,21 @@ RSpec.describe Faraday::Adapter::Patron do
 
     expect { conn.get('/') }.to raise_error(RuntimeError, 'Configuration block called')
   end
+
+  it 'reuses the patron session for keep-alive' do
+    last_session = nil
+    conn = Faraday.new do |f|
+      f.adapter :patron do |session|
+        if last_session
+          expect(session).to eq last_session
+        else
+          last_session = session
+        end
+      end
+    end
+
+    stub_request(:get, 'http://example.com')
+    conn.get('http://example.com/')
+    conn.get('http://example.com/')
+  end
 end
--- a/docs/adapters/patron.md
+++ b/docs/adapters/patron.md
@@ -18,6 +18,12 @@ conn = Faraday.new(...) do |f|
 end
 ```
 
+## Multithreading
+
+This adapter use a mutex around the patron session to be thread-safe.
+A [connection_pool](https://rubygems.org/gems/connection_pool) can be
+used to share multiple connections between threads.
+
 ## Links
 
 * [Gem RDoc][rdoc]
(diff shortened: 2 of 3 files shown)


spec/faraday/adapter/patron_spec.rb
```diff
@@ -15,4 +15,21 @@
 
     expect { conn.get('/') }.to raise_error(RuntimeError, 'Configuration block called')
   end
+
+  it 'reuses the patron session for keep-alive' do
+    last_session = nil
+    conn = Faraday.new do |f|
+      f.adapter :patron do |session|
+        if last_session
+          expect(session).to eq last_session
```
iMacTia: Given `@config_block&.call(session)` will be called only once due to the cache, I wouldn't expect this `expect` to ever be executed.

Instead, you could use a variable to check that this block is only called once:

```ruby
    config_calls = 0
    conn = Faraday.new do |f|
      f.adapter :patron do |session|
        config_calls += 1
      end
    end
```

and then after the two `conn.get('http://example.com/')` you can

```ruby
    expect(config_calls).to eq(1)
```

iMacTia:
@derSascha I really appreciate you taking a stab at this, however as the discussion in #1001 shows, this is not the solution we'd like to go with, so I'm closing this PR.

I'll do my best to help agreeing and implementing a solution.

PR #1003, opened 2019-07-18
Add support for setting Ruby Net::HTTP `read_timeout` option separately

springerigor:
## Description
Right now `timeout` setting sets all the timeout values (`read`, `open` & `write` if available). To unify the API with `Net::HTTP` one (which has a dedicated method for `read_timeout` as well) I propose extending the `RequestOptions` by `read_timeout`.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -173,6 +173,8 @@ module Faraday
           end
         end
         http.open_timeout = req[:open_timeout] if req[:open_timeout]
+        http.read_timeout = req[:read_timeout] if req[:read_timeout]
+
         if req[:write_timeout] && http.respond_to?(:write_timeout=)
           http.write_timeout = req[:write_timeout]
         end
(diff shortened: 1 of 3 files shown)


iMacTia:
Hi @springerigor, thanks for the addition.
I'm not sure why we were not supporting this yet, so thanks for adding it to the `Net::HTTP` adapter.
Do you think you might have some extra time to have a look at the other adapters and see if the new options can be easily supported there as well? This would help keeping them consistent as much as possible.

technoweenie:
Hi @springerigor: You never replied, so I've continued work on it in #1022. Thanks for your help, I'll preserve your original commit :)

@iMacTia:
> I'm not sure why we were not supporting this yet

I think you can track my understanding of HTTP timeouts through Faraday ;) Basically, `:timeout` was the original setting, while `:write_timeout` and `:open_timeout` came later as users had to set other timeout options. I don't think `:read_timeout` is required, but it's consistent with the other timeout settings. I suppose `:timeout` can stay as a default setting.

PR #1006, opened 2019-07-24
Introduces Connection pool in compatible adapters

iMacTia:
## Description
Attempt to introduce a `ConnectionPool` in compatible adapters.

## Todos
- [ ] Tests: Can we write better tests? I tried executing a request before the previous one was completed, but the number of available connections in the pool was not decreasing (see https://github.com/lostisland/faraday/commit/596d420adc8d7e75ad6d78d2827f9ca91ff49dc2)
- [ ] There's a strange error popping up only sometimes, possibly a race condition. Any help on finding the actual cause would be welcome:
```
NoMethodError:
    undefined method `+' for nil:NilClass
    # /Users/mattiagiuffrida/.rvm/gems/ruby-2.5.1/gems/connection_pool-2.2.2/lib/connection_pool.rb:89:in `checkout'
    # /Users/mattiagiuffrida/.rvm/gems/ruby-2.5.1/gems/connection_pool-2.2.2/lib/connection_pool.rb:62:in `block in with'
    # /Users/mattiagiuffrida/.rvm/gems/ruby-2.5.1/gems/connection_pool-2.2.2/lib/connection_pool.rb:61:in `handle_interrupt'
    # /Users/mattiagiuffrida/.rvm/gems/ruby-2.5.1/gems/connection_pool-2.2.2/lib/connection_pool.rb:61:in `with'
    # ./lib/faraday/adapter/patron.rb:19:in `call'
    # ./lib/faraday/response.rb:11:in `call'
    # ./lib/faraday/request/url_encoded.rb:23:in `call'
    # ./lib/faraday/request/multipart.rb:25:in `call'
    # ./lib/faraday/rack_builder.rb:153:in `build_response'
    # ./lib/faraday/connection.rb:504:in `run_request'
    # ./lib/faraday/connection.rb:233:in `options'
    # ./spec/support/shared_examples/request_method.rb:94:in `public_send'
    # ./spec/support/shared_examples/request_method.rb:94:in `block (2 levels) in <top (required)>'
```

## Additional Notes
I've only added the connection pool to Patron for now, trying to address #1001.
`Net::HTTP` adapter is actually incompatible by design, because we create a different object based on the `env`, so there's no way to reuse connections unless we first refactor that bit.
I'll have a look at the other adapters after receiving some preliminary feedbacks (cc @technoweenie @olleolleolle @julik)

### DIFF
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.3'
 
+  spec.add_dependency 'connection_pool', '~> 2.2'
   spec.add_dependency 'multipart-post', '>= 1.2', '< 3'
 
   spec.require_paths = %w[lib spec/external_adapters]
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faraday/adapter/concerns/connection_pooling'
+
 module Faraday
   # Base class for all Faraday adapters. Adapters are
   # responsible for fulfilling a Faraday request.
@@ -40,10 +42,14 @@ module Faraday
     extend Parallelism
     self.supports_parallel = false
 
+    include ConnectionPooling
+    self.supports_pooling = false
+
     def initialize(_app = nil, opts = {}, &block)
       @app = ->(env) { env.response }
       @connection_options = opts
       @config_block = block
+      initialize_pool(opts[:pool] || {}) if self.class.supports_pooling
     end
 
     def call(env)
--- a/lib/faraday/adapter/concerns/connection_pooling.rb
+++ b/lib/faraday/adapter/concerns/connection_pooling.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+# This module marks an Adapter as supporting connection pooling.
+module ConnectionPooling
+  def self.included(base)
+    base.extend(ClassMethods)
+  end
+
+  # Class methods injected into the class including this module.
+  module ClassMethods
+    attr_accessor :supports_pooling
+
+    def inherited(subclass)
+      super
+      subclass.supports_pooling = supports_pooling
+    end
+  end
+
+  attr_reader :pool
+
+  MISSING_CONNECTION_ERROR = 'You need to define a `connection` method' \
+    'in order to support connection pooling!'
+
+  # Initializes the connection pool.
+  #
+  # @param opts [Hash] the options to pass to `ConnectionPool` initializer.
+  def initialize_pool(opts = {})
+    ensure_connection_defined!
+    @pool = ConnectionPool.new(opts, &method(:connection))
+  end
+
+  # Checks if `connection` method exists and raises an error otherwise.
+  def ensure_connection_defined!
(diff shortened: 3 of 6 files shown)


technoweenie:
Dang, I totally forgot to review this. It looks solid on an initial pass.

> Can we test concurrent use of the pool?

I think this would be better handled in faraday-live. I'll work on adding tests as a way to get familiar with this patch, and hopefully find that intermittent bug.

technoweenie:
Unfortunately, I don't think this is going to work without some major changes. Ideally, the pool should spit out a fully configured session object so the adapter can make the request. Faraday adapters were designed around a single method (`#call(env)`), so they configure and use the client each request. With this PR, the Patron and HTTPClient adapters modify the object after it's out of the pool.

The Patron adapter runs the config block _each_ for each request:

```ruby
Faraday.new do |f|
  f.adapter :patron do |session|
    session.max_redirects = 10
  end
end
```

https://github.com/lostisland/faraday/blob/85672891af424f9fadf42dacccc618d058e956e0/lib/faraday/adapter/patron.rb#L19-L22

This can easily be fixed by calling `@config_block` when the pool is initialized:

```diff
--- a/lib/faraday/adapter/concerns/connection_pooling.rb
+++ b/lib/faraday/adapter/concerns/connection_pooling.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'connection_pool'

 # This module marks an Adapter as supporting connection pooling.
 module ConnectionPooling
@@ -26,11 +27,20 @@ module ConnectionPooling
   # @param opts [Hash] the options to pass to `ConnectionPool` initializer.
   def initialize_pool(opts = {})
     ensure_connection_defined!
-    @pool = ConnectionPool.new(opts, &method(:connection))
+
+    if !@config_block
+      @pool = ConnectionPool.new(opts, &method(:connection))
+      return
+    end
+
+    @pool = ConnectionPool.new opts do
+      connection.tap { |c| @config_block.call(c) }
+    end
   end
```

Both adapters only set certain SSL settings on the session if the current request is to an HTTPS server:

https://github.com/lostisland/faraday/blob/85672891af424f9fadf42dacccc618d058e956e0/lib/faraday/adapter/httpclient.rb#L59-L61

https://github.com/lostisland/faraday/blob/85672891af424f9fadf42dacccc618d058e956e0/lib/faraday/adapter/patron.rb#L66-L69

## The Work

Here's another way to approach this. It's a lot more work, but I think the benefits will be good.

1. Add `#build_connection` to every adapter, which outputs the fully-configured http client from options.
2. Move settings that configure the HTTP connection from `RequestOptions` to `ConnectionOptions`: `:proxy, :bind, :timeout, :open_timeout, :write_timeout`
3. Teach `Adapter#initialize` to save the connection if it can:
  * If _any_ of the moved `RequestOptions` settings are used, skip this!
  * If the adapter supports pooling, `@pool = ConnectionPool.new(opts, &method(:build_connection))`
  * If the adapter does not support pooling, `@conn = build_connection`
4. Add `connection(&block)` to yield a connection for the adapter to use:
  * If _any_ of the moved `RequestOptions` settings are used, yield `#build_connection)
  * If the adapter supports pooling, pass block to `@pool.with`
  * If the adapter does not support pooling, yield `@conn`

How does this sound? Is it worth trying to get in for Faraday v1.0?

Issue #989, opened 2019-06-03
FYI: Warning after upgrading this gem to 2.7.0-preview1


jasnow:
## Basic Info

* Faraday Version: 0.15.4
* Ruby Version: 2.7.0-preview1

## Issue description
Getting this warning after upgrading to Ruby 2.7.0-preview1:
```
faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the 
given block  using Proc.new is deprecated; use `&block` instead
```
and
```
faraday-0.15.4/lib/faraday/rack_builder.rb:55: warning: Capturing the 
given block using Proc.new is deprecated; use `&block` instead
```
## Steps to reproduce
Just upgrade a repo to Ruby 2.7.0-preview1 and then run "rake" command.
Here is a link to other repos working on this issue:
```
https://github.com/aws/aws-sdk-ruby/issues/2058
```

iMacTia:
Hi @jasnow, thanks for raising this.
I can confirm this has already been fixed in `master` as of #962 👍

jasnow:
I have 8 repos that use faraday gem (mostly as a sub-dependency).

jasnow:
@iMacTia - Are we close to releasing a new version of this gem?

iMacTia:
@jasnow quite close, although next version is going to be v1.0.
Don't be too scared about the major jump though, we did our best to make it backwards-compatible to minimise disruptions. I'm sorry I can't be more precise at the moment on the exact release date, but all the maintainers are working hard to make it happen.

jasnow:
@iMacTia - Thanks for the update - I would normally just use the edge version on GitHub, but this gem is a sub-dependency for repos.

jasnow:
FYI: https://github.com/lostisland/faraday/pull/993

PR #1009, opened 2019-08-15
This fixes warnings on Ruby 2.7

tenderlove:
Ruby 2.7 issues warnings for `Proc.new` without a block:

```
warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
```

This patch converts the `Proc.new` calls in to `&block`

### DIFF
--- a/lib/faraday/adapter/em_http.rb
+++ b/lib/faraday/adapter/em_http.rb
@@ -229,11 +229,11 @@ module Faraday
           @running
         end
 
-        def add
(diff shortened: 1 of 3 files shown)


jasnow:
Connected with https://github.com/lostisland/faraday/issues/989

zenspider:
This was merged in August and there's been a release since then in October, but it doesn't seem to contain these fixes. Can we get this released to the non-`pre` line?

PR #1013, opened 2019-08-29
Backport #1009

iakobos:
## Description
This backports #1009 to v0.15.x. I'm not sure what the base branch should be since versions are just tags, so I left it as `master`. I just cherry-picked 27d9c93f3ab622e17893320d02a36251e9b110d1 onto `v0.15.4`.

## Rationale
We use many third-party gems that effectively specify `faraday < 1`, so this allows us to continue to use those gems without warnings on Ruby 2.7 (assuming it's included in a `v0.15.5` release).

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -64,8 +64,7 @@ module Faraday
     #     :params => {:page => 1}
     #
     # Returns a Faraday::Connection.
-    def new(url = nil, options = nil)
-      block = block_given? ? Proc.new : nil
+    def new(url = nil, options = nil, &block)
       options = options ? default_connection_options.merge(options) : default_connection_options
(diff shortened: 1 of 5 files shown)


iMacTia:
Sounds reasonable.
The correct base branch for 0.15 is the one called `0.1x`
If you branch out from that and cherry-pick again, the conflicts should go away.

technoweenie:
I changed the base above to `0.1x`, but I think it failed because it doesn't have our CI config. I'll work on that separately, and come back to this once I get CI working.

technoweenie:
@iakobos thanks again. I just pushed a `0.16.x` branch for future backports. The actual v0.16.0 gem will be pushed today.

Issue #769, opened 2018-02-08
Issue with multipart/form data request with json body


ff-sushild:
## Basic Info

* Faraday Version: Faraday v0.12.2 & v0.14.0
* Ruby Version: 2.3.3

## Issue description
When trying to POST multipart data (file & json), remote server is not recognizing json data parameter and failing with validation error as it is required parameter.
```ruby
connection = Faraday.new(url: "#{uri.scheme}://#{uri.host}:#{uri.port}") do |faraday|
              faraday.request :multipart
              faraday.request :url_encoded

              faraday.adapter :net_http
            end

request_data = { file: Faraday::UploadIO.new('somefile.png', 'image/png'), 
                 jsonBody: JSON.dump({ id: 'foo', name: 'bar' }) }

response = connection.post(uri.request_uri) do |request|
  request = Ff::Api::RequestHeaders.set(request, self.api_options)
  request.headers['content-type'] = 'multipart/form-data; boundary=-----------RubyMultipartPost'
  request.body = request_data
end
```

Does anyone see any obvious mistake here?

iMacTia:
Hi @ff-sushild,
unfortunately the Faraday middleware for multipart doesn't currently allow to mix JSON with files.
If you send a multipart request, all the "body parts" (the parameters) will be missing the "content-type" header, so the server won't be able to automatically detect the content type.

I believe I've identified the issue in [this line](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/multipart.rb#L39). The `Part` constructor accepts a forth parameter `headers` that should allow to do what you need, but we're not providing this at the moment.

Main reason for this is that mixing JSON and multipart is not really RESTful, so **if you don't have the possibility to change the server implementation**, a PR to fix this is more then welcome

However **if you CAN change the server**, you have other 2 options and I would strongly suggest to go for one of these:
1. Always force the server to parse the request body as JSON.
2. Send your files as JSON items and avoid the multipart/form-data request entirely (this works by encoding binary files in Base64). Unfortunately I only found a [java example](https://www.linkedin.com/pulse/how-transfer-binary-files-pdfimages-etc-json-rahul-budholiya) with a quick search, but I wouldn't be surprised if there's a gem for that.
3. Use a 2 phases approach (file and JSON body on separate requests) [example](https://stackoverflow.com/questions/3938569/how-do-i-upload-a-file-with-metadata-using-a-rest-web-service).

Let me know what do you end up doing

iMacTia:
[...]

PR #1014, opened 2019-08-30
Allow JSON and IO in the same payload

jeremy-israel:
## Description
This PR aims to make it possible to multipart POST an IO and a JSON payload at the same time.
Fixes https://github.com/lostisland/faraday/issues/769

## Additional Notes
Example of API which requires this :
https://developer.concur.com/api-reference/receipts/endpoints.html#endpoint-post-a-receipt

cURL data and image:

```
curl -v -k -X POST https://us.api.concursolutions.com/receipts/v4/users/{USER ID FROM YOUR ID TOKEN} \
-H "Authorization: Bearer {YOUR ACCESS TOKEN}" \
-H "Content-Type:multipart/form-data" \
-H "link: <http://schema.concursolutions.com/{VALIDATION SCHEMA FROM SCHEMA ENDPOINT}.schema.json>;rel=describedBy" \
-F "receipt=<{PATH TO YOUR RECEIPT JSON};type=application/json" \
-F "image=@{PATH TO YOUR IMAGE};type={FILE MIME TYPE OF YOUR IMAGE}"
```

### DIFF
--- a/lib/faraday/request/multipart.rb
+++ b/lib/faraday/request/multipart.rb
@@ -52,13 +52,28 @@ module Faraday
       def create_multipart(env, params)
         boundary = env.request.boundary
         parts = process_params(params) do |key, value|
-          Faraday::Parts::Part.new(boundary, key, value)
+          part(boundary, key, value)
         end
         parts << Faraday::Parts::EpiloguePart.new(boundary)
 
         body = Faraday::CompositeReadIO.new(parts)
         env.request_headers[Faraday::Env::ContentLength] = body.length.to_s
         body
+      end
+
+      def part(boundary, key, value)
+        if json?(value)
+          header = { 'Content-Type' => 'application/json' }
+          Faraday::Parts::Part.new(boundary, key, value, header)
+        else
+          Faraday::Parts::Part.new(boundary, key, value)
+        end
+      end
+
+      def json?(value)
+        JSON.parse(value)
+      rescue StandardError
+        false
       end
 
       # @return [String]
--- a/spec/faraday/request/multipart_spec.rb
+++ b/spec/faraday/request/multipart_spec.rb
@@ -68,6 +68,45 @@ RSpec.describe Faraday::Request::Multipart do
     end
   end
 
+  context 'when providing json and IO content in the same payload' do
+    let(:io_content) { StringIO.new('io-content') }
+    let(:json_content) do
+      {
+        b: 1,
+        c: 2
+      }.to_json
+    end
+
+    let(:payload) do
+      {
+        json_content: json_content,
+        io_content: Faraday::UploadIO.new(io_content, 'application/pdf')
+      }
+    end
+
+    it_behaves_like 'a multipart request'
+
+    it 'forms a multipart request' do
+      response = conn.post('/echo', payload)
+
+      boundary = parse_multipart_boundary(response.headers['Content-Type'])
+      result = parse_multipart(boundary, response.body)
+      expect(result[:errors]).to be_empty
+
+      part_json, body_json = result.part('json_content')
+      expect(part_json).to_not be_nil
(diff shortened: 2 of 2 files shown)


technoweenie:
Gah, I merged this on accident when I was trying to provide some feedback.

I would prefer supporting something like this when defining the payload on the client:

```ruby
      {
        json_content: Faraday::UploadIO.new(json_content, 'applicaton/json'),
        io_content: Faraday::UploadIO.new(io_content, 'application/pdf')
      }
```

That doesn't quite work because `json_content` is being treated as a file path because it's not an IO object. One workaround is to wrap it in a StringIO, which only fails the filename check on the json part in the added specs.


```ruby
      {
        json_content: Faraday::UploadIO.new(StringIO.new(json_content), 'applicaton/json'),
        io_content: Faraday::UploadIO.new(io_content, 'application/pdf')
      }
```

I prefer this approach, since it's not specific to json, and allows the user to opt-in (as @iMacTia said above).

technoweenie:
Oh, I should've made this more clear. I'm not suggesting that `Faraday::UploadIO.new` is the answer here. My suggestion right now is more along the lines of:

```ruby
{
  json_content: SOMETHING(json_content, 'application/json')
}
```

I'm not sure what `SOMETHING` should be yet.

jeremy-israel:
@technoweenie I'll try to make it work the way you suggest and open another PR.

technoweenie:
Sounds good. I'm re-familiarizing myself with this part of the codebase. I'm also open to new class or method names than `Faraday::UploadIO.new`. It should be possible to add something new without breaking old usage of `Faraday::UploadIO`.

technoweenie:
Okay, I've looked into this a bit more, and I think I have a better handle on what's going on. Let me know if my understanding is off.

* Faraday param values can either be a string (or something easily convertable to one, such as an integer), or the values can be a `Faraday::UploadIO` (a reference to `::UploadIO` from the `multipart-post` gem.
* These values are just [thrown into `Faraday::Parts::Part.new`](https://github.com/lostisland/faraday/blob/f4ae96ea9b7850f0c2bf03a101b0d0beda8a1049/lib/faraday/request/multipart.rb#L54-L56) with no header. Note: `Faraday::Parts::Part` is a reference to `::Parts::Part` from the `multipart-post` gem.
* `::Parts::Part.new` checks [if the value responds to `#content_type` and `#local_filename`](https://github.com/socketry/multipart-post/blob/f119fc9739735ad4126e8edeaca1f292d039ba0c/lib/parts.rb#L18-L20). If so, it returns a `::Parts::FilePart`, which [uses the `UploadIO` object to get the value's content type](https://github.com/socketry/multipart-post/blob/f119fc9739735ad4126e8edeaca1f292d039ba0c/lib/parts.rb#L75-L76).
* Any values that aren't `UploadIO` are returned as a `::Parts::ParamPart`, which [needs its headers given explicitly](https://github.com/socketry/multipart-post/blob/f119fc9739735ad4126e8edeaca1f292d039ba0c/lib/parts.rb#L52-L61).

jeremy-israel:
> Oh, I should've made this more clear. I'm not suggesting that `Faraday::UploadIO.new` is the answer here. My suggestion right now is more along the lines of:
> 
> ```ruby
> {
>   json_content: SOMETHING(json_content, 'application/json')
> }
> ```
> 
> I'm not sure what `SOMETHING` should be yet.

I gave it a thought. We could maybe create a class `Faraday::Parts::PartWithHeaders`. This class would be initialized with a string value and a headers hash.

In `Request::Multipart#create_multipart`, we could then create a `::Part` with the specified  headers from the `PartWithHeaders`.

Is that what you had in mind for the approach ? What do you think of the naming ? I also thought of `PartParams` instead.

jeremy-israel:
I opened https://github.com/lostisland/faraday/pull/1017

PR #1017, opened 2019-09-09
Create PartParams class to allow multipart posts with JSON content and file upload at the same time

jeremy-israel:
## Description
This PR aims to make it possible to multipart POST an IO and a JSON payload at the same time.
Fixes https://github.com/lostisland/faraday/issues/769

Opened after discussions in https://github.com/lostisland/faraday/pull/1014

## Additional Notes
Example of API which requires this :
https://developer.concur.com/api-reference/receipts/endpoints.html#endpoint-post-a-receipt

cURL data and image:

```
curl -v -k -X POST https://us.api.concursolutions.com/receipts/v4/users/{USER ID FROM YOUR ID TOKEN} \
-H "Authorization: Bearer {YOUR ACCESS TOKEN}" \
-H "Content-Type:multipart/form-data" \
-H "link: <http://schema.concursolutions.com/{VALIDATION SCHEMA FROM SCHEMA ENDPOINT}.schema.json>;rel=describedBy" \
-F "receipt=<{PATH TO YOUR RECEIPT JSON};type=application/json" \
-F "image=@{PATH TO YOUR IMAGE};type={FILE MIME TYPE OF YOUR IMAGE}"
```

### DIFF
--- a/lib/faraday/request/multipart.rb
+++ b/lib/faraday/request/multipart.rb
@@ -52,13 +52,21 @@ module Faraday
       def create_multipart(env, params)
         boundary = env.request.boundary
         parts = process_params(params) do |key, value|
-          Faraday::Parts::Part.new(boundary, key, value)
+          part(boundary, key, value)
         end
         parts << Faraday::Parts::EpiloguePart.new(boundary)
 
         body = Faraday::CompositeReadIO.new(parts)
         env.request_headers[Faraday::Env::ContentLength] = body.length.to_s
         body
+      end
+
+      def part(boundary, key, value)
+        if value.respond_to?(:to_part)
+          value.to_part(boundary, key)
+        else
+          Faraday::Parts::Part.new(boundary, key, value)
+        end
       end
 
       # @return [String]
--- a/spec/faraday/request/multipart_spec.rb
+++ b/spec/faraday/request/multipart_spec.rb
@@ -68,6 +68,45 @@ RSpec.describe Faraday::Request::Multipart do
     end
   end
 
+  context 'when providing json and IO content in the same payload' do
+    let(:io) { StringIO.new('io-content') }
+    let(:json) do
+      {
+        b: 1,
+        c: 2
+      }.to_json
+    end
+
+    let(:payload) do
+      {
+        json: Faraday::ParamPart.new(json, 'application/json'),
+        io: Faraday::UploadIO.new(io, 'application/pdf')
+      }
+    end
+
+    it_behaves_like 'a multipart request'
+
+    it 'forms a multipart request' do
+      response = conn.post('/echo', payload)
+
+      boundary = parse_multipart_boundary(response.headers['Content-Type'])
+      result = parse_multipart(boundary, response.body)
+      expect(result[:errors]).to be_empty
+
+      part_json, body_json = result.part('json')
+      expect(part_json).to_not be_nil
+      expect(part_json.mime).to eq('application/json')
+      expect(part_json.filename).to be_nil
+      expect(body_json).to eq(json)
+
+      part_io, body_io = result.part('io')
+      expect(part_io).to_not be_nil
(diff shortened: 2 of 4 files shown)


lib/faraday/request/multipart.rb
```diff
@@ -61,6 +61,14 @@ def create_multipart(env, params)
         body
       end
 
+      def part(boundary, key, value)
+        if value.is_a?(Faraday::Request::PartParams)
```
technoweenie: What do you think about changing this to check if `value` responds to `to_part`? It'll make this more generic, giving users the ability to add their own "Part" implementation.

lib/faraday/request/part_params.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Faraday
+  class Request
+    # Allows multipart posts while specifying headers of a part
+    class PartParams
```
technoweenie: I like the name `ParamsPart` instead, as it describes what kind of Part it is. I could even see us renaming `UploadIO` to something like `FilePart` for consistency (while keeping the old class around for backwards compatibility, of course).

spec/faraday/request/multipart_spec.rb
```diff
@@ -68,6 +68,90 @@
...
+      }.to_json
+    end
+
+    let(:payload) do
+      {
+        json_content: Faraday::Request::PartParams.new(
+          json_content,
+          'Content-Type' => 'application/json'
+        ),
+        io_content: Faraday::UploadIO.new(io_content, 'application/pdf')
```
technoweenie: A few inconsistencies stand out to me here.

1. Should the module namespace be `Faraday` or `Faraday::Request`? 
  * `Faraday::Request` is currently used by only the middleware classes themselves, like `Faraday::Request::Multipart`. It sure makes sense for `UploadIO` to live there too... I don't recall if I put it on `Faraday` because it'd be shorter for users to type, or it's part of the [multipart-post gem](https://github.com/socketry/multipart-post/blob/master/lib/composite_io.rb#L57-L108).
2. Hash vs content type argument. Since a `multipart-post` gem `ParamPart` [only uses `Content-Type` and `Content-ID`](https://github.com/socketry/multipart-post/blob/f119fc9739735ad4126e8edeaca1f292d039ba0c/lib/parts.rb#L52-L60), what do you think of a method signature like `PartParams.new(value, content_type, content_id=nil)`?

lib/faraday/request/multipart.rb
jeremy-israel: great idea, I'm buying

jeremy-israel: ✅

lib/faraday/request/part_params.rb
jeremy-israel: I didn't do anything about this, but I agree `FilePart` would be a good name.

spec/faraday/request/multipart_spec.rb
jeremy-israel: I un-namespaced the `PartParams` class and used the method signature you suggested (it makes sense !!)

lib/faraday/request/part_params.rb
jeremy-israel: I miss-read your comment. I changed the name to `ParamsPart`.

technoweenie:
Hey, after seeing the class name in the code and docs, I'm slightly bothered by the plural `ParamsPart` name. It represents a single parameter of a multipart post request. A singular name like `ParamPart` will be more consistent with the renamed `FilePart` class too.

```ruby
{ raw: Faraday::ParamPart.new(some_json, "application/json") }
```

I'm happy to make this quick change too if you don't want to bother :)

jeremy-israel:
Renamed `ParamsPart` to `ParamPart`.

PR #1021, opened 2019-09-18
Rename UploadIO -> FilePart

technoweenie:
This adds a new alias of `::UploadIO` (provided by the multipart-post gem) to `Faraday::FilePart`. This makes it consistent with `ParamPart`, just added in #1017.


technoweenie:
The [code climate failure](https://codeclimate.com/github/lostisland/faraday/pull/1021) is regarding the CompositeIO class that I copied from the multipart-post gem, with some custom tweak that Faraday needs.

PR #1022, opened 2019-09-19
read timeout request option (part 2)

technoweenie:
This extends #1003 by adding `Faraday::RequestOptions#fetch_timeout` to clean up the logic of `#configure_request` in the net/http adapter. Instead of checking for `req[:timeout]` before the more specific timeout settings in the adapter, this happens in `#fetch_timeout`. 

If this pattern is okay, I'll update the other adapters to use `#fetch_timeout`, ensuring they all properly support the new `read_timeout` setting. I'm open to other ideas, or even just different names.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -165,19 +165,16 @@ module Faraday
       end
 
       def configure_request(http, req)
-        if req[:timeout]
-          http.read_timeout = req[:timeout]
-          http.open_timeout = req[:timeout]
-          if http.respond_to?(:write_timeout=)
-            http.write_timeout = req[:timeout]
-          end
+        if sec = req.fetch_timeout(:read)
+          http.read_timeout = sec
         end
-        http.open_timeout = req[:open_timeout] if req[:open_timeout]
-        http.read_timeout = req[:read_timeout] if req[:read_timeout]
+        if sec = http.respond_to?(:write_timeout=) && req.fetch_timeout(:write)
+          http.write_timeout = sec
+        end
+        if sec = req.fetch_timeout(:open)
+          http.open_timeout = sec
+        end
 
-        if req[:write_timeout] && http.respond_to?(:write_timeout=)
-          http.write_timeout = req[:write_timeout]
-        end
         # Only set if Net::Http supports it, since Ruby 2.5.
         http.max_retries = 0 if http.respond_to?(:max_retries=)
 
--- a/spec/faraday/adapter/net_http_spec.rb
+++ b/spec/faraday/adapter/net_http_spec.rb
@@ -11,21 +11,25 @@ RSpec.describe Faraday::Adapter::NetHttp do
     let(:http) { adapter.send(:net_http_connection, url: url, request: {}) }
 
     it { expect(http.port).to eq(80) }
+
     it 'sets max_retries to 0' do
       adapter.send(:configure_request, http, {})
 
       expect(http.max_retries).to eq(0) if http.respond_to?(:max_retries=)
     end
+
     it 'supports write_timeout' do
       adapter.send(:configure_request, http, write_timeout: 10)
 
       expect(http.write_timeout).to eq(10) if http.respond_to?(:write_timeout=)
     end
+
     it 'supports open_timeout' do
       adapter.send(:configure_request, http, open_timeout: 10)
 
       expect(http.open_timeout).to eq(10)
     end
+
     it 'supports read_timeout' do
       adapter.send(:configure_request, http, read_timeout: 10)
 
--- a/spec/faraday/options/request_options_spec.rb
+++ b/spec/faraday/options/request_options_spec.rb
@@ -1,8 +1,50 @@
 # frozen_string_literal: true
 
 RSpec.describe Faraday::RequestOptions do
+  subject(:options) { Faraday::RequestOptions.new }
+
+  context '#fetch_timeout' do
+    it 'gets :read timeout' do
+      expect(options.fetch_timeout(:read)).to eq(nil)
+
+      options[:timeout] = 5
+      options[:write_timeout] = 1
+
+      expect(options.fetch_timeout(:read)).to eq(5)
+
+      options[:read_timeout] = 2
+
+      expect(options.fetch_timeout(:read)).to eq(2)
+    end
+
+    it 'gets :open timeout' do
+      expect(options.fetch_timeout(:open)).to eq(nil)
+
+      options[:timeout] = 5
+      options[:write_timeout] = 1
+
+      expect(options.fetch_timeout(:open)).to eq(5)
+
+      options[:open_timeout] = 2
+
+      expect(options.fetch_timeout(:open)).to eq(2)
+    end
+
+    it 'gets :write timeout' do
+      expect(options.fetch_timeout(:write)).to eq(nil)
+
+      options[:timeout] = 5
+      options[:read_timeout] = 1
+
+      expect(options.fetch_timeout(:write)).to eq(5)
+
+      options[:write_timeout] = 2
+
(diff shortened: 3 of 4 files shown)


lib/faraday/adapter/excon.rb
```diff
@@ -90,17 +90,17 @@ def amend_opts_with_ssl!(opts, ssl)
...
+        if (sec = request_timeout(:write, req))
+          opts[:write_timeout] = sec
+        end
 
-        open_timeout = req[:open_timeout]
-        return unless open_timeout
+        return unless (sec = request_timeout(:open, req))
 
-        opts[:connect_timeout] = open_timeout
+        opts[:connect_timeout] = sec
```
technoweenie: Rubocop forced me to turn this into a guard check ^^^

technoweenie:
I updated the EMHttp and Excon adapters with positive results and one small issue: the internal adapter code treats the `Faraday::RequestOptions` as a hash with symbol keys (`req[:timeout]` for example). 

Externally, end users should never be able to pass a plain request options hash. Faraday will handle this conversion. But, a lot of specs tend to pass in plain hashes. So, I moved `Faraday::RequestOptions#fetch_timeout(type)` to `Faraday::Adapter#request_timeout(type, req)` in b35f6a24809830d5c039725bcb5f6bff6355c646.

That will let us clean the timeout option logic a bit, while still (mostly) treating the request options as a hash. 

However, other Options classes have methods on them. Even [`RequestOptions#stream_response?`](https://github.com/lostisland/faraday/blob/b35f6a24809830d5c039725bcb5f6bff6355c646/lib/faraday/options/request_options.rb#L18-L20) exists, requiring the existential operator in places:

https://github.com/lostisland/faraday/blob/b35f6a24809830d5c039725bcb5f6bff6355c646/lib/faraday/adapter/excon.rb#L19-L24

Big Question Time: Are we cool with adding methods to the Options classes, or should we strive to move that logic to `Faraday::Adapter` or somewhere else? 

Personally, I'd like to bring `RequestOptions#fetch_timeout` back, and fix the tests to use `RequestOptions.new` instead of `{}`... but only if we're sure it's not going to cause breakage for some folks.

iMacTia:
It really depends on what meaning we give to the option classes. They are simple Structs, not fully-fledged classes. In my mind, this means they should only be used to store values and provide getter/setters methods. Any other method or logic should NOT be defined there.

That said, I guess we're a bit borderline here.
The `stream_response?` method simply asks info about one of the stored properties, without manipulating the content of the Struct. 
Same goes for `request_timeout`, although the implementation is less trivial.
Another solution would be to override the `*_timeout` getters in RequestOptions to default to `timeout`, that would probably fit better in the `Struct` definition

technoweenie:
> In my mind, this means they should only be used to store values and provide getter/setters methods. Any other method or logic should NOT be defined there.

I agree, to a point. I like the properties on some of the options classes that only operate on the config properties themselves. Examples: `ProxyOptions`, `RequestOptions`, and `SSLOptions`. I do think Env pushes that boundary, but that's on the chopping block for Faraday 2 ;)

> Another solution would be to override the *_timeout getters in RequestOptions to default to timeout, that would probably fit better in the Struct definition

Possibly, but most of the options are accessed as symbol hash keys. A `case` block in `def [](key)` could do this:

```ruby
module Faraday
  class RequestOptions # ...
    def [](key)
      key_sym = key.to_sym
      case key_sym
      when :read_timeout then super(:timeout) || super(key)
      else super(key)
      end
    end
```

Currently I'm leaning towards sticking with `Faraday::Adapter#request_timeout`, rather than putting more code into the Options classes. I think a method call in the Adapter is more obvious that something custom is happening, than a basic `req[:read_timeout]` hash access.

iMacTia:
Right, considering the access to properties through symbols, it seems like `Faraday::Adapter#request_timeout` is the most logical solution 😄

lib/faraday/adapter.rb
```diff
@@ -65,5 +65,31 @@ def save_response(env, status, body, headers = nil, reason_phrase = nil)
...
+    # @param type [Symbol] Describes which timeout setting to get: :read,
+    #                      :write, or :open.
+    # @param options [Hash] Hash containing Symbol keys like :timeout,
+    #                       :read_timeout, :write_timeout, :open_timeout, or
+    #                       :timeout
+    #
+    # @return [Integer, nil] Timeout duration in seconds, or nil if no timeout
+    #                        has been set.
+    def request_timeout(type, options)
+      key = TIMEOUT_KEYS[type]
```
olleolleolle: Could use a `key = TIMEOUT_KEYS.fetch(type) do ... end` and raise the error in there?

lib/faraday/adapter/em_http.rb
```diff
@@ -70,10 +70,9 @@ def configure_ssl(options, env)
 
         # Reads out timeout settings from env into options
         def configure_timeout(options, env)
-          timeout, open_timeout = request_options(env)
-                                  .values_at(:timeout, :open_timeout)
-          options[:connect_timeout] = options[:inactivity_timeout] = timeout
-          options[:connect_timeout] = open_timeout if open_timeout
+          req = request_options(env)
+          options[:inactivity_timeout] = request_timeout(:read, req)
```
olleolleolle: This is a good thing, a boundary between EM and Faraday.

And the granularity has increased!

lib/faraday/adapter/net_http.rb
```diff
@@ -165,17 +165,17 @@ def configure_ssl(http, ssl)
...
-        if req[:timeout]
-          http.read_timeout = req[:timeout]
-          http.open_timeout = req[:timeout]
-          if http.respond_to?(:write_timeout=)
-            http.write_timeout = req[:timeout]
-          end
+        if (sec = request_timeout(:read, req))
+          http.read_timeout = sec
+        end
+        if (sec = http.respond_to?(:write_timeout=) &&
```
olleolleolle:
```suggestion

        if (sec = http.respond_to?(:write_timeout=) &&
```

lib/faraday/adapter/net_http.rb
```diff
@@ -165,17 +165,17 @@ def configure_ssl(http, ssl)
...
+                  request_timeout(:write, req))
+          http.write_timeout = sec
         end
-        http.open_timeout = req[:open_timeout] if req[:open_timeout]
-        if req[:write_timeout] && http.respond_to?(:write_timeout=)
-          http.write_timeout = req[:write_timeout]
+        if (sec = request_timeout(:open, req))
```
olleolleolle:
```suggestion

        if (sec = request_timeout(:open, req))
```

spec/faraday/adapter/net_http_spec.rb
```diff
@@ -11,17 +11,31 @@
     let(:http) { adapter.send(:net_http_connection, url: url, request: {}) }
 
     it { expect(http.port).to eq(80) }
+
     it 'sets max_retries to 0' do
       adapter.send(:configure_request, http, {})
 
       expect(http.max_retries).to eq(0) if http.respond_to?(:max_retries=)
     end
+
```
olleolleolle: Thanks for these breathing-room newlines!

spec/faraday/options/request_options_spec.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 RSpec.describe Faraday::RequestOptions do
+  subject(:options) { Faraday::RequestOptions.new }
```
olleolleolle: Neat!

lib/faraday/adapter.rb
olleolleolle: (We can do that in a follow-up PR, no biggie.)

technoweenie:
Thanks for the reviews :wave:

PR #1030, opened 2019-09-27
Restore backward-compatibility layer for errors

michaelherold:
Faraday has been maintaining a backward-compatibility layer for errors
in the `Faraday::Error::` "namespace". These were removed with v0.16.0,
which caused issues for dependents that use semantic versioning
constraints.

It would be best if these were deprecated with a warning message and
then dropped in v1.0 in order to make sure transitive dependencies do
not break.

See octokit/octokit.rb#1154

### DIFF
--- a/lib/faraday/deprecated_constant.rb
+++ b/lib/faraday/deprecated_constant.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Faraday
+  # Allows for the deprecation of constants between versions of Faraday
+  #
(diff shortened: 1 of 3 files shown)


michaelherold:
I rebased this on top of the 0.16.x branch because it should only go on those releases, not on the 1.0 release. The linter is now complaining about missing documentation for something that I didn't touch - I figure I shouldn't try to add those.

michaelherold:
I added a very simple constant deprecation proxy for these backward-compatible error classes. Feel free to push back on the implementation!

Issue #1038, opened 2019-09-29
Breaking changes for fastlane in 0.16.2


janpio:
## Basic Info

* Faraday Version: 0.16.2
* Ruby Version: 
	- Windows: ruby 2.4.6p354 (2019-04-01 revision 67394) [x64-mingw32]
	- Mac: Ruby 2.5

## Issue description
Even with faraday `0.16.2` and the fixes of @BobbyMcWho there are test failures for fastlane:

```
Failures:
  1) Spaceship::Portal::Persons should change role
     Failure/Error: expect { person.change_role("member") }.to_not(raise_error)
       expected no Exception, got #<RuntimeError: can't modify frozen String> with backtrace:
         # ./spaceship/lib/spaceship/client.rb:811:in `force_encoding'
         # ./spaceship/lib/spaceship/client.rb:811:in `log_response'
         # ./spaceship/lib/spaceship/client.rb:846:in `block in send_request'
         # ./spaceship/lib/spaceship/client.rb:620:in `with_retry'
         # ./spaceship/lib/spaceship/client.rb:844:in `send_request'
         # ./spaceship/lib/spaceship/client.rb:688:in `request'
         # ./spaceship/lib/spaceship/portal/portal_client.rb:449:in `team_set_role'
         # ./spaceship/lib/spaceship/portal/person.rb:45:in `change_role'
         # ./spaceship/spec/portal/person_spec.rb:58:in `block (3 levels) in <top (required)>'
         # ./spaceship/spec/portal/person_spec.rb:58:in `block (2 levels) in <top (required)>'
     # ./spaceship/spec/portal/person_spec.rb:58:in `block (2 levels) in <top (required)>'
  2) Spaceship::Tunes::IAPList IAPList can delete
     Failure/Error: deleted = app.in_app_purchases.find("go.find.me").delete!
     RuntimeError:
       can't modify frozen String
     # ./spaceship/lib/spaceship/client.rb:811:in `force_encoding'
     # ./spaceship/lib/spaceship/client.rb:811:in `log_response'
     # ./spaceship/lib/spaceship/client.rb:846:in `block in send_request'
     # ./spaceship/lib/spaceship/client.rb:620:in `with_retry'
     # ./spaceship/lib/spaceship/client.rb:844:in `send_request'
     # ./spaceship/lib/spaceship/client.rb:688:in `request'
     # ./spaceship/lib/spaceship/tunes/tunes_client.rb:1235:in `delete_iap!'
[...]

PR #1039, opened 2019-09-30
Fix instances of frozen empty string literals

BobbyMcWho:
When the frozen string literals magic comments were added, not all empty strings were made mutable. This causes issues down the line like the ones being seen in #1038 for fastlane/fastlane

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -59,7 +59,7 @@ module Faraday
           end
 
           save_response(env, http_response.code.to_i,
-                        http_response.body || '', nil,
+                        http_response.body || +'', nil,
                         http_response.message) do |response_headers|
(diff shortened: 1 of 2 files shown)


janpio:
At fastlane we have confirmation that merging these changes would probably fix our problems: https://github.com/fastlane/fastlane/issues/15397#issuecomment-536415270 So we would really appreciate a 0.16.3 with this PR ❤️

janpio:
I actually just relayed the information and did nothing else - @blueberry6401 deserves the credit here <3

BobbyMcWho:
This hasn't been released to rubygems yet, not sure if best practice is to include the version bump in the PR or let the maintainers do that post merge

olleolleolle:
(Regular practice is that the releasing person does any version file change.)

Issue #1028, opened 2019-09-27
Nil status returned from specific URLs


jonnyom:
## Basic Info

* Faraday Version: 0.15.3
* Ruby Version: 2.5.5

## Issue description
Occasionally when POSTing to 3rd party URLs from our webhook service (which is using Faraday), we see `nil` statuses in our responses. This doesn't seem like expected behaviour.

## Steps to reproduce
1. Create a connection with the following configuration:
```ruby
headers = {
              "Content-Type" => "application/json",
              "Accept" => "application/json",
            }
connection = Faraday.new(ssl: { verify: true }) do |c|
              c.adapter :httpclient
              c.headers = headers
              c.options.timeout = 5
              c.options.open_timeout = 1
            end
```
2. Make a POST request to the following URL `https://intercomwebhook.gabrielramos.tk/reassign`
```ruby
connection.post(url, { test: "Testing" }.to_json)
```
3. Check the returned status. When I made did this this was the resulting Response object
```ruby
con.post(url, {hello: "hi"}.to_json)
=> #<Faraday::Response:0x0000557802a9d420
 @env=
  #<struct Faraday::Env
   method=:post,
   body=
[...]

Issue #953, opened 2019-03-20
v2 Wishlist


technoweenie:
I want to start collecting a wishlist for what we want to change in Faraday v2. Nothing here is concrete yet.

* Use SNAKE_CASE constants consistently
* Kill `Faraday::Utils`
* `Faraday::Connection` => `Faraday::Client`
* Remove Options structs in favor of properties on Client/Request/Response
* Remove env in favor of request/response object properties
* Kill adapter/middleware autoloading in favor of good ole ruby
* Reorganize param encoders: `lib/faraday/params_encoders/[nested, flat]` ?
* Reorganize middleware `lib/faraday/middleware/*`
* Extract adapters as gems
* Streaming by default, while providing easy access to cached string response bodies

grosser:
- make net-http-persistent less hacky by keeping a connection object around so we don't need a global cache
 - make net-http-persistent not use the gem ... 90% of it is managing net-http which we already do, so the only thing we need is the `rescue  + reopen` logic which is a few lines of code and would remove coupling/translation logic for the gem see https://github.com/drbrain/net-http-persistent/pull/100

technoweenie:
Thanks, those are great suggestions!

> make net-http-persistent less hacky by keeping a connection object around so we don't need a global cache

Ah yes, another reason why Faraday's implementation of Rack semantics for the adapter and middleware classes needs to go. If the current adapter was a long lived `Faraday::Connection#adapter` property, the `net-http` adapter could hold on to the connection object. I just added "Revisit adapter/middleware internal API (drop Rack adapter semantics)" to the wishlist to support this.

> make net-http-persistent not use the gem

I'm on board. Thanks for the pointer to the PR.

> allow using net-http-pipeline

Faraday does support parallel requests, but I'm not sure offhand if we could use `net-http-pipeline` to implement them for `net-http`. I've added "Revisit pipelining or parallel requests (net-http-pipeline, typhoeus)" to the wishlist.

grosser:
For pipeline: I decided to not use that in my projects because it means
rewriting most of the handler logic, so low prio for me

iMacTia:
@grosser you might have heard we're now in the process of pushing adapters and middleware out of Faraday.
We have already done a lot of work to make this as easy as possible, including [exporting tests]() and providing a few examples ([faraday-net_http]() [faraday-http]())

I was wondering, given your past contributions, if you'd like to take ownership of net_http_persistent?
[...]

PR #1042, opened 2019-09-30
Raise exception on nil status

jonnyom:
## Description
If a host responds badly to HTTP requests it's possible for the HTTP status to be returned `nil`.
It makes sense to raise an exception here that can be rescued from and handled appropriately instead of silently allowing them through.
Fixes https://github.com/lostisland/faraday/issues/1028

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -85,7 +85,7 @@ module Faraday
 
   # Raised by Faraday::Response::RaiseError in case of a nil status in response.
   class NilStatusError < ServerError
-    def initialize(response = nil)
+    def initialize(_exc, response: nil)
       message = 'http status could not be derived from the server response'
       super(message, response)
     end
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -83,6 +83,14 @@ module Faraday
     end
   end
 
+  # Raised by Faraday::Response::RaiseError in case of a nil status in response.
+  class NilStatusError < ServerError
+    def initialize(response = nil)
+      message = 'http status could not be derived from the server response'
+      super(message, response)
+    end
+  end
+
   # A unified error for failed connections.
   class ConnectionFailed < Error
   end
--- a/lib/faraday/response/raise_error.rb
+++ b/lib/faraday/response/raise_error.rb
@@ -29,7 +29,7 @@ module Faraday
         when 422
           raise Faraday::UnprocessableEntityError, response_values(env)
         when nil
-          raise Faraday::NilStatusError, response_values(env)
+          raise Faraday::NilStatusError, response: response_values(env)
         when ClientErrorStatuses
           raise Faraday::ClientError, response_values(env)
         when ServerErrorStatuses
--- a/lib/faraday/response/raise_error.rb
+++ b/lib/faraday/response/raise_error.rb
@@ -28,6 +28,8 @@ module Faraday
           raise Faraday::ConflictError, response_values(env)
         when 422
           raise Faraday::UnprocessableEntityError, response_values(env)
+        when nil
+          raise Faraday::NilStatusError, response_values(env)
         when ClientErrorStatuses
           raise Faraday::ClientError, response_values(env)
         when ServerErrorStatuses
--- a/spec/faraday/response/raise_error_spec.rb
+++ b/spec/faraday/response/raise_error_spec.rb
@@ -14,6 +14,7 @@ RSpec.describe Faraday::Response::RaiseError do
         stub.get('conflict') { [409, { 'X-Reason' => 'because' }, 'keep looking'] }
         stub.get('unprocessable-entity') { [422, { 'X-Reason' => 'because' }, 'keep looking'] }
         stub.get('4xx') { [499, { 'X-Reason' => 'because' }, 'keep looking'] }
+        stub.get('nil-status') { [nil, { 'X-Reason' => 'bailout' }, 'fail'] }
         stub.get('server-error') { [500, { 'X-Error' => 'bailout' }, 'fail'] }
       end
     end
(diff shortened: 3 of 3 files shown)


olleolleolle:
This got me curious about the case where the response code is numeric but in an invalid range. 

(I'm out traveling, so this happens from a phone.)

jonnyom:
That's a good point. A base case here for any invalid status could be useful

lib/faraday/error.rb
```diff
@@ -13,7 +13,10 @@ def initialize(exc, response = nil)
         super(exc.message)
         @wrapped_exception = exc
       elsif exc.respond_to?(:each_key)
-        super("the server responded with status #{exc[:status]}")
+        nil_status_message = 'the server responded with a nil status'
```
BobbyMcWho: Does it make sense to say the server responded with a nil status if the server did not send a status Header at all?

jonnyom: That's a fair point, maybe a more explicit message, something like `status couldn't be derived from server response`?

lib/faraday/error.rb
```diff
@@ -69,6 +72,10 @@ class ConflictError < ClientError
   class UnprocessableEntityError < ClientError
   end
 
+  # Raised by Faraday::Response::RaiseError in case of a nil status in response.
+  class NilStatusError < ClientError
```
BobbyMcWho:
```suggestion
  class NilStatusError < ServerError
```

lib/faraday/error.rb
BobbyMcWho: I like it, I'd say move it down into the error itself as well rather than in this base error class

lib/faraday/error.rb
BobbyMcWho:
```ruby
    def initialize(exc = nil, response = nil)
      message = 'http status could not be derived from the server response'
      super(message, response)
    end
  end
```

lib/faraday/response/raise_error.rb
```diff
@@ -28,6 +28,8 @@ def on_complete(env)
           raise Faraday::ConflictError, response_values(env)
         when 422
           raise Faraday::UnprocessableEntityError, response_values(env)
+        when nil
+          raise Faraday::NilStatusError, response_values(env)
```
BobbyMcWho: If you make the suggestions I made above, you'll need to change this line as well 
`raise Faraday::NilStatusError, nil, response_values(env)`

jonnyom: Sounds good 👍 I'll get to this later today

jonnyom: So passing `nil` to this actually caused a bunch of issues for me when setting the backtrace. So I've opted to not pass it at all. Let me know if there's a better way around this.

jonnyom:
This is failing lint on an unrelated file 
```shell
Inspecting 100 files
.........W..........................................................................................

Offenses:

lib/faraday/adapter/em_http_ssl_patch.rb:62:34: W: Lint/SendWithMixinArgument: Use include EmHttpSslPatch instead of send(:include, EmHttpSslPatch).
EventMachine::HttpStubConnection.send(:include, EmHttpSslPatch)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I don't have permissions to retrigger the build, but this passes locally without JUnit, but once I've installed it it fails

olleolleolle:
I'd appreciate a separate PR for the linting fix.

BobbyMcWho:
I had implemented it in #1043, I can open a new PR to target it alone

lib/faraday/response/raise_error.rb
BobbyMcWho: 🤦‍♂  silly me, we're not using Faraday::NilStatusError.new so the backtrace is implicit

jonnyom:
Cool! I'm happy to roll it into this either @BobbyMcWho but its own PR would be cleaner I guess

BobbyMcWho:
@olleolleolle merged it, @jonnyom you should be able to `git pull [your lostisland remote] 0.16.x --rebase` and be good to (force) push back up to your branch to get that change

lib/faraday/error.rb
```diff
@@ -83,6 +83,14 @@ def initialize(exc = 'timeout', response = nil)
     end
   end
 
+  # Raised by Faraday::Response::RaiseError in case of a nil status in response.
+  class NilStatusError < ServerError
+    def initialize(response = nil)
```
BobbyMcWho: I submitted a PR to your fork, this will need two params to handle the stack trace that's sent

jonnyom: I saw that, just lifted what you had and popped it in here, easier than merging things into the fork

jonnyom:
@olleolleolle @BobbyMcWho just wanted to give this a bump

lib/faraday/response/raise_error.rb
```diff
@@ -28,6 +28,8 @@ def on_complete(env)
           raise Faraday::ConflictError, response_values(env)
         when 422
           raise Faraday::UnprocessableEntityError, response_values(env)
+        when nil
+          raise Faraday::NilStatusError, response: response_values(env)
```
technoweenie: 1. I'd rather have the `nil` check at the bottom, since it's the least likely to occur than any numerical status. Pretty minor issue, so I'm not sure if it's worth the effort :)
2. I like passing `response` as a named variable, rather than positional. I've added a note to the v2 wishlist (#953) to make this consistent with the other error types.

jonnyom:
@technoweenie if you want to refactor it then maybe we should just merge this and you can make that refactor including these changes?

(I also wasn't sure where the documentation was, which is why I didn't add it)

Thanks!

technoweenie:
Thanks @jonnyom! I'll take care of the docs.

PR #1043, opened 2019-09-30
Update to a more extensible deprecation method.

BobbyMcWho:
The previous implementation would not warn about deprecation upon being inherited, and warned in stdout. Upon initialization or Inheritance, this implementation warns in stderr similar to Gem::Deprecate, but gives a semver instead of a date like the default Gem::Deprecate does.


lib/faraday/adapter/em_http_ssl_patch.rb
```diff
@@ -59,4 +59,4 @@ def certificate_store
   end
 end
 
-EventMachine::HttpStubConnection.send(:include, EmHttpSslPatch)
+EventMachine::HttpStubConnection.include(EmHttpSslPatch)
```
BobbyMcWho: Rubocop made me do it 🚓 🤖

spec/faraday/error_spec.rb
```diff
@@ -43,12 +43,26 @@
...
+        expect(error_message).to match(
+          Regexp.new(
+            'NOTE: Faraday::Error::ClientError.new is deprecated; '\
+            'use Faraday::ClientError.new instead. It will be removed in or after version 1.0'
+          )
+        )
+      end
+
+      it 'does not raise an error for inherited error-namespaced classes but prints an error message' do
+        error_message, = with_warn_squelching { class E < Faraday::Error::ClientError; end }
```
BobbyMcWho: this looks like a typo, but it's not. Rubocop likes `a, = [:foo, :bar]` better than `a, _ = [:foo, :bar]` 🤷‍♂

technoweenie:
FYI: This closed because I killed the `0.16.x` branch. I've opened #1052 to create a 0.17.x release branch and port this PR over.

PR #1051, opened 2019-10-14
First stab at GitHub Actions

iMacTia:
## Description
First configuration of a CI workflow using GitHub Actions that runs Rubocop, Yard-Junk and Rspec.

## Todos
- [x] Rubocop
- [x] Yard-Junk
- [x] RSpec
- [x] Rubygems deployment
- [x] CodeClimate


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Ruby 2.6
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6.x
+    - name: Rubocop
+      run: |
+        gem install rubocop rubocop-junit_formatter rubocop-performance --no-document
+        rubocop --require rubocop/formatter/junit_formatter \
```
olleolleolle:
```suggestion
        rubocop \
```

iMacTia:
@olleolleolle I was wondering what that was, I thought it was related to CodeClimate 😅 
Thanks for clarifying!

iMacTia:
Looks good now but I had to explicitly install `libcurl`, which should not be necessary.
I've opened an issue [here](https://github.com/actions/virtual-environments/issues/37) to shed some light on it

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Set up Ruby 2.6
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.6.x
+    - name: Rubocop
+      run: |
+        gem install rubocop rubocop-performance --no-document
+        rubocop --require rubocop-performance \
+                --format progress \
+                --out ~/test-results/linting/rubocop.xml
```
olleolleolle: We can stop creating the output as a file. We are not using that.

So, remove the `--out ...` option.

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![Gem Version](https://badge.fury.io/rb/faraday.svg)](https://rubygems.org/gems/faraday)
 [![CircleCI](https://circleci.com/gh/lostisland/faraday/tree/master.svg?style=svg)](https://circleci.com/gh/lostisland/faraday/tree/master)
```
olleolleolle: We can drop the CircleCI badge in the same go.

```suggestion
```

.github/workflows/ci.yml
```diff
@@ -15,7 +15,6 @@ jobs:
         gem install rubocop rubocop-performance --no-document
         rubocop --require rubocop-performance \
                 --format progress \
-                --out ~/test-results/linting/rubocop.xml
```
olleolleolle: Oh, the line continuation line on the line before should also be removed.

olleolleolle:
The feedback runs are so fast.

.github/workflows/ci.yml
olleolleolle: It still _worked_, though.

iMacTia:
Yeah it looks much faster than any external CI 😮!
When I opened the PR, it already started building as well so it was a better of seconds before I could see the result!

.github/workflows/ci.yml
olleolleolle: Perhaps we can reduce all the way down to `rubocop --require rubocop-performance`

Issue #1052, opened 2019-10-15
Update 0.1x branch


technoweenie:
We need a stable branch for 0.17.x for two reasons:

1. Ship important bug fix releases.
2. Warn 0.17.x users of the deprecations facing them in 1.0.

* [x] Create `0.1x` branch from the 0.17.0 tag
* [x] Get CI working #1015 #1027 (#1055)
* [ ] Error classes: Port #841, #1042, and error class changes from #1054 (#1064)
* [x] Port #1043 (#1054)

PR #1054, opened 2019-10-17
Create faraday deprecation class

BobbyMcWho:
## Description
This implementation warns in stderr similar to Gem::Deprecate, but gives a semver instead of a date like the default Gem::Deprecate does. 

```ruby
require 'faraday/deprecate'

class SomeFaradayClass
  # deprecate instance method
  extend Faraday::Deprecate

  def old_method; end
  deprecate :old_method, :new_method, '1.0.0'

  def new_method; end

  class << self
    # when deprecating a class method, you must extend from within the eigenclass
    extend Faraday::Deprecate

    def old_class_method; end
    deprecate :old_class_method, :new_class_method, '1.0.0'

    def new_class_method; end
  end
end
```

Relates to #1052 #1043


lib/faraday/deprecate.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+module Faraday
+  # @param new_klass [Class] new Klass to use
+  #
+  # @return [Class] A modified version of new_klass that warns on
+  #   usage about deprecation.
+  # @see Faraday::Deprecate
+  module DeprecatedClass
+    def self.proxy_class(new_klass)
```
technoweenie: I think this should accept a `version` argument so we can proxy deprecated classes beyond 1.0. Probably not necessary for this PR on the `0.1x` branch, but definitely once this gets ported to `master`.

PR #1059, opened 2019-10-17
0.1x deprecate fixes

technoweenie:
This fixes a few issues with #1054

First: You can't rescue a current Faraday exception (`Faraday::ClientError`, for example) with its deprecated error class (`Faraday::Error::ClientError`). This would let us remove all `Faraday::Error::*` references for Faraday `v0.17.1`. We should update them too, so that users don't get a bunch of deprecation warnings they can't fix. Even if that's too risky of a change, I still think this behavior is correct and worth including.

Second, the regex for the class name changed so it could grab the class name of a root level class. Only came up because some of the test constants have no namespace.

```diff
-/^#<Class:(\w*::\w*)>$/
+/^#<Class:(\w{1}[\w:]*)>$/
```

Finally, I copied some of the tests in `spec/faraday/error_spec.rb` to `spec/faraday/deprecate_spec.rb`, so that those things are still tested once the `Faraday::Error::*` stuff is all removed for good. This raised some problems with the `CustomError` and `E` classes created in `error_spec.rb`, so I changed those to use `Class.new` instead.

@BobbyMcWho: I'd really appreciate your thoughts on this :)

### DIFF
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -12,7 +12,7 @@ module Faraday
         class << k
           extend Faraday::Deprecate
           # Make this more human readable than #<Class:Faraday::ClientError>
-          klass_name = superclass.to_s[/^#<Class:(\w{1}[\w:]*)>$/, 1]
+          klass_name = superclass.to_s[/^#<Class:([\w:]+)>$/, 1]
           deprecate :new, "#{klass_name}.new", '1.0'
           deprecate :inherited, klass_name, '1.0'
 
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -12,7 +12,7 @@ module Faraday
         class << k
           extend Faraday::Deprecate
           # Make this more human readable than #<Class:Faraday::ClientError>
-          klass_name = superclass.to_s[/^#<Class:(\w*::\w*)>$/, 1]
+          klass_name = superclass.to_s[/^#<Class:(\w{1}[\w:]*)>$/, 1]
           deprecate :new, "#{klass_name}.new", '1.0'
           deprecate :inherited, klass_name, '1.0'
         end
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -15,6 +15,10 @@ module Faraday
           klass_name = superclass.to_s[/^#<Class:(\w{1}[\w:]*)>$/, 1]
           deprecate :new, "#{klass_name}.new", '1.0'
           deprecate :inherited, klass_name, '1.0'
+
+          def ===(other)
+            superclass === other || super
+          end
         end
       end
     end
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -8,8 +8,8 @@ module Faraday
   # @see Faraday::Deprecate
   module DeprecatedClass
     def self.proxy_class(new_klass)
-      Class.new(new_klass).tap do |k|
-        class << k
+      Class.new(new_klass) do
+        class << self
           extend Faraday::Deprecate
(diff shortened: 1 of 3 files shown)


lib/faraday/deprecate.rb
```diff
@@ -12,9 +12,13 @@ def self.proxy_class(new_klass)
         class << k
           extend Faraday::Deprecate
           # Make this more human readable than #<Class:Faraday::ClientError>
-          klass_name = superclass.to_s[/^#<Class:(\w*::\w*)>$/, 1]
+          klass_name = superclass.to_s[/^#<Class:(\w{1}[\w:]*)>$/, 1]
```
BobbyMcWho:
```suggestion
          klass_name = superclass.to_s[/^#<Class:([\w:]+)>$/, 1]
```

Can you think of a case this wouldn't work for?

technoweenie: Yeah, that looks fine. I was thinking the name should start with a letter, but that's silly considering we have no control over what ruby decides the superclass name will be.

spec/faraday/error_spec.rb
```diff
@@ -67,7 +67,45 @@
...
+        rescue Faraday::Error::ClientError
+        rescue Faraday::Error => exc
+          fail "rescued #{exc.class.name} instead"
+        end
+      end
+
+      it 'allows rescuing of a deprecated error with a current error' do
+        begin
+          raise Faraday::Error::ClientError, nil
+        rescue Faraday::ClientError
```
olleolleolle: Can we find a way to output a message of success when we pass here?

technoweenie: What is a "not-fail" test case? Is that something you see in the results? What would a message of success here look like?

technoweenie: I changed those examples to use `expect { raise ... }.to raise_error(...)` instead. Much shorter, and still tests the correct thing. If I remove the `#===` implementation on the inherited meta class that `Faraday::DeprecatedClass.proxy_class` creates, I get these failures:

```
  1) Faraday::DeprecatedClass allows rescuing of a current error with a deprecated error
     Failure/Error: expect { raise SampleClass, nil }.to raise_error(SampleDeprecatedClass)

       expected SampleDeprecatedClass, got #<SampleClass: SampleClass> with backtrace:
         # ./spec/faraday/deprecate_spec.rb:44:in `block (3 levels) in <top (required)>'
         # ./spec/faraday/deprecate_spec.rb:44:in `block (2 levels) in <top (required)>'
     # ./spec/faraday/deprecate_spec.rb:44:in `block (2 levels) in <top (required)>'

  2) Faraday::ClientError.initialize maintains backward-compatibility until 1.0 allows rescuing of a current error with a deprecated error
     Failure/Error: expect { raise Faraday::ClientError, nil }.to raise_error(Faraday::Error::ClientError)

       expected Faraday::Error::ClientError, got #<Faraday::ClientError Faraday::ClientError> with backtrace:
         # ./spec/faraday/error_spec.rb:76:in `block (5 levels) in <top (required)>'
         # ./spec/faraday/error_spec.rb:76:in `block (4 levels) in <top (required)>'
     # ./spec/faraday/error_spec.rb:76:in `block (4 levels) in <top (required)>'
```

olleolleolle: I was thinking of "if we  get here, `assert_equal(true, true, "it works")`" kind of thing.

technoweenie: Ohh I see. I think the `expect` syntax works too. I don't think an `expect(true).to be_true` underneath is necessary.

olleolleolle:
The raise_error matchers worked out great!

BobbyMcWho:
Awesome, thanks for catching those cases I missed originally

PR #1062, opened 2019-10-19
Proxy selector

technoweenie:
I started poking at the proxy code, and then decided trying to implement the [`net/http/httpproxy` package in Go](https://godoc.org/golang.org/x/net/http/httpproxy). Basically, I thought what it'd be like to replace 14 lines of code with my own version in 326 lines (plus docs):

https://github.com/lostisland/faraday/blob/e4b4b973073b8bea739eefe41efc0647ea1f7469/lib/faraday/connection.rb#L587-L600

Three benefits:

1. `no_proxy` is correctly handled in cases where `URI#find_proxy` don't exist. I don't know what versions are affected, so this may be a moot point.
2. Potential speed improvements on repeat `#proxy_for` calls, especially with lots of `no_proxy` entries.
3. Actually support `https_proxy`

Is it worth it? I'm not really sure. But it was fun to write :)

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -5,6 +5,7 @@ require 'set'
 require 'forwardable'
 require 'faraday/middleware_registry'
 require 'faraday/dependency_loader'
+require 'faraday/proxy_selector'
 
 # This is the main namespace for Faraday.
 #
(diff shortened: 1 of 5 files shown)


technoweenie:
I listed the 3 benefits in the original PR message above. Basically, the code more correctly implements the environment vars, and it should be more efficient in long running scripts with the way it checks uri hosts against `no_proxy`. 

I'm okay leaving this PR here until we're sure we want it. I extracted #1063 out of it too.

PR #1063, opened 2019-10-19
remove temp_proxy and improve proxy tests

technoweenie:
This removes `@temp_proxy` and updates the tests accordingly. Instead of checking that value, the tests now check `req.options.proxy`:

```diff
-          expect(conn.instance_variable_get('@temp_proxy').host).to eq('proxy.com')
-          conn.get('http://example.com')
-          expect(conn.instance_variable_get('@temp_proxy')).to be_nil
+          expect(conn.proxy.uri.host).to eq('proxy.com')
+
+          conn.get('http://example.com') do |req|
+            expect(req.options.proxy).to be_nil
+          end
```

### DIFF
--- a/spec/faraday/connection_spec.rb
+++ b/spec/faraday/connection_spec.rb
@@ -330,21 +330,21 @@ RSpec.describe Faraday::Connection do
 
   describe 'proxy support' do
     it 'accepts string' do
-      with_env 'http_proxy' => 'http://proxy.com:80' do
+      with_env 'http_proxy' => 'http://env-proxy.com:80' do
         conn.proxy = 'http://proxy.com'
         expect(conn.proxy.host).to eq('proxy.com')
       end
     end
 
     it 'accepts uri' do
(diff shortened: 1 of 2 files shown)


spec/faraday/connection_spec.rb
```diff
@@ -453,8 +453,40 @@
       expect { conn.proxy = { uri: :bad_uri, user: 'rick' } }.to raise_error(ArgumentError)
     end
 
-    it 'gives priority to manually set proxy' do
+    it 'uses env http_proxy' do
+      with_env 'http_proxy' => 'http://proxy.com' do
```
olleolleolle: This one does not call itself env-proxy in the domain name. Is there a difference here?

technoweenie: In the other tests, the `env-proxy.com` domain name helps distinguish the `http_proxy` setting from the `conn.proxy` setting that should be overshadowing it. 

This test is simply reading from the env, so there's no need to change it.

Issue #241, opened 2013-02-03
Why is there no close method?


minad:
Hi,

I am missing a close method in faraday. Assuming that you use persistent connections it would be nice to have the possibility to explicitly close the connection. If you perform a request afterwards the connection could be transparently reopened.

The middlewares just have to pass through the close calls.

I don't think you net-http-persistent adapter works correctly this way, if you always recreate the connection object.

l8on:
I've been looking around for a way to close connections as well.
I understand the difficulty of providing a consistent interface as each library has a different method (`shutdown` for net-http-persistent, `finish` for net-http, `close` for `excon`) for closing connections after they are established.
However, it would be a big win to be able to call `close` on `Faraday::Connection` to actively close the connection for all supported adapters and middleware.

technoweenie:
This won't happen in Faraday most likely. #454 explains the decision to start focusing on Hurley (essentially a rewrite). The problem is that your `Faraday::Connection` doesn't know what middleware is the actual connection. It's just a stack of procs. Hurley solves this by separating the connection adapter from the stack of callbacks. However, Hurley's current interface doesn't support the notion of a close operation either :) But due to its design, it's much easier to add it to Hurley, than Faraday. I'll try to look into this soon, and ping you on the PR if you're interested.

l8on:
Thanks @technoweenie that would be great and that explanation makes sense.

dgolombek:
From my understanding, Hurley is dead -- any thoughts on re-opening this request if that's true? We'd love to have a close method, since we're running into a lot of "Net::HTTP::Persistent::Error: too many connection resets (due to Connection reset by peer - Errno::ECONNRESET)" errors and would like to try force-closing the connection afterwards, but this issue makes that challenging.

iMacTia:
Hi all,

I would be happy to bring this back to life for a future Faraday release.
I personally like the idea of having a method in all Middlewares that could be called when the connection is closed (some sort of callback, if you want).

Happy to discuss about it here if anyone have some time to work on it 👍

ioquatix:
>  The problem is that your Faraday::Connection doesn't know what middleware is the actual connection.

We do this in `async-http`, and we just pass the `#close` method thorough all the middlewares so they can do their own work, e.g. releasing caches, whatever.

ioquatix:
Please make this part of the 1.0 release. Even if it's not implemented in all middleware.

[...]

PR #1069, opened 2019-11-01
Add default implementation of `Middleware#close`.

ioquatix:
I am waiting for airplane so this is brief just to start the conversation around how this should work.

https://github.com/lostisland/faraday/issues/241

### DIFF
--- a/lib/faraday/middleware.rb
+++ b/lib/faraday/middleware.rb
@@ -11,7 +11,11 @@ module Faraday
     end
 
     def close
-      @app&.close
+      if @app.respond_to?(:close)
+        @app.close
+      else
+        warn "#{@app} does not implement \#close!"
+      end
     end
   end
 end
--- a/lib/faraday/middleware.rb
+++ b/lib/faraday/middleware.rb
@@ -9,5 +9,9 @@ module Faraday
     def initialize(app = nil)
       @app = app
     end
+
+    def close
+      @app&.close
+    end
   end
 end
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -13,7 +13,7 @@ Metrics/AbcSize:
 # Offense count: 4
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 233
+  Max: 234
 
 # Offense count: 17
 Metrics/CyclomaticComplexity:
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -117,6 +117,13 @@ module Faraday
 
     def_delegators :builder, :build, :use, :request, :response, :adapter, :app
 
+    # Closes the underlying resources and/or connections. In the case of
+    # persistent connections, this closes all currently open connections
+    # but does not prevent new connections from being made.
+    def close
+      app.close
+    end
+
     # @!method get(url = nil, params = nil, headers = nil)
     # Makes a GET HTTP request without a body.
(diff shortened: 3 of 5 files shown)


lib/faraday/middleware.rb
```diff
@@ -9,5 +9,9 @@ class Middleware
     def initialize(app = nil)
       @app = app
     end
+
+    def close
+      @app&.close
```
technoweenie:
```suggestion
      @app.close if @app.respond_to?(:close)
```

This will prevent it from blowing up when it gets to the Adapter. However, it will still call `Adapter#close` if it's implemented in a specific Faraday adapter.

iMacTia:
Thank you @ioquatix, this is exactly the implementation I was thinking about and it totally make sense. Agree with @technoweenie that we need to check if `close` exists before calling it.

The only problem with this implementation is that we're assuming all middleware inherit from `Faraday::Middleware`. Although that's definitely the recommendation, it's not strictly enforced in v0.x and v1.x. If someone does NOT add the close method to a custom-built middleware that does not inherit from `Faraday::Middleware`, then the result is that the `close` call-chain will be silently interrupted at that middleware.

If we're happy with that, we can polish this PR, add some tests and update the documentation, but that's one of the reasons why I initially left it for v2.0 (where the plan is to make backwards-incompatible changes and force things like all middleware to inherit)

technoweenie:
> then the result is that the close call-chain will be silently interrupted at that middleware.

I agree, a silent fail on `#close` wouldn't be ideal. Another option is to raise if `@app.close` doesn't exist. Subclassing `Faraday::Middleware` or defining `#close` isn't required for middleware, but anyone wanting to release internal resources should know where it failed. Maybe that could be the difference between `#close` and `#close!`

ioquatix:
One option would be to make it a documented requirement for 1.0, and we can make it a warning if it doesn’t exist. That way we don’t break existing code but people are aware it’s wrong.

technoweenie:
I like this, as this will only come up if a developer tries to call `Faraday::Connection#close`, which would be a new 1.0 feature and not common Faraday usage. It should be understandable to anyone using `#close` that not every middleware will support it. If it helps, the functionality in Faraday could be extracted to a module, so it's a trivial fix that doesn't force anyone to disrupt a middleware class' inheritance hierarchy.

ioquatix:
Okay, I think this should be okay now? It's the bare minimum required I guess. It's such a simple interface... I probably don't recommend adding a module for it unless you can roll it into other life-cycle related behaviour.

technoweenie:
This looks good, but how do you actually plan to use it? If this merges, I think you'll have to call something like `conn.builder.app.close` to run this. Are you planning on adding something like `Connection#close` in a future PR?

ioquatix:
Oh I just assumed connection was a kind of middleware, okay ill need to expose that too.

ioquatix:
I don't know how to fix:

```
lib/faraday/connection.rb:15:3: C: Metrics/ClassLength: Class has too many lines. [234/233]
27
  class Connection ...
28
  ^^^^^^^^^^^^^^^^
```

iMacTia:
@ioquatix thanks for continuing to work on this!

Change Line 16 in `.rubocop_todo.yml` (https://github.com/lostisland/faraday/blob/master/.rubocop_todo.yml#L16) from 233 to 234, that should fix it!

technoweenie:
Thanks for your help!

PR #1076, opened 2019-11-14
0.1x deprecate fixes 2

technoweenie:
Fixes some minor issues with the deprecate stuff that I found while trying to port this to the master branch.

1. Fix [rubocop compliant about `===`](https://rubystyle.guide/#no-case-equality)
2. Refactor `DeprecatedClass.proxy_class` so it can accept any version. Bonus: now that the `klass_name` is not built in an isolated `metaclass` scope, it can just use the method's `origclass` argument for the replacement messages, _without regexes_.
3. Use a shared `Gem::Version` instance in wrappers around deprecated methods.

### DIFF
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -7,20 +7,23 @@ module Faraday
   #   usage about deprecation.
   # @see Faraday::Deprecate
   module DeprecatedClass
-    def self.proxy_class(new_klass)
-      Class.new(new_klass) do
-        class << self
+    def self.proxy_class(origclass, ver = '1.0')
+      metaclass = nil
+      proxyclass = Class.new(origclass) do
+        metaclass = class << self
           extend Faraday::Deprecate
-          # Make this more human readable than #<Class:Faraday::ClientError>
-          klass_name = superclass.to_s[/^#<Class:([\w:]+)>$/, 1]
-          deprecate :new, "#{klass_name}.new", '1.0'
-          deprecate :inherited, klass_name, '1.0'
 
           def ===(other)
             other.is_a?(superclass) || super
           end
+
+          self
         end
       end
+
+      metaclass.send(:deprecate, :new, "#{origclass}.new", ver)
+      metaclass.send(:deprecate, :inherited, origclass.name, ver)
+      proxyclass
     end
   end
 
@@ -73,6 +76,7 @@ module Faraday
     # @param ver [String] the semver the method will be removed.
     def deprecate(name, repl, ver)
       class_eval do
+        gem_ver = Gem::Version.new(ver)
         old = "_deprecated_#{name}"
         alias_method old, name
         define_method name do |*args, &block|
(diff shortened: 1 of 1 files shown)


technoweenie:
@BobbyMcWho: Do you have any thoughts on these changes to the Deprecate code? I want to port it to the master branch, so I need `#proxy_class` to accept a `ver` argument.

BobbyMcWho:
@technoweenie minor changes I propose in https://github.com/lostisland/faraday/pull/1077

technoweenie:
Thanks @BobbyMcWho. I like the nested PR trick too :)

lib/faraday/deprecate.rb
```diff
@@ -7,20 +7,19 @@ module Faraday
   #   usage about deprecation.
   # @see Faraday::Deprecate
   module DeprecatedClass
-    def self.proxy_class(new_klass)
-      Class.new(new_klass) do
+    def self.proxy_class(origclass, ver = '1.0')
```
BobbyMcWho: This is a little misleading now actually that I think about it, this method always takes the _new_ class, since the way this is used is 
`const_set(OldClass, DeprecatedClass.proxy_class(NewClassThatShouldBeUsedInstead))` do we need another method to distinguish a truly deprecated class vs a replaced one?

BobbyMcWho: I suppose you could always just `extend Faraday::Deprecate` and write the `:new` and `:inherited` DSL in that class itself, I'm just trying to make sure we have a clear case when to use what

technoweenie: > do we need another method to distinguish a truly deprecated class vs a replaced one?

What's the difference, in your mind? Is a truly deprecated class one that's going away in the next major version? And a replaced class was just moved for some reason (like the Error classes)? 

If so, I think this `proxy_class` method is just fine. I changed the argument name from `new_klass` to `origclass` because the distinction whether the class is new or old didn't seem to matter. There are two use cases I see `#proxy_class` being used:

1. A const is moving to another module. In this case, consts like `Faraday::Error::TimeoutError` were moved to `Faraday::TimeoutError`.

https://github.com/lostisland/faraday/blob/2f1a8c9b75cbe80e8364235311f3c3ca0d361b8d/lib/faraday/error.rb#L112-L118

Instead of new/old, I'd say the code usage is more like:

    const_set(:OldName, DeprecatedClass.proxy_class(MovedClass))

2. A const is no longer used. I'm not even sure how we'd do this to be honest. For example, once we abolish network timeouts:

```ruby
# this will still warn on #inherited
class TimeoutError < DeprecatedClass.proxy_class(ClientError)
end

# this is unusual, but should work
timeoutError = Class.new(ClientError)
const_set(:TimeoutError, DeprecatedClass.proxy_class(timeoutError))
```

FWIW I did experiment with removing `Faraday::DeprecatedClass` in favor of `Faraday::Deprecate.proxy_class` instead. I didn't like the idea of `proxy_class` being available when `Faraday::Deprecate` extends other classes, so I kept it separate.

BobbyMcWho: Yeah, that's why I had namespaced it differently originally, I agree that's weird. I think I was overthinking it, and I don't see any reason not to merge this 👍

PR #1077, opened 2019-11-14
Remove the intermediate variable

BobbyMcWho:
Edit: Note the target branch for this PR, I suppose I coulda just linked a git diff in the other PR, but this makes it easier to pick from I think 

Could `.tap` this to remove the proxy variable as well, but this reads a little cleaner than the tap there.


technoweenie:
Right on, even better. I forgot `#singleton_class` even existed :)

Agreed about not using `#tap` here :+1:

PR #1084, opened 2019-12-09
Fix empty array for FlatParamsEncoder {key: []} -> "key="

mrexox:
## Description
For an empty array in params hash should not ignore the key.
- [x] Test

## Additional Notes
The problem is: **each on empty array does not yield**.

### DIFF
--- a/lib/faraday/encoders/flat_params_encoder.rb
+++ b/lib/faraday/encoders/flat_params_encoder.rb
@@ -45,9 +45,13 @@ module Faraday
         if value.nil?
           buffer << "#{encoded_key}&"
         elsif value.is_a?(Array)
-          value.each do |sub_value|
-            encoded_value = escape(sub_value)
-            buffer << "#{encoded_key}=#{encoded_value}&"
+          if value.empty?
+            buffer << "#{encoded_key}=&"
+          else
+            value.each do |sub_value|
+              encoded_value = escape(sub_value)
+              buffer << "#{encoded_key}=#{encoded_value}&"
+            end
           end
         else
           encoded_value = escape(value)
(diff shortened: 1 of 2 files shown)


lib/faraday.rb
```diff
@@ -19,7 +19,7 @@
 #   conn.get '/'
 #
 module Faraday
-  VERSION = '0.16.0'
+  VERSION = '0.16.1'
```
olleolleolle: Please let the maintenance team do the update to this file.

Can you undo this change?

mrexox: Yep!

mrexox:
What about CI? Seems that it is broken just for ruby `2.3.x`. Should I add a fix for it?

olleolleolle:
@mrexox If you can create a separate PR fixing the 2.3 - or if the 2.3 is truly inaccessible, removing it from GitHub actions - that'd be super. 

Don't fix it in this focused PR. This one is good as it is.

olleolleolle:
According to https://github.com/actions/setup-ruby/blob/master/action.yml 2.4 seems to be lower bound.

mrexox:
@olleolleolle, done with it: https://github.com/lostisland/faraday/pull/1085

mrexox:
Can this PR be merged?

technoweenie:
I agree with this change, because it matches how an empty string is decoded.

```irb
>> Faraday::FlatParamsEncoder.encode(a: [])
=> ""
>> Faraday::FlatParamsEncoder.encode(a: '')
=> "a="
```

In my opinion, if you don't want the encoded url to have an empty value (`a=`), then don't send an empty value (`a: []` or `a: ''`). But if empty values are being encoded, they should behave roughly the same.

PR #1085, opened 2019-12-10
Fix CI for Ruby 2.3

mrexox:
Support for github action of ruby 2.3 was removed. Now according to https://github.com/actions/setup-ruby/blob/master/action.yml the minimal version is 2.4.


mrexox:
Should README be fixed according to minimal Ruby version changes?

PR #1086, opened 2019-12-10
Fix 1.9 syntax error

BanzaiMan:
## Description
0.1x is supposed to be 1.9-compat, so change this method signature to reflect it.

## Todos
- [ ] Tests
- [ ] Documentation

## Additional Notes
https://travis-ci.com/travis-ci/travis.rb/jobs/263075645#L271

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -85,7 +85,7 @@ module Faraday
 
   # Raised by Faraday::Response::RaiseError in case of a nil status in response.
   class NilStatusError < ServerError
-    def initialize(_exc, response: nil)
+    def initialize(_exc, response = nil)
       message = 'http status could not be derived from the server response'
       super(message, response)
     end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [2.3.x, 2.4.x, 2.5.x, 2.6.x]
+        ruby: [2.4.x, 2.5.x, 2.6.x]
 
     steps:
     - uses: actions/checkout@v1


BanzaiMan:
Not sure how to fix the CI.

olleolleolle:
@BanzaiMan GitHub actions doesn't support 2.3 any longer. On master, that element of the ci matrix was dropped, to go green.

Such a PR is needed in this branch, too.

BanzaiMan:
Not sure why GitHub thinks 2.3 is still required and expected. ¯\_😕_/¯

lib/faraday/error.rb
```diff
@@ -85,7 +85,7 @@ def initialize(exc = 'timeout', response = nil)
 
   # Raised by Faraday::Response::RaiseError in case of a nil status in response.
   class NilStatusError < ServerError
-    def initialize(_exc, response: nil)
+    def initialize(_exc, response = nil)
```
iMacTia: This is NOT the same thing as `def initialize(_exc, response: nil)`

iMacTia:
Well, according to the 0.1x README we do 😃! But no tests are executed against 1.9 anymore, so hard to catch this sort of things 🤔

lib/faraday/error.rb
BanzaiMan: oh, that's true. However, I will point out that this is the same call signature as that of `TimeoutError` on line 80. If this is not the right way, then `TimeoutError` should also be revisited.

BanzaiMan:
https://github.com/lostisland/faraday/pull/1086/files#diff-1b35e95f85a819bcd3c16d920ecdcffeR112-R113 has `%i`, which I believe is not 1.9-compatible.

lib/faraday/error.rb
mrexox: Can it be rewritten to accept a hash parameter? Like that:
```ruby
def initialize(_exc, params)
  response = params[:response]
```

iMacTia:
Unfortunately I was unable to get 1.9 working again on our GitHub Actions CI, so it's hard to say if we can or can't merge this PR.
Moreover we've consistently merged incompatible changes with 2.0, possibly more than the ones already highlighted in this PR.

@BanzaiMan are you still using Ruby 1.9 together with the latest Faraday version?
@olleolleolle @technoweenie latest releases (0.17.x, possibly sooner) have already broken the compatibility with 1.9, should we come back on track or just drop the official support?

BanzaiMan:
@iMacTia  > are you still using Ruby 1.9 together with the latest Faraday version?
Yes. For many reasons, we need Faraday to provide a 1.9-compatible library in some form. I understand that 1.9 is certifiably EOL, but if it is no longer supported, the incompatible versions should be yanked so that we have some reasonable version that supports 1.9 can be found.

Looks to me that #1094 supersedes this, so I'll close this now.

PR #1087, opened 2019-12-10
Update ci.yml

iMacTia:
## Description
Uses RVM to test 0.1x branch against correct ruby versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [1.9, 2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -16,19 +16,24 @@ jobs:
...
+      run: |
+        curl -sSL https://get.rvm.io | bash
+        
     - name: Set up Ruby
-      uses: actions/setup-ruby@v1
-      with:
-        ruby-version: ${{ matrix.ruby }}
+      run: |
+        source $HOME/.rvm/scripts/rvm
+        rvm install ${{ matrix.ruby }} --binary
```
olleolleolle:
```suggestion
        rvm install ${{ matrix.ruby }} --disable-binary
```

The build failures seem to be about there being no prepared binaries.

iMacTia:
Couldn't manage to make it work with 1.9, but I really run out of time so this will do for now 😅 
Added back CI support for 2.0+ 🎉

PR #1091, opened 2019-12-23
Add `Adapter#close` so that derived classes can call super.

ioquatix:


ioquatix:
https://github.com/socketry/async-http-faraday/blob/master/lib/async/http/faraday/adapter.rb now supports this pretty well. It has both persistent connections, and concurrent requests.

iMacTia:
Ignore the rubocop issues, I'm fixing them on master 👍

PR #1094, opened 2019-12-27
Fix 1.9 syntax error (with Error fixes)

technoweenie:
This PR adds fixes some new failing tests added to #1086.

Changing the signature from:

```diff
# lib/faraday/error.rb
# Faraday::NilStatusError
-def initialize(_exc, response: nil)
+def initialize(_exc, response = nil)
```

Is a major change, but it's only used in one spot:

```diff
# lib/faraday/response/raise_error.rb
-raise Faraday::NilStatusError, response: response_values(env)
+raise Faraday::NilStatusError, response_values(env)
```

This PR updates it, and adds a deprecation warning.

```sh
>> Faraday::NilStatusError.new(response: {status: 1}).inspect
NOTE: Faraday::NilStatusError#unwrap_resp is deprecated; use  instead. It will be removed in or after version 1.0
Faraday::NilStatusError#unwrap_resp called from /Users/rick/p/faraday/lib/faraday/error.rb:121.
=> "#<Faraday::NilStatusError response={:status=>1}>"

>> Faraday::NilStatusError.new({status: 1}).inspect
=> "#<Faraday::NilStatusError response={:status=>1}>"
```

For the `master` branch, and eventually v1.0:

* [ ] Apply updated specs and fixes
* [ ] Ensure all error constructors are consistent
* [ ] Remove deprecation warnings
* [ ] Stop advertising ruby 1.9 support

PR #1095, opened 2019-12-27
improve error initializer consistency

technoweenie:
This brings over the changes to Error initialization and the specs from #1094. 

* [x] Apply updated specs and fixes
* [x] Ensure all error constructors are consistent
* [x] Remove deprecation warnings

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -98,7 +98,7 @@ module Faraday
   end
 
   # A unified client error for timeouts.
-  class TimeoutError < ClientError
+  class TimeoutError < ServerError
     def initialize(exc = 'timeout', response = nil)
       super(exc, response)
     end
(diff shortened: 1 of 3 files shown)


lib/faraday/error.rb
```diff
@@ -74,35 +98,35 @@ class ServerError < Error
...
       super(exc, response)
     end
   end
 
   # Raised by Faraday::Response::RaiseError in case of a nil status in response.
   class NilStatusError < ServerError
-    def initialize(_exc, response: nil)
-      message = 'http status could not be derived from the server response'
-      super(message, response)
+    def initialize(exc, response = nil)
```
iMacTia: Since 1.0 is not supposed to be compatible with 1.9, we should be free to leave the new syntax in place

lib/faraday/error.rb
```diff
@@ -74,35 +98,35 @@ class ServerError < Error
...
+      exc_msg_and_response!(exc, response)
+      super('http status could not be derived from the server response')
     end
   end
 
   # A unified error for failed connections.
-  class ConnectionFailed < Error
+  class ConnectionFailed < ClientError
```
iMacTia: I previously changed the superclass of these classes because it made more sense.
These changes have been reflected on the `UPGRADING.md` so people should be aware.
Error classes are a bit messy in Faraday, 1.0 is our only chance to change that while making people aware.
If we want to leave them as they are then this will need to wait 2.0.
Do you think it will be too much disruptive?

lib/faraday/error.rb
technoweenie: Normally I'd agree, but I think I made a mistake not insisting on consistency with the other error classes in the original PR.

lib/faraday/error.rb
technoweenie: Doh, I'll revert those changes. It was a misunderstanding of mine.

Issue #1089, opened 2019-12-12
Faraday::Error::ClientError deprecation warning is disruptive


DannyBen:
Displaying deprecation warnings to STDOUT/STDERR is very disruptive. 

Although it is "just a warning", we need to consider that in some cases, these warnings break CI (when checking for expected output) and in other cases, displaying a warning to the end user of a CLI gem is undesirable.

It causes people who are not even using Faraday directly, to [waste time][3] on trying to find workarounds, and depend on [the offending package maintainer][1] to [release an update][2].

I would like to suggest avoiding this practice in the future, and instead, let the magic of semantic versioning do what it was designed to do. I would even go further and suggest to release an updated version without this warning.

[1]: https://github.com/octokit/octokit.rb/pull/1154
[2]: https://github.com/octokit/octokit.rb/issues/1170
[3]: https://github.com/DannyBen/madman/pull/19

iMacTia:
Have you considered telling the logger to skip warnings and just log errors in your CI?

DannyBen:
You should not treat your libraries as if they are only used by developers. Especially when it is such a big and widely used library. At the end of the dependency chain, there are users, who end up seeing useless deprecation warnings, and think that *my* tool is broken. 

Runtime is user-land, not developer-land. Never disrupt runtime.

Forget the CI case, I am not sure why you picked that aspect of my issue to respond to of all problems. This is a development warning, should stay in development land.

iMacTia:
OK, I wasn't trying to minimise the issue, but according to your initial question it seems like `warning` logs have no meaning at all. I'd be interested to know in which case you'd use one?

I was just pointing out that `warning` has been defined as standalone log level specifically for these cases, they're logged for developers, not end users.
Moreover, "end users" should not see warnings or have access to log files. I don't know what sort of  application you're using Faraday in, but if it's Web/UI based like normal people are used to, then warnings should not be visible to them.
User usually only see `error` or `fatal` logs.

If that's not the case for you, then happy to explore further how your specific use case works, but please don't assume everyone is having your same issue, otherwise we'd be seeing many more issues like this one

DannyBen:
Well, one of us might be missing some piece of information.

I am not talking about logs, of course logs are usually considered developer-land.

I am talking about STDOUT / STDERR. 

I was specifically exposed to this problem by using octokit. See this issue: 
https://github.com/octokit/octokit.rb/issues/1170

[...]

PR #1098, opened 2019-12-30
require FARADAY_DEPRECATE=warn to show Faraday v1.0 deprecation warnings

technoweenie:
This makes the Faraday deprecation warnings opt-in, as described in https://github.com/lostisland/faraday/issues/1089#issuecomment-569337738.

By default, no Faraday 1.0 deprecations are shown. If the ruby app is run with `FARADAY_DEPRECATE=warn` or `FARADAY_DEPRECATE=1` in ENV, then a ruby warning will be given for each deprecated method call.


iMacTia:
@technoweenie something doesn't look right, the check against `@skip` is a boolean one: https://github.com/lostisland/faraday/pull/1098/files#diff-b0f6120f9db43c327fe8c66333df82b3R99

But after this PR, `@skip` is set to either `:skip` or `:warn`.
Wouldn't both symbols be evaluated as `true`?

technoweenie:
The `#skip` method actually returns `@skip == :skip`. I did this so the `@skip ||= ...` wouldn't keep evaluating if `@skip` is false. 

Here's the behavior on the current 0.1x branch (e8a6662):

```
$ irb -Ilib
# skip by default now!
>> Faraday::Deprecate.skip
=> true
>> Faraday::Deprecate.instance_variable_get :@skip
=> :skip
>> Faraday::Deprecate.skip = false
>> Faraday::Deprecate.instance_variable_get :@skip
=> :warn

$ FARADAY_DEPRECATE=warn irb -Ilib
>> require 'faraday'
=> true
>> Faraday::Deprecate.skip
=> false
>> Faraday::Deprecate.instance_variable_get :@skip
=> :warn
```

iMacTia:
Right, I missed the `@skip == :skip` line at the end 😅!
All good then!

PR #1101, opened 2019-12-30
Ship 1.0

technoweenie:
* [x] Review/merge error class changes #1095
* [x] push Faraday::Deprecate upstream for eventual 2.0 deprecations?
  * no need now. we can grab it any time.
* [x] double check readme/gemfile/etc for consistency in faraday website, supported ruby versions, gems, etc. also update minimum ruby version to 2.4.0
* [x] bump supported addressable version from <2.4.0 to <2.7.0?
  * addressable version is unspecified already
* [x] bump supported ruby version to 2.3.x or 2.4.x?
* [x] drop CONNECT method support?
* [x] Changelog
* [x] Verify against technoweenie/faraday-live

Items ending in `?` are suggestions...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ Misc:
 * Describe clearing cached stubs #1045 (@viraptor)
 * Add project metadata to the gemspec #1046 (@orien)
 
-## v0.17.2
+## v0.17.3
 
 Fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Faraday Changelog
 
-## v1.0-rc1
+## v1.0
 
 Features:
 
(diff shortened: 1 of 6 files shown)


technoweenie:
I ran the [faraday-live](https://github.com/technoweenie/faraday-live) test suite against this branch, and the current version of Typhoeus in Git (until v1.4.0 is released). I ran into this issue with the `CONNECT` calls:

```
  1) typhoeus with HTTPS server (typhoeus) #CONNECT request behaves like an idempotent request returns Content-Type response header
     Failure/Error: @response = conn.public_send(http_method, 'idempotent_request')

     Faraday::ConnectionFailed:
       Stream error in the HTTP/2 framing layer
     Shared Example Group: "any request" called from ./spec/support/examples/request_idempotent.rb:11
     Shared Example Group: "an idempotent request" called from ./spec/support/examples/connection.rb:24
     Shared Example Group: "a connection making requests" called from ./spec/secure_spec.rb:5
```

I propose that we cut `CONNECT` method support before 1.0. It's never actually showed up in a release version of Faraday, so it won't be a breaking change. While [`TRACE`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/TRACE) is obscure, it's easy enough to support. However, [`CONNECT`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) seems to be used for proxy servers, which doesn't work well with Faraday's request/response nature.

technoweenie:
I think this is ready to go now. I've updated the changelog, including the 0.17.x updates that weren't ever written to the master branch's changelog.

I also added author attributions for items in Faraday 0.14.0, but that was because I was accidentally adding the new changelog items in the wrong place :)

PR #1106, opened 2020-01-02
Handle separation of positional and keyword arguments in ruby 2.7

mcgregordan:
Ruby 2.7 separates positional and keyword arguments. This resulted in several deprecation warnings in `Faraday::RackBuilder`. Keyword arguments are often passed to it that are implicitly passed through as a hash (as the last positional argument). Ruby 2.7 raises a warning for this and 3.0 will error.

For example, faraday-http-cache, which is configured something like `builder.use Faraday::HttpCache, serializer: Marshal, shared_cache: false`, will raise warnings, because RackBuilder doesn’t handle the Ruby-2.7 change.

This PR removes the deprecation in a backwards- and forward-compatible way. It uses a new compatability method in ruby 2.7 and a shim to support older rubies, which is recommended on the [ruby blog](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/).

### DIFF
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -19,6 +19,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 2.3'
 
   spec.add_dependency 'multipart-post', '>= 1.2', '< 3'
+  spec.add_dependency 'ruby2_keywords'
 
   files = %w[CHANGELOG.md LICENSE.md README.md Rakefile examples lib spec]
   spec.files = `git ls-files -z #{files.join(' ')}`.split("\0")
--- a/lib/faraday/rack_builder.rb
+++ b/lib/faraday/rack_builder.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'faraday/adapter_registry'
+require 'ruby2_keywords'
 
 module Faraday
   # A Builder that processes requests into responses by passing through an inner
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -31,3 +31,5 @@ group :test do
 end
 
 gemspec
+
+gem 'ruby2_keywords', '~> 0.0.1'
```
olleolleolle: This perhaps ought to live in the gemspec itself?

mcgregordan: 🤦‍♂️ fixed

iMacTia:
@olleolleolle @mcgregordan sorry for the delay, busy return from Holidays.
I'm having a look at this now!

iMacTia:
OK, I initially thought this was unnecessary as the RackBuilder doesn't really accept keyword arguments, however I believe this change will technically allow middleware and adapters to receive keyword arguments without warnings in new and old Ruby versions.

I'd however feel a little more comfortable if we also had a test with a dummy middleware/adapter that shows this is working as expected (and since our tests run against all ruby version between 2.4 and 2.7, we will be sure this works as expected).

Is that something we can add before merging the PR?

mcgregordan:
Thanks @iMacTia. I think middleware and adapters are the crucial point. Otherwise this would be unnecessary, as you say.

Agreed on the extra test. I’ll add something over the next few days.

iMacTia:
This has now been implemented and test-covered in #1153

PR #1109, opened 2020-01-08
Documentation fix

iMacTia:
## Description
Mentions in the documentation that the `Retry` middleware handles the `Retry-After` header.
Fixes https://github.com/lostisland/faraday/pull/773#issuecomment-571089608


vvo:
Hi @iMacTia I believe if you want retry-after to be honored, the minimum configuration is: `faraday.request :retry, retry_statuses: [429]` do you confirm? If so I think the doc should provide that, Ill open a PR

PR #1111, opened 2020-01-09
docs(retry): precise usage of retry-after

vvo:
This PR adds documentation about the Retry middleware's handling of the `Retry-After` HTTP header.

- adds an example
- adds a hyperlink to MDN

fixes https://github.com/lostisland/faraday/pull/1109#issuecomment-572448991

### DIFF
--- a/docs/middleware/request/retry.md
+++ b/docs/middleware/request/retry.md
@@ -81,7 +81,9 @@ retry_options = {
 
 #### Automatically handle the `Retry-After` header
 
-Some APIs, like the [Slack API](https://api.slack.com/docs/rate-limits), will inform you when you reach their API limits by replying with a response status code of `429` and a response header of `Retry-After` containing a time in seconds. You should then only retry querying after the amount of time provided by the `Retry-After` header, otherwise you won't get a response. You can automatically handle this and have Faraday pause and retry for the right amount of time by configuring it this way:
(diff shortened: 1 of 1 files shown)


docs/middleware/request/retry.md
```diff
@@ -79,6 +79,16 @@ retry_options = {
 }
 ```
 
+#### Automatically handle the `Retry-After` header
+
+Some APIs, like the [Slack API](https://api.slack.com/docs/rate-limits), will inform you when you reach their API limits by replying with a response status code of `429` and a response header of `Retry-After` containing a time in seconds. You should then only retry querying after the amount of time provided by the `Retry-After` header, otherwise you won't get a response. You can automatically handle this and have Faraday pause and retry for the right amount of time by configuring it this way:
```
olleolleolle: Perhaps a wording that mentions "including the HTTP status 429", so that it is additive to the above example (which has a list of 2 retry statuses).

I love this example's precision.

vvo: Hey thanks, do you mean precising this way: "You can automatically handle this by including the HTTP status 429 in the config..", sorry I am not sure what to change but if you want to add something to this PR definitely do it too!

olleolleolle: Yeah, by mentioning "including", so that the reader does not understand it as "replace the list" of statuses, but to "add to the list".

olleolleolle: Thanks for the note - I was emboldened to make the change in this PR.

PR #1114, opened 2020-01-14
Backport wrapping Errno::EADDRNOTAVAIL in the net/http adapter to 0.1x

embs:
Just as 148b99ee0fde8 also does but for the v0.1x releases.

### DIFF
--- a/lib/faraday/adapter/net_http.rb
+++ b/lib/faraday/adapter/net_http.rb
@@ -11,6 +11,7 @@ module Faraday
     class NetHttp < Faraday::Adapter
       NET_HTTP_EXCEPTIONS = [
         IOError,
+        Errno::EADDRNOTAVAIL,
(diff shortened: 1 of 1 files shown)


embs:
@iMacTia this is a follow up on what we've discussed on [gitter](https://gitter.im/lostisland/faraday?at=5e14a6a404adee486b48e32a).

I wonder what would be the better strategy for testing this change. We don't seem to have unit specs for the Net::HTTP adapter in the `0.1x` trunk. It looks like the automated tests undergone a comprehensible refactor in 1.x (I can't see the `test/` dir there anymore).

I imagine the following two possibilities for testing this (if desirable right now for the `0.1x`):

- introducing unit specs for the adapter and making sure the error is wrapped (I guess this would demand less effort than)
- setting up a new env so we can reproduce the issue in an integration test

What do you think?

Issue #1119, opened 2020-01-22
rack + rack-test: Test suite fails on master


olleolleolle:
TL;DR: `undefined method 'split' for nil:NilClass` on tests about rack adapter.

## Basic Info

* Faraday Version: current master
* Ruby Version: any

## Issue description

There seems to be a rack-test/rack mismatch.

```
     NoMethodError:
       undefined method `split' for nil:NilClass
     Shared Example Group: "a request method" called from ./spec/support/shared_examples/adapter.rb:101
     Shared Example Group: "adapter examples" called from ./spec/support/shared_examples/adapter.rb:8
     Shared Example Group: "an adapter" called from ./spec/faraday/adapter/rack_spec.rb:7
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-2.1.1/lib/rack/mock.rb:236:in `identify_cookie_attributes'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-2.1.1/lib/rack/mock.rb:220:in `block in parse_cookies_from_header'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-2.1.1/lib/rack/mock.rb:218:in `each'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-2.1.1/lib/rack/mock.rb:218:in `parse_cookies_from_header'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-2.1.1/lib/rack/mock.rb:169:in `initialize'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-test-1.1.0/lib/rack/mock_session.rb:31:in `new'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-test-1.1.0/lib/rack/mock_session.rb:31:in `request'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-test-1.1.0/lib/rack/test.rb:266:in `process_request'
     # /home/runner/.rvm/gems/ruby-2.5.7/gems/rack-test-1.1.0/lib/rack/test.rb:119:in `request'
```

Source: https://github.com/lostisland/faraday/runs/400710918#step:8:3379

## Steps to reproduce

Run the test suite from current master.

Or, watch the CI output linked above.

olleolleolle:
Thanks a lot!

Issue #974, opened 2019-04-25
Setting Authentication Header in Rails results in NameError


JanOwiesniak:
## Basic Info

* Faraday Version: faraday (0.15.4)
* Ruby Version: ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]
* Rails Version: rails (5.2.3)

## Issue description

I have issues with setting `Authentication` header in a Rails environment. It breaks with

```
uninitialized constant Faraday::Request::Authorization::KEY
```

## Steps to reproduce

**irb (good)**

```
$ irb

2.5.1 :001 > require 'faraday'
 => true
2.5.1 :002 > conn = Faraday.new(url: "http://example.com") # create a new Connection with base URL
 => #<Faraday::Connection:0x00007feb7f059408 @parallel_manager=nil, @headers={"User-Agent"=>"Faraday v0.15.4"}, @params={}, @options=#<Faraday::RequestOptions (empty)>, @ssl=#<Faraday::SSLOptions (empty)>, @default_parallel_manager=nil, @builder=#<Faraday::RackBuilder:0x00007feb7f02bf58 @handlers=[Faraday::Request::UrlEncoded, Faraday::Adapter::NetHttp]>, @url_prefix=#<URI::HTTP http://example.com/>, @manual_proxy=false, @proxy=nil, @temp_proxy=nil>
2.5.1 :003 > conn.basic_auth("user", "pass")               # set the Authentication header
 => "Basic dXNlcjpwYXNz"
2.5.1 :004 > conn.get("/foo")
[...]

PR #1122, opened 2020-02-03
Fully qualify constants to be checked

olleolleolle:
Refer to constants in ways that their names  never collide.

## Description

See https://github.com/lostisland/faraday/issues/974 for details.

This  quote lists  the route this PR took:

> Another fix would be changing the defined? checks to look at the ::Expanded::Module::Path. I'm going to list other cases in the code base that this is done, so we can make a single consistent fix:

~**In order to see if tests are green** - this PR disables rack adapters tests which suffer #1119~

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -153,7 +153,7 @@ module Faraday
     @default_connection_options = ConnectionOptions.from(options)
   end
 
-  unless const_defined? :Timer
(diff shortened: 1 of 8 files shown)


iMacTia:
@olleolleolle you can now pull latest from `master` and re-enable rack tests 👍

olleolleolle:
@iMacTia There - squash-ready!

PR #1123, opened 2020-02-05
allow private parse method in response middleware

nic-lan:
## Description
Currently the Response Middleware enforces children classes to define public `parse` method.
This is due to the implicit arg `all_included=false` passed to [respond_to? method](https://ruby-doc.org/core-2.7.0/Object.html#method-i-respond_to-3F)

Now I would like to allow also private and protected `parse` method to allow more flexibility for the children classes.

<details>

## Todos
- [x] Tests
- [x] Documentation

## Additional Notes
When running `bundle exec rspec` on master i noticed a bounce of failing specs which are not depending on this PR code changes.

The error is:
```console
Failure/Error: @session.request(env[:url].to_s, rack_env)

      NoMethodError:
        undefined method `split' for nil:NilClass
```

and it refers to [lib/faraday/adapter/rack.rb:62](https://github.com/lostisland/faraday/blob/099dd45f63ff99bbb343eebf7504a3cf0b10bc63/lib/faraday/adapter/rack.rb#L63)

Maybe an issue to fix that should be open.

</details>

### DIFF
--- a/lib/faraday/response.rb
+++ b/lib/faraday/response.rb
@@ -15,8 +15,12 @@ module Faraday
 
       # Override this to modify the environment after the response has finished.
       # Calls the `parse` method if defined
+      # `parse` method can be defined as private, public and protected
       def on_complete(env)
-        env.body = parse(env.body) if respond_to?(:parse) && env.parse_body?
+        all_included = true
+        return unless respond_to?(:parse, all_included) && env.parse_body?
+
+        env.body = parse(env.body)
       end
     end
 
--- a/lib/faraday/response.rb
+++ b/lib/faraday/response.rb
@@ -17,8 +17,7 @@ module Faraday
       # Calls the `parse` method if defined
       # `parse` method can be defined as private, public and protected
       def on_complete(env)
-        all_included = true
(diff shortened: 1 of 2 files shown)


lib/faraday/response.rb
```diff
@@ -15,8 +15,12 @@ def call(env)
 
       # Override this to modify the environment after the response has finished.
       # Calls the `parse` method if defined
+      # `parse` method can be defined as private, public and protected
       def on_complete(env)
-        env.body = parse(env.body) if respond_to?(:parse) && env.parse_body?
+        all_included = true
+        return unless respond_to?(:parse, all_included) && env.parse_body?
```
iMacTia: Any reason for defining `all_included` as a separate variable and then using it on the next line?
Why not just using `respond_to?(:parse, true)`? People should know what the 2nd param for `respond_to?` is.

nic-lan: I added the separate variable because to me was a way to increase the readability of the code since i was not expecting everybody to know signature for `respond_to?`.

That been said i don't think it is necessary so I am happy to apply to your change request.

:-)

nic-lan:
change as you required.

Let me know if you have anything else that does not convince you.

PR #1125, opened 2020-02-06
Encode Spaces in Query Strings as '%20' Instead of '+'

### LINKED ISSUE #1124: Encode Spaces in Query Strings as '%20' Instead of '+'
## Issue description
While `+` is supposed to be a valid encoding for spaces in query strings, it seems like it is causing some issues with some servers (namely [Apple's](https://github.com/nhosoya/omniauth-apple/pull/16))

Given that encodeURIComponent encodes spaces as `%20`, and `%20` is what is recommended by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.1) maybe that could be the default?

Or provide some way to make this configurable?

## Steps to reproduce
See [omniauth-apple](https://github.com/nhosoya/omniauth-apple/pull/16)

karlentwistle:
While `+` is supposed to be a valid encoding for spaces in query strings, it seems like it is causing some issues with some servers (namely [Apple's](https://github.com/nhosoya/omniauth-apple/pull/16))

Given that encodeURIComponent encodes spaces as `%20`, and `%20` is what is recommended by [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.1) maybe this could be the default. 

Fixes #1124

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -16,12 +16,12 @@ module Faraday
       NestedParamsEncoder.encode(params)
     end
 
-    ESCAPE_RE = /[^a-zA-Z0-9 .~_-]/.freeze
+    ESCAPE_RE = /[^a-zA-Z0-9.~_-]/.freeze
 
     def escape(str)
       str.to_s.gsub(ESCAPE_RE) do |match|
         '%' + match.unpack('H2' * match.bytesize).join('%').upcase
-      end.tr(' ', '+')
+      end
     end
 
     def unescape(str)
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -16,12 +16,20 @@ module Faraday
       NestedParamsEncoder.encode(params)
     end
 
-    ESCAPE_RE = /[^a-zA-Z0-9.~_-]/.freeze
+    def default_space_encoding
+      @default_space_encoding ||= '+'
+    end
+
+    class << self
+      attr_writer :default_space_encoding
+    end
+
+    ESCAPE_RE = /[^a-zA-Z0-9 .~_-]/.freeze
 
     def escape(str)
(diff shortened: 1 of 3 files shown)


iMacTia:
@karlentwistle although I see your point, as `+` is more of a ruby convention, I'd still prefer to maintain backwards compatibility.
For this reason, would you mind changing your PR so that this new behaviour is configurable? And the old behaviour (`+`) stays the default one?

karlentwistle:
@iMacTia thanks for taking the time to review my code. I will update this PR very soon to make the behaviour configurable.

karlentwistle:
@iMacTia I have updated my PR and would be really grateful if you could take another look 😺

olleolleolle:
If that ain't the case, then it's possible to use an `around` hook: 

```ruby
context 'here' do
  around do |example|
    Faraday::Utils.default_space_encoding = '%20'
    example.run
    Faraday::Utils.default_space_encoding = nil
  end

  it 'the existing thing' do
    # ...
   end
end

karlentwistle:
Sorry for the delay @iMacTia this had totally slipped my mind with all the COVID-19 stuff going on. I have refactor the test now if you'd like?

iMacTia:
@karlentwistle no worries at all, totally understand! Happy for you to do it if you want to see this through 😄!

karlentwistle:
Thanks @iMacTia I have just updated the PR. It was a bit fiddly to figure out what to put in the `context` and `it` descriptions so more than happy to apply any suggestions if you want them tweaked/rewritten.

iMacTia:
Thanks @karlentwistle 🎉
LGTM!

karlentwistle:
Oh brilliant. Thanks very much, @iMacTia and @olleolleolle appreciate you taking the time to review and merge my pull request. I love Faraday so it's wonderful to be able to contribute something to the project.

Issue #956, opened 2019-03-29
Inconsistent response attribute for Faraday::RetriableResponse raised on Retry loop


Alan-M-Thomaz:
## Basic Info

* Faraday Version: 0.15.4
* Ruby Version: 2.5.1p57

## Issue description
The response on the exception Faraday::RetriableResponse raised by  retry loop is a instance of Faraday::Response, while errors raised by the middleware RaiseError are simple hash.

that is kinda bad for checking body, headers, and status, because in one they are attributes, and in the other they are keys

## Steps to reproduce
Using webmock , and rspec.
i removed some unnecessary code, you probably can also use just faraday stubs instead of webmock.
```ruby
let(:stub_send_request_unavaible_server) do
  i = 0
  stub_request(:post, "https://www.test.com/send").with(
    body: "{}",
    headers: {
 	  'Accept'=>'*/*',
 	  'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
 	  'User-Agent'=>'Faraday v0.15.4'
  }).to_return do |stub|
    body = {}
    status = ( i == 0) ? 501 : 200
    body[:result] = "error" if i == 1
    i += 1
    {status: status, headers: { "Retry-After": "1"}, body: body.to_json}
  end
end

before(:each) do
  stub_send_request_unavaible_server
end

it 'should retry 2 times' do
  retry_if_func = lambda do |env, exception|
    p exception.response
    case (exception.response[:status] || exception.response.status)
    when (500..599)
      true
    when (400..499)
      false
    when 200
      body = JSON.parse(exception.response.body)
      body["result"] == "error"
    end
  end
  retryable_exceptions =  Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [ Faraday::ClientError]
  connection = ::Faraday.new(:url => "https://www.test.com") do |faraday|
    faraday.request :retry, max: 5, interval: 0.1, interval_randomness: 0.5, backoff_factor: 2,
                      exceptions: retryable_exceptions, methods: [], retry_statuses: [200],
                      retry_if: retry_if_func
    faraday.response :raise_error
    faraday.adapter Faraday.default_adapter
  end
  connection.post('/send', "{}")
  stub_send_request_unavaible_server.should have_been_made.times(3)
end
```

technoweenie:
[...]

PR #1135, opened 2020-03-15
Improve Faraday::Error to get consistent response information

qsona:
Hi, I faced this problem https://github.com/lostisland/faraday/issues/956 on my development using faraday. I need to make patch in our private codebase but I also have a motivation to make an improvement to the upstream codebase. I'm glad if I can hear your opinions.

## Description

I'm glad if this pull request can achieve these goals:

- Make improvements for faraday v1 (current version) as far as not breaking backward compatibility
  - In this PR, I added some instance methods returning response information to Faraday::Error.
- Make gradual consensus about how to fix that inconsistency problem in faraday v2

### Current Problem

I think it is described well in https://github.com/lostisland/faraday/issues/956 , but I'd like to clarify it here.

The `Faraday::Error` instance wraps `response` object. This object is desired to be Faraday::Response instance, but currently this is just a hash including status/body/headers keys. This makes an inconsistency which is a bit annoying to treat for Faraday users.

One more problem (which is my motivation to solve it) is, the `body` in the response hash is always a raw one, not one rewritten by the response middlewares. For example, I configured as below

(**Edited.** First I specified the response middlewares in the wrong order. Now it's corrected)

```rb
conn = Faraday.new(...) do |conn|
  conn.response :raise_error
  conn.response :json, :content_type => /\bjson$/
end
```

and send a request. Let me assume that the response status code is 500, and body is JSON data like `{ "error": "msg"}` with the `Content-Type: application/json` header.

In this case, an error of Faraday::ServerError instance will be raised by the raise_error middleware. I expect that `error.response[:body]` is a hash `{ "error" => "msg"}`  because json middleware exists before the raise_error middleware, but actually it is just a String `'{ "error": "msg"}'` .

### Improvement for v1 codebase

Currently, `Faraday::Error#response` returns a hash which is not an instance of `Faraday::Response`. This cannot be changed in v1 without breaking the backward compatibility.

As @technoweenie mentioned [here](https://github.com/lostisland/faraday/issues/956#issuecomment-478072433) , we can extend the initializers to receive an actual `Faraday::Response` object, and add some methods to try to get the data from that object. (note: we need to pay attention for that it cannot be got if the error is created in userland)

**In this PR, I added 3 methods: `response_status`, `response_body`, `response_headers`.**

### How it should be in faraday v2 (allowing breaking changes)?

It's not mandatory to discuss this in this PR, but I think it's better to have gradual consensus for the sake of the better change considering the seamless migration.

In my understanding, `Faraday::Error` instance should wrap the `Faraday::Response` instance. **The current response data (hash) will be no longer necessary** because it is a subset of `Faraday::Response` instance.

One point we can discuss is the signature of initializer. Current one is below. (`exc` is expected to  error message or response hash)

```rb
def initialize(exc, response = nil)
end
```

As mentioned [here](https://github.com/lostisland/faraday/pull/1042/files/2c81ed67ed6bfb7b25db0c3b5687faa338ca44ad#r334571035) , the `response` might be desired to an named argument. So I'd like to suggest the signature basically like below:

```rb
# response is a Faraday::Response instance
def initialize(exc = nil, response:)
  @response = response
  # ...
end
attr_reader :response
```
I'm glad if I can hear your opinion.

## Todos

- Documentation (if needed. I think it's not necessary until v2 will be released)

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -6,10 +6,11 @@ module Faraday
   class Error < StandardError
     attr_reader :response, :wrapped_exception
 
-    def initialize(exc, response = nil)
+    def initialize(exc, deprecated_response = nil, response = nil)
       @wrapped_exception = nil unless defined?(@wrapped_exception)
       @response = nil unless defined?(@response)
-      super(exc_msg_and_response!(exc, response))
+      @v2_response = response
+      super(exc_msg_and_response!(exc, deprecated_response))
     end
 
     def backtrace
@@ -26,6 +27,18 @@ module Faraday
       inner << " response=#{@response.inspect}" if @response
       inner << " #{super}" if inner.empty?
       %(#<#{self.class}#{inner}>)
+    end
+
+    def response_status
+      @v2_response&.status
+    end
+
+    def response_body
+      @v2_response&.body
+    end
+
+    def response_headers
+      @v2_response&.headers
     end
 
     protected
--- a/lib/faraday/response/raise_error.rb
+++ b/lib/faraday/response/raise_error.rb
@@ -13,25 +13,34 @@ module Faraday
       def on_complete(env)
         case env[:status]
         when 400
-          raise Faraday::BadRequestError, response_values(env)
+          raise Faraday::BadRequestError
+            .new(response_values(env), nil, env.response)
         when 401
-          raise Faraday::UnauthorizedError, response_values(env)
+          raise Faraday::UnauthorizedError
+            .new(response_values(env), nil, env.response)
         when 403
-          raise Faraday::ForbiddenError, response_values(env)
+          raise Faraday::ForbiddenError
+            .new(response_values(env), nil, env.response)
         when 404
-          raise Faraday::ResourceNotFound, response_values(env)
+          raise Faraday::ResourceNotFound
+            .new(response_values(env), nil, env.response)
         when 407
           # mimic the behavior that we get with proxy requests with HTTPS
           msg = %(407 "Proxy Authentication Required")
-          raise Faraday::ProxyAuthError.new(msg, response_values(env))
+          raise Faraday::ProxyAuthError
+            .new(msg, response_values(env), env.response)
         when 409
-          raise Faraday::ConflictError, response_values(env)
+          raise Faraday::ConflictError
+            .new(response_values(env), nil, env.response)
         when 422
-          raise Faraday::UnprocessableEntityError, response_values(env)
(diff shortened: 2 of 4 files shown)


lib/faraday/error.rb
```diff
@@ -6,10 +6,11 @@ module Faraday
   class Error < StandardError
     attr_reader :response, :wrapped_exception
 
-    def initialize(exc, response = nil)
+    def initialize(exc, deprecated_response = nil, response = nil)
```
qsona: Let me assume that in faraday v2 this signature should be changed to `def initialize(msg = nil, response:)` (as I mentioned in the PR description). Then, first I thought to change the signature at this time to such like

```rb
def initialize(exc, deperecated_response = nil, response: nil)
```

but I couldn't to do that because the old response object is hash and it conflicts with the keyword argument.

qsona:
Can I ignore either Rubocop LineLength check or codeclimate `Method on_complete has 32 lines of code (exceeds 25 allowed).` ?

iMacTia:
@qsona thanks for fixing Rubocop line-length cop, I've told CodeClimate not to complain 😄!
I'll review the PR now, thanks for the detailed summary!

qsona:
> The body is not parsed in your example because the raise_error middleware is added AFTER the json one.

First of all, I'm very sorry about this - actually I mistakenly wrote the PR description, but I specified in the correct order in my private source code. I'll fix the description.

qsona:
@iMacTia I appreciate for your kind advice.

I agree with the plan you suggested. I mixed up a suggestion for future Faraday::Error/raise_error spec (which was my main motivation) and changes for current code and actually I was also a bit confusing. I'd like to separate these two concerns.

About future spec, let me create another issue; meanwhile, I uploaded [my private code](https://gist.github.com/qsona/fdb962538f84e8b7ca4249521a08f6ac) . This is basically what I want, and so I want to discuss this topic (I don't want to remain such a monkey patch in our private code forever). 

About changes for current code (v1), I'd like to fix in this PR based on your advice. Please give me a few days!

iMacTia:
Thanks again @qsona for all the work in collecting details and providing a first solution (with tests!) 🙌
I've collected your code and my comments into #1229 so this can now be closed

qsona:
Oh, I'm really sorry that I completely forgot to work for this. I just noticed by seeing your release note. Thank you very much for your great works.

iMacTia:
No worries at all @qsona! On the contrary, thanks for getting the ball rolling 🙏

Issue #1068, opened 2019-10-26
#get is not thread safe


ioquatix:
https://github.com/lostisland/faraday/blob/d5eaa4ffe04b1cba00691b1cc2b2541e91093bf1/lib/faraday/adapter_registry.rb#L14-L26

Reading from a hash table while there are updates happening is not thread safe.

iMacTia:
Ah I see what you mean, should we move the `@lock.synchronise` call into `get` instead?
I'm not sure we'd need it in `set` anymore after that?

ioquatix:
You need it in both methods because you are reading and writing to `@constants`. You can't have synchronised access to the same hash table in any sitautions. Read only operations are not safe with simultaneous writes.

olleolleolle:
Alright, I merged a change to wrap the `@constants` in a synchronize.

olleolleolle:
Thanks, @ioquatix for the inspiration!

PR #1136, opened 2020-03-25
Use the lock also when reading, in AdapterRegistry

olleolleolle:
## Description

Adapter Registry _stores_ information in "a thread-safe way" (using a synchronize block). But, when _reading_, it doesn't use the lock.

This PR tries to repair the issue reported in #1068, by using a synchronize block around the reading, too.

## Todos

Remaining things

- [ ] Tests
- [ ] Document this kind of thing in the DEVELOP guide?
- [ ] Perhaps have a concurrency label? Is that the right word?

## Additional Notes

This was inspired by watching a video from RWC2019. Here it is: https://www.youtube.com/watch?v=qX_FRa43r4k

### DIFF
--- a/lib/faraday/adapter_registry.rb
+++ b/lib/faraday/adapter_registry.rb
@@ -12,7 +12,9 @@ module Faraday
     end
 
     def get(name)
-      klass = @constants[name]
+      klass = @lock.synchronize do
+        @constants[name]
+      end
       return klass if klass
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
@ioquatix Hi 👋 ! Cool, I've taken the first step, now I guess the rest of the road is still to go.

ioquatix:
It looks okay to me but it's been a while.

PR #1138, opened 2020-03-27
Set middleware_mutex when middleware/adapter classes are defined

olleolleolle:
## Description

See #1074.

This PR is a rebase on `master` in order to be able to run the tests in GitHub Actions. (At the time of #1074 being authored, CI was in a period of change.)

## Todos

- [ ] Tests

## Additional Notes
Optional section

### DIFF
--- a/lib/faraday/adapter.rb
+++ b/lib/faraday/adapter.rb
@@ -33,6 +33,7 @@ module Faraday
 
       def inherited(subclass)
         super
+        subclass.middleware_mutex
         subclass.supports_parallel = supports_parallel?
       end
     end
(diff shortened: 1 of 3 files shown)


olleolleolle:
@ioquatix 👋 We had this lying around.

olleolleolle:
OK, I know that this caused no special conflicts. Will rebase the other branch which had all the good metadata, instead.

ioquatix:
Just FYI, `klass.middleware_mutex` should probably hit a dedicated method... you could even define `Module.initialize` or something similar (not sure if that name would conflict).

PR #1149, opened 2020-04-17
Update RuboCop

AlexWayfer:
Require its specific version in `Gemfile`.

Fix it's installation in CI.

Enable new cops.

Resolve new offenses.

Drop support of Ruby 2.3
(it's not supported by RuboCop and by MRI:https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/)

Deprecate `Faraday::Request#method`, replace it with `#http_method`.

## Todos
- [x] Tests
- [ ] Documentation

## Additional Notes

It requires for neighbor pull requests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,11 @@ jobs:
 
     - name: Rubocop
       run: |
-        gem install rubocop rubocop-performance --no-document
-        rubocop --require rubocop-performance --format progress
+        gem install bundler
+        bundle config set without 'development test'
+        bundle config set with 'lint'
+        bundle install
+        bundle exec rubocop --format progress
 
     - name: Yard-Junk
       run: |
--- a/Gemfile
+++ b/Gemfile
@@ -5,13 +5,18 @@ source 'https://rubygems.org'
 ruby RUBY_VERSION
 
 gem 'jruby-openssl', '~> 0.8.8', platforms: :jruby
-gem 'rake'
 
 group :development, :test do
   gem 'pry'
+  gem 'rake'
 end
 
-group :test do
+group :lint, :development do
+  gem 'rubocop', '~> 0.82.0'
+  gem 'rubocop-performance', '~> 1.0'
+end
+
+group :test, :development do
   gem 'coveralls', require: false
   gem 'em-http-request', '>= 1.1', require: 'em-http'
   gem 'em-synchrony', '>= 1.0.3', require: %w[em-synchrony em-synchrony/em-http]
@@ -24,7 +29,6 @@ group :test do
   gem 'rack-test', '>= 0.6', require: 'rack/test'
   gem 'rspec', '~> 3.7'
   gem 'rspec_junit_formatter', '~> 0.4'
-  gem 'rubocop-performance', '~> 1.0'
   gem 'simplecov'
   gem 'typhoeus', '~> 1.3',
       git: 'https://github.com/typhoeus/typhoeus.git',
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -12,7 +12,7 @@ module Faraday
   #     req.body = 'abc'
   #   end
   #
-  # @!attribute method
+  # @!attribute http_method
   #   @return [Symbol] the HTTP method of the Request
   # @!attribute path
   #   @return [URI, String] the path
@@ -26,7 +26,9 @@ module Faraday
   #   @return [RequestOptions] options
   #
   # rubocop:disable Style/StructInheritance
-  class Request < Struct.new(:method, :path, :params, :headers, :body, :options)
+  class Request < Struct.new(
+    :http_method, :path, :params, :headers, :body, :options
+  )
     # rubocop:enable Style/StructInheritance
 
     extend MiddlewareRegistry
@@ -54,6 +56,14 @@ module Faraday
       new(request_method).tap do |request|
         yield(request) if block_given?
       end
+    end
+
+    def method
+      warn <<~TEXT
+        WARNING: `Faraday::Request##{__method__}` is deprecated; use `#http_method` instead. It will be removed in or after version 2.0.
+        `Faraday::Request##{__method__}` called from #{caller_locations(1..1).first}
+      TEXT
+      http_method
     end
 
     # Replace params, preserving the existing hash type.
@@ -116,7 +126,7 @@ module Faraday
     # @return [Hash] the hash ready to be serialized in Marshal.
     def marshal_dump
       {
-        method: method,
+        http_method: http_method,
         body: body,
         headers: headers,
         path: path,
@@ -129,17 +139,17 @@ module Faraday
     # Restores the instance variables according to the +serialised+.
     # @param serialised [Hash] the serialised object.
     def marshal_load(serialised)
-      self.method  = serialised[:method]
-      self.body    = serialised[:body]
-      self.headers = serialised[:headers]
-      self.path    = serialised[:path]
-      self.params  = serialised[:params]
-      self.options = serialised[:options]
+      self.http_method = serialised[:http_method]
+      self.body        = serialised[:body]
+      self.headers     = serialised[:headers]
+      self.path        = serialised[:path]
+      self.params      = serialised[:params]
+      self.options     = serialised[:options]
     end
 
     # @return [Env] the Env for this Request
     def to_env(connection)
-      Env.new(method, body, connection.build_exclusive_url(path, params),
+      Env.new(http_method, body, connection.build_exclusive_url(path, params),
               options, headers, connection.ssl, connection.parallel_manager)
     end
   end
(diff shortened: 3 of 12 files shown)


lib/faraday/request.rb
```diff
@@ -26,7 +26,9 @@ module Faraday
   #   @return [RequestOptions] options
   #
   # rubocop:disable Style/StructInheritance
-  class Request < Struct.new(:method, :path, :params, :headers, :body, :options)
+  class Request < Struct.new(
+    :http_method, :path, :params, :headers, :body, :options
```
iMacTia, line 30: As far as I agree with the renaming, this is a breaking change in case someone is using or extending the `Request` struct, and thus it can't be merged. Was this change triggered for some specific reason?

Gemfile
```diff
@@ -11,6 +11,11 @@ group :development, :test do
   gem 'pry'
 end
 
+group :lint, :test do
+  gem 'rubocop', '~> 0.82.0'
```
iMacTia, line 15: I'd prefer allowing `rubocop` to update, as otherwise we'd leave it stuck on this version indefinitely

.github/workflows/ci.yml
```diff
@@ -25,8 +25,8 @@ jobs:
 
     - name: Rubocop
       run: |
-        gem install rubocop rubocop-performance --no-document
-        rubocop --require rubocop-performance --format progress
+        bundle install --without test
```
iMacTia, line 28: This will still install `jruby-openssl`, `rake` and `pry`. Can we install only the `lint` group?

.github/workflows/ci.yml
```diff
@@ -25,8 +25,8 @@ jobs:
...
+        bundle exec rubocop --format progress
```
iMacTia, line 29: If removed from here, should `rubocop-performance` be added to the `.rubocop.yml` file? See [doc](https://github.com/rubocop-hq/rubocop-performance#rubocop-configuration-file)

Gemfile
AlexWayfer: You can connect a service like https://dependabot.com/ or https://depfu.com/ and you'll receive pull requests with updates. Then contributors with pull requests not related to RuboCop and Ruby code at all, as me, will not get failed CI.

.github/workflows/ci.yml
AlexWayfer: No, unfortunately: https://github.com/rubygems/bundler-features/issues/59

.github/workflows/ci.yml
AlexWayfer: It's already there: https://github.com/lostisland/faraday/blob/d2d52e404aa4e4d0d78a4c0a1473d3b369023b22/.rubocop.yml#L4

lib/faraday/request.rb
AlexWayfer: I know that's breaking, but I think the earlier we break this — the better.

The root problem was in RuboCop offense:

```
Lint/StructNewOverride: `:method` member overrides `Struct#method` and it may be unexpected.	
```

We can disable this offense, but I agree that we should not re-define such core methods.

[`Object.method`](https://rubyapi.org/2.7/o/object#method-i-method)

iMacTia: Oh I see. Again, it totally makes sense and I'd like to do it as well, but I'm really worried about introducing a breaking change like this in a hotfix/minor release, it could backfire.

@olleolleolle what do you think? I'd add this to the v2 list of fixes and just disable the offence for this struct for the time being?

Gemfile
iMacTia: Sounds good 👍
Surprised we didn't do that sooner 😄
Opened #1150 for this

.github/workflows/ci.yml
iMacTia: The I'd keep the `gem install rubocop rubocop-performance --no-document`, unsure how much the extra gems will slow the build down. Particularly worried about `jruby-openssl` and `rake`

lib/faraday/request.rb
AlexWayfer: If you want to delay v2 with this change — we also can add a warning about deprecation, and working alias for now.

AlexWayfer: I've done this. Now, we have deprecated (but working) `Request#method`. I see no date for [`v2` milestone](https://github.com/lostisland/faraday/milestone/10), so I've set June 2020 (I can change it if you want).

I'm not sure about `method=`, I see no tests for it.

And there is warning during tests from result check (it suppressing when we check the output itself), and I think it's OK, but I can suppress it if you want (in tests of method result, and leave tests for the output itself).

.github/workflows/ci.yml
AlexWayfer: I've changed to `bundle-only`, [recommended by Bundler member](https://github.com/rubygems/bundler/pull/4907#issuecomment-241675023).

---

Oh, [it's broken](https://github.com/lostisland/faraday/pull/1149/checks?check_run_id=597784850) because of `:git` source for one of gems.

OK, I've changed to manually `with` and `without`. Via config, not deprecated flags:

```
[DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'development test'`, and stop using this flag
[DEPRECATED] The `--with` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set with 'lint'`, and stop using this flag
```

---

> This will still install `jruby-openssl`

No, only for JRuby environment. You can see that there is no installation of any `ssl` gem messages in RuboCop task. I can move it into some group, but I don't know in which one, it look generic.

> `rake` and `pry`

Moved into `development` group.

> Can we install only the lint group?

Now it's done.

> The I'd keep the `gem install rubocop rubocop-performance --no-document`

It's the bad approach! We lock `rubocop` versions in the `Gemfile`, and we need to install these specific version. And duplicate versions (like via `-v` flag for `gem install`) is very bad practice. So, I'm ready to resolve it in correct direction and speed up builds as you want.

---

Bundler flags does not work with bundled Bundler version, so we need to run `gem install bundler` (to update it).

lib/faraday/request.rb
olleolleolle: My take: Avoid dates - we just don't know when a release is happening.

AlexWayfer: > My take: Avoid dates - we just don't know when a release is happening.

It's not like "guaranteed", it's like "not earlier than". Do you still want to drop date?

olleolleolle: Readers will compare to any date given and may make interpretations, so I'm still with "no dates, but version numbers are alright".

AlexWayfer: My bad: I meant "exactly", not "guaranteed", because "not earlier than" is kind of guaranteed too.

---



> Readers will compare to any date given and may make interpretations, so I'm still with "no dates, but version numbers are alright".

I can agree, but these arguments are required: https://ruby-doc.org/stdlib-2.7.1/libdoc/rubygems/rdoc/Gem/Deprecate.html#method-c-deprecate

Do you want me to implement a custom warning?

olleolleolle: Ah, my whiny-sounds are powered by following the development of the code which is replacing that thing which used to have dates mentioned. A long and good exploration of how to tell users to stop using something which will be going away, definitely, in the future. There's nuance, and compatibility-keeping in those PRs. To walk backwards from the place it all ended at: https://github.com/rubygems/rubygems/issues/3512

So, yes, a small method of our own is probably easiest.

AlexWayfer: Thank you for explanation.

> So, yes, a small method of our own is probably easiest.

Done.

script/proxy-server
```diff
@@ -15,7 +15,7 @@ if (found = ARGV.index('-u'))
 end
 
 match_credentials = lambda { |credentials|
-  got_username, got_password = credentials.to_s.unpack('m*')[0].split(':', 2)
+  got_username, got_password = credentials.to_s.unpack1('m*')[0].split(':', 2)
```
olleolleolle, line 18: This seems like a behaviour change. 

https://rubyapi.org/2.7/o/string#method-i-unpack1

Contrast the example - return a list and return a scalar string. 

String#[] exists, so the got_password becomes a very short string.

AlexWayfer: Yes, thank you, my bad: I forgot to delete `[0]`. Done.

AlexWayfer:
> And sorry if I came across as a bit fussy

No problem.

> we really appreciate all the improvements you've added here!

Thank you, I see. :wink: And thank you for help and discussing.

> One last request (but not mandatory to merge, I can do it later if you don't have time), since we decided to rename `#method` into `#http_method` and deprecated the old method, it would be good to update the `UPGRADING.md` file as well to keep track of all backwards-incompatible changes.
> I think this is going to be the first v2.0 note!

I agree, added. If you want another content of this note — please, write, it's not too hard for me to change.

Issue #1152, opened 2020-05-01
Faraday::Adapter::NetHttpPersistent timeout tests failing with net-http-persistent 4.0.0


dgholz:
## Basic Info

* Faraday Version: master (87ecaf96822ef03ef6e86a9b9b57caa97e4c0fa8), also with v1.0.1
* Ruby Version: 2.7.0

## Issue description
Hello, I just checked out master & ran the tests, and I see all the timeout-related tests in `spec/faraday/adapter/net_http_persistent_spec.rb` failing with stacktraces like:
```
  1) Faraday::Adapter::NetHttpPersistent behaves like an adapter with SSL enabled #get behaves like a request method supports timeout option
     Failure/Error: expect { conn.public_send(http_method, '/') }.to raise_error(exc)
     
       expected Faraday::TimeoutError, got #<Faraday::ConnectionFailed wrapped=#<Net::OpenTimeout: execution expired>> with backtrace:
```

I see `net-http-persistent` just released `v4.0.0`; when I pinned `net-http-persistent` to `~> 3.0` in the `Gemfile` & `bundle install`, the tests pass as expected.

## Steps to reproduce
```
git clone git@github.com:lostisland/faraday.git
cd faraday
bundle install
bundle exec rspec ./spec/faraday/adapter/net_http_persistent_spec.rb
```

dgholz:
Here's all the failing tests:
```
rspec ./spec/faraday/adapter/net_http_persistent_spec.rb[1:1:1:2:1:10] # Faraday::Adapter::NetHttpPersistent behaves like an adapter with SSL enabled #get behaves like a request method supports timeout option
rspec ./spec/faraday/adapter/net_http_persistent_spec.rb[1:1:1:2:1:11] # Faraday::Adapter::NetHttpPersistent behaves like an adapter with SSL enabled #get behaves like a request method supports open_timeout option
rspec ./spec/faraday/adapter/net_http_persistent_spec.rb[1:1:1:1:1:10] # Faraday::Adapter::NetHttpPersistent behaves like an adapter with SSL enabled #delete behaves like a request method supports timeout option
rspec ./spec/faraday/adapter/net_http_persistent_spec.rb[1:1:1:1:1:11] # Faraday::Adapter::NetHttpPersistent behaves like an adapter with SSL enabled #delete behaves like a request method supports open_timeout option
[...]

PR #1153, opened 2020-05-01
Avoid 'last arg as keyword param' warning when building user middleware on Ruby 2.7

dgholz:
This change marks the methods which construct user middleware, so they are called with the 2.6 keyword argument behaviour in 2.7 (and avoid printing a warning about it).

Some users add their own middleware & write their own constructors with named parameters, and use keyword arguments when adding their middleware to the builder:

```
class MyMiddleware < Faraday::Middleware
  def initialize(app, my_argument:)
    @app = app
    @my_argument = my_argument
  end

[...]

connection = Faraday.new() do |configuration|
  configuration.use MyMiddleware, my_argument: 'hello'
end
```

This works great in 2.6 & 2.7, but 2.7 prints a deprecation warning:
```
/Users/danielholz/.gem/ruby/2.7.0/gems/faraday-1.0.1/lib/faraday/dependency_loader.rb:21: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/danielholz/code/some_project/lib/some_project/faraday_middleware.rb:2: warning: The called method `initialize' is defined here
```
The warning is a bit frustrating, since the calling code is not in the user's code, so they cannot add `**` to the call. The best option for them is to change their middleware to accept a Hash as a positional parameter & pass a Hash when adding their middleware (and checking for required parameters themselves).

I tried an alternative to `ruby2_keywords` by adding an explicit `**kwargs` argument to `Faraday::RackBuilder#use`, `Faraday::RackBuilder#use_symbol`, `Faraday::RackBuilder::Handler#initialize`, and `Faraday::DependencyLoader#new`, but it hit the explicit delegation of keyword arguments issue (https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/#why-deprecated) in 2.6.

### DIFF
--- a/spec/faraday/rack_builder_spec.rb
+++ b/spec/faraday/rack_builder_spec.rb
@@ -206,6 +206,9 @@ RSpec.describe Faraday::RackBuilder do
         end
       end
     end
+    let(:dog) do
+      subject.handlers.find { |handler| handler == dog_middleware }.build
+    end
 
     let(:cat_middleware) do
       Class.new(Faraday::Middleware) do
@@ -216,11 +219,22 @@ RSpec.describe Faraday::RackBuilder do
         end
       end
     end
+    let(:cat) do
+      subject.handlers.find { |handler| handler == cat_middleware }.build
+    end
+
     it 'adds a handler to construct middleware with options passed to use' do
       subject.use dog_middleware, 'Rex'
-      expect(subject.handlers.find { |handler| handler == dog_middleware }.build.name).to eq('Rex')
+      expect { dog }.to_not output(
+        /warning: Using the last argument as keyword parameters is deprecated/
+      ).to_stderr
+      expect(dog.name).to eq('Rex')
+
       subject.use cat_middleware, name: 'Felix'
-      expect(subject.handlers.find { |handler| handler == cat_middleware }.build.name).to eq('Felix')
+      expect { cat }.to_not output(
+        /warning: Using the last argument as keyword parameters is deprecated/
+      ).to_stderr
+      expect(cat.name).to eq('Felix')
     end
   end
 end
--- a/spec/faraday/rack_builder_spec.rb
+++ b/spec/faraday/rack_builder_spec.rb
@@ -194,13 +194,13 @@ RSpec.describe Faraday::RackBuilder do
     end
   end
 
-  context 'when middleware is added with arguments' do
+  context 'when middleware is added with named arguments' do
     let(:conn) { Faraday::Connection.new {} }
 
     let(:dog_middleware) do
       Class.new(Faraday::Middleware) do
         attr_accessor :name
-        def initialize(app, name)
+        def initialize(app, name:)
           super(app)
           @name = name
         end
@@ -210,31 +210,12 @@ RSpec.describe Faraday::RackBuilder do
       subject.handlers.find { |handler| handler == dog_middleware }.build
     end
 
-    let(:cat_middleware) do
-      Class.new(Faraday::Middleware) do
-        attr_accessor :name
-        def initialize(app, name:)
-          super(app)
-          @name = name
-        end
-      end
-    end
-    let(:cat) do
-      subject.handlers.find { |handler| handler == cat_middleware }.build
-    end
-
     it 'adds a handler to construct middleware with options passed to use' do
-      subject.use dog_middleware, 'Rex'
+      subject.use dog_middleware, name: 'Rex'
       expect { dog }.to_not output(
         /warning: Using the last argument as keyword parameters is deprecated/
       ).to_stderr
       expect(dog.name).to eq('Rex')
-
-      subject.use cat_middleware, name: 'Felix'
-      expect { cat }.to_not output(
-        /warning: Using the last argument as keyword parameters is deprecated/
-      ).to_stderr
-      expect(cat.name).to eq('Felix')
     end
   end
 end
--- a/spec/faraday/rack_builder_spec.rb
+++ b/spec/faraday/rack_builder_spec.rb
@@ -328,5 +328,14 @@ RSpec.describe Faraday::RackBuilder do
       ).to_stderr
       expect(rock.name).to eq('Rocky')
     end
+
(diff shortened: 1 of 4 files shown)


dgholz:
I'm not sure about the tests, I don't see a simple way to check if the warning was not emitted. I'm open to any suggestions about how to test this change more thoroughly!

dgholz:
I'm expecting the `Faraday::Adapter::NetHttpPersistent` tests to fail, due to #1152.

olleolleolle:
Thanks for working on this!

Splendid find.

There's a _very small_ gem which acts as a shim around the method: https://github.com/ruby/ruby2_keywords

We can start depending on that gem, to solve this in a compatible (and very widespread) way.

What do you think about that?

dgholz:
I think that sounds like a great idea! I was worried about adding a new gem dependency when the implementation was trivial, no problem for me to switch over to it.

spec/faraday/rack_builder_spec.rb
```diff
@@ -193,4 +193,29 @@ class Broken < Faraday::Middleware
...
+        @name = name
+      end
+    end
+    class Cat < Faraday::Middleware
+      attr_accessor :name
+      def initialize(app, name:)
+        super(app)
+        @name = name
+      end
+    end
```
olleolleolle, line 213: RSpec is powerful, we can inline these into not-accessible anywhere else variables, easily:


```suggestion
    let(:dog_middleware_class) do
      Class.new(Faraday::Middleware) do
        attr_accessor :name
        def initialize(app, name)
          super(app)
          @name = name
        end
      end
    end

    let(:cat_middleware_class) do
      Class.new(Faraday::Middleware) do
        attr_accessor :name
        def initialize(app, name:)
          super(app)
          @name = name
        end
      end
    end
```

And then use those in place of the `Dog` and `Cat`.

olleolleolle:
(Right, and then the new local variables need to be used, as well.)

dgholz:
Thanks for your patience!

olleolleolle:
I appreciate the PR and _your_ patience.

iMacTia:
The `net-http-persistent` issue should be fixed now, however I also noticed that this was previously attempted in #1106.
@dgholz did you notice that?

dgholz:
I did not notice that other PR! Thanks for the comment, I'll add `ruby2_keywords` to those other delegating functions and check for warnings in the test.

spec/faraday/rack_builder_spec.rb
```diff
@@ -193,4 +193,149 @@ class Broken < Faraday::Middleware
...
+          @name = name
+        end
+      end
+    end
+    let(:cat) do
+      subject.handlers.find { |handler| handler == cat_request }.build
+    end
+
+    it 'adds a handler to construct request adapter with options passed to request' do
+      skip 'still raising last argument kwarg deprecation warnings'
```
dgholz, line 239: I think I'm not declaring the Request middleware properly here, because he test still complains about keyword args even when all the methods have the `ruby2_keywords` prefix. I'd appreciate it if you could take a look and let me know how this should be set up for the test!

dgholz:
I removed the test case that exercised middleware that didn't use named parameters , since it worked fine in 2.7 & 2.7. I feel like the tests are a bit verbose and repetitive, and the last few look awfully similar to the `when having two handlers` test cases; but for my cases, I'm only worried about what the handler ends up building, and not at all about the order of the handlers.

spec/faraday/rack_builder_spec.rb
iMacTia: I think the problem here might be the `Faraday::Request` on line 226.
We don't have a specific base class for Request middleware like we have for response ones (`Faraday::Response::Middleware`), so `cat_request` should probably inherit from `Faraday::Middleware` instead.

iMacTia:
@dgholz sorry, it took me some time to find the possible cause by looking at the code.
I've left a comment that I believe MIGHT fix the issue.
If that still doesn't help, then please let me know and I'll probably have to checkout this locally and have a closer look.

spec/faraday/rack_builder_spec.rb
olleolleolle: It did work!

spec/faraday/rack_builder_spec.rb
```diff
@@ -193,4 +193,149 @@ class Broken < Faraday::Middleware
...
+      ).to_stderr
+      expect(dog.name).to eq('Rex')
+    end
+  end
+
+  context 'when a request adapter is added with named arguments' do
+    let(:conn) { Faraday::Connection.new {} }
+
+    let(:cat_request) do
+      Class.new(Faraday::Request) do
```
olleolleolle, line 226:
```suggestion
      Class.new(Faraday::Middleware) do
```

spec/faraday/rack_builder_spec.rb
olleolleolle, line 239:
```suggestion
```

olleolleolle:
I got excited by the fix, tried it locally, then applied it in a GitHub Suggestions commit, and this is ready to go, @iMacTia and @dgholz  - 🎉

iMacTia:
Thank you @dgholz for the effort! And thanks @olleolleolle for the last push 😉

grosser:
I'm not a fan of having another dependency or the added complexity.
- can this be removed since 2.7.2 removes these warnings ?
- can we either force hashes or keyword args to be used and then remove this ?

olleolleolle:
@grosser Hi! Glad for the feedback! 

We did this change in order to avoid warnings. If we can avoid warnings in some other way (such as those you listed), it'd be neat.

- Any existing-middleware-compatible improvement is welcome
- CI runs 2.7.2 already, so a removal could be attempted.
- I added a feature request: [Ruby 3 in the CI matrix](https://github.com/lostisland/faraday/issues/1189), to see if anything bad would happen with the upgrade.

grosser:
ruby 3 is a great idea so we don't run into unexpected errors of any new
hackery :)

PR #1154, opened 2020-05-04
NetHttpPersistent: ConnectionError on OpenTimeout

### LINKED ISSUE #1152: Faraday::Adapter::NetHttpPersistent timeout tests failing with net-http-persistent 4.0.0

olleolleolle:
## Description

This PR adds a special-case from the net-http adapter to the net-http-persistent adapter.

[open timeout](https://rubyapi.org/2.7/o/net/opentimeout) was treated as a failure to connect, but our tests required it to be a timeout. Turns out, that was a call which was necessary before 4.0.0 of net-http-persistent.

[There was a 4.0 released of net-http-timeout.](https://github.com/drbrain/net-http-persistent/blob/master/History.txt)

This change fixes the build by reraising as ConnectionError, just like in Faraday's `net-http` adapter.

Fixes #1152 by making the build pass.

## Questions for reviewer

- The 4.0.0 release requires Ruby 2.5 (I think that it mentions the 2.4 going out of support as a reason to be able to let go of the exception handling it had in itself)
- This is a **breaking change** for this adapter

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,7 +23,7 @@ group :test, :development do
   gem 'excon', '>= 0.27.4'
   gem 'httpclient', '>= 2.2'
   gem 'multipart-parser'
-  gem 'net-http-persistent'
+  gem 'net-http-persistent', '~> 4.0'
   gem 'patron', '>= 0.4.2', platforms: :ruby
   gem 'rack', '< 2.1'
   gem 'rack-test', '>= 0.6', require: 'rack/test'
--- a/lib/faraday/adapter/net_http_persistent.rb
+++ b/lib/faraday/adapter/net_http_persistent.rb
@@ -53,6 +53,8 @@ module Faraday
         http.request env[:url], create_request(env)
       rescue Errno::ETIMEDOUT => e
         raise Faraday::TimeoutError, e
+      rescue Net::OpenTimeout => e
+        raise Faraday::ConnectionFailed, e
       rescue Net::HTTP::Persistent::Error => e
         raise Faraday::TimeoutError, e if e.message.include? 'Timeout'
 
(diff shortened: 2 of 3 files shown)


olleolleolle:
@iMacTia Does this rhyme with how the other adapters handle open timeout?

olleolleolle:
I took a look at the history, and learned that the net-http adapter treats it as a ConnectionError on purpose: https://github.com/lostisland/faraday/commit/9ad0fcd38a27e51fdbaaa034d00a54cb26d59a9c

olleolleolle:
Learning more: There was a recent release of net-http-persistent, which got in a new change. https://github.com/drbrain/net-http-persistent/commit/2d236224fa797629dd80831148b3cbaa645b5d9b

olleolleolle:
Yes, to make this work for < 3.x + 3.x + 4.x, we need a new conditional in the "rescue and re-raise" section which this PR changed (and its tests).

iMacTia:
I THINK this is now fixed in #1226, although to keep backwards-compatibility I've kept `Faraday::Timeout` as the exception instead of `Faraday::ConnectionFailed`

PR #1155, opened 2020-05-05
Add 1.0 release to wording in CONTRIBUTING

olleolleolle:
## Description

This PR updates the CONTRIBUTING document to **reflect the fact that v1.0 has been released**.

**Unrelated:** One thing that stood out to me was the policy of accepting new middleware. That seems like an outdated policy?

I came across this while preparing to copy content to the faraday-http project, to support https://github.com/lostisland/faraday-http/pull/14

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -34,22 +34,17 @@ When adding a feature in Faraday:
 the green light by the core team start working on the PR.
 
 
-### New Middleware
+### New Middleware & Adapters
 
-We will accept middleware that:
+We prefer new adapters and middlewares to be added as separate gems. We can link to such gems from this project.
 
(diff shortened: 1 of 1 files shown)


iMacTia:
> One thing that stood out to me was the policy of accepting new middleware. That seems like an outdated policy?

Yes, agree. We definitely don't want to encourage adding middleware to `faraday` or `faraday_middleware` anymore. I suggest we delete that completely or we explicitly say that we prefer new adapters/middleware to be added as separate gems.

Maybe we should think of some sort of review process and consequent addition to an official "middleware/adapters marketplace"? You know, to give these external projects some visibility

olleolleolle:
@iMacTia I changed some wording. This is an improvement, at least.

PR #1162, opened 2020-06-30
Refactor parameter sorting

wishdev:
## Description
This is a quasi-second attempt at #469 which respects the comment in it https://github.com/lostisland/faraday/pull/469#issuecomment-81458828

As opposed to adding any new options - this simply moves the sort call into a method which can be overridden as opposed to requiring the "cut and paste" of the standard modules into a custom module.

So with this patch one can accomplish no param sorting by simply adding

```
module Faraday
  module NestedParamsEncoder
    def self.sort_params(_); end
  end

  module FlatParamsEncoder
    def self.sort_params(_); end
  end
end
```

To their code, because sort_params can be overridden.

Again, I completely respect the desire to keep it clean - I'm just wondering if this slight refactor would be ok.

### DIFF
--- a/lib/faraday/encoders/flat_params_encoder.rb
+++ b/lib/faraday/encoders/flat_params_encoder.rb
@@ -35,7 +35,7 @@ module Faraday
         end
         # Useful default for OAuth and caching.
         # Only to be used for non-Array inputs. Arrays should preserve order.
-        params.sort!
+        params.sort! if @sort_params
       end
 
       # The params have form [['key1', 'value1'], ['key2', 'value2']].
@@ -94,5 +94,11 @@ module Faraday
         end
       end
     end
+
+    class << self
+      attr_accessor :sort_params
+    end
   end
+
+  FlatParamsEncoder.sort_params = true
 end
--- a/lib/faraday/encoders/flat_params_encoder.rb
+++ b/lib/faraday/encoders/flat_params_encoder.rb
@@ -33,7 +33,7 @@ module Faraday
           key = key.to_s if key.is_a?(Symbol)
           [key, value]
         end
-        # Useful default for OAuth and caching.
+
         # Only to be used for non-Array inputs. Arrays should preserve order.
         params.sort! if @sort_params
       end
@@ -98,7 +98,8 @@ module Faraday
     class << self
       attr_accessor :sort_params
     end
-  end
 
-  FlatParamsEncoder.sort_params = true
+    # Useful default for OAuth and caching.
+    @sort_params = true
+  end
 end
--- a/lib/faraday/encoders/nested_params_encoder.rb
+++ b/lib/faraday/encoders/nested_params_encoder.rb
@@ -21,7 +21,7 @@ module Faraday
           key = key.to_s if key.is_a?(Symbol)
           [key, value]
         end
-        # Useful default for OAuth and caching.
+
         # Only to be used for non-Array inputs. Arrays should preserve order.
         params.sort! if @sort_params
       end
@@ -167,9 +167,10 @@ module Faraday
       def_delegators :'Faraday::Utils', :escape, :unescape
     end
 
+    # Useful default for OAuth and caching.
+    @sort_params = true
+
     extend EncodeMethods
     extend DecodeMethods
   end
-
-  NestedParamsEncoder.sort_params = true
 end
--- a/lib/faraday/encoders/nested_params_encoder.rb
+++ b/lib/faraday/encoders/nested_params_encoder.rb
@@ -23,7 +23,7 @@ module Faraday
         end
         # Useful default for OAuth and caching.
         # Only to be used for non-Array inputs. Arrays should preserve order.
-        params.sort!
+        params.sort! if @sort_params
       end
 
       # The params have form [['key1', 'value1'], ['key2', 'value2']].
@@ -161,6 +161,8 @@ module Faraday
   # for your requests.
   module NestedParamsEncoder
     class << self
+      attr_accessor :sort_params
+
       extend Forwardable
       def_delegators :'Faraday::Utils', :escape, :unescape
     end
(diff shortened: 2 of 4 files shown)


iMacTia:
Hey @wishdev, appreciate you trying to fit this in following the first feedback in the other PR.
I think I can relate to Mislav's comment, it's really easy to add small tweaks into your gem, until over time the number of tweaks grows beyond control.

The biggest issue I have with your change is that you're changing the default behaviour of the gem.
This might fit your requirements, but may potentially break other existing implementations.
For that reason, we would need to change your PR so that `sort_params` actually does not do any sorting, and you'd have the opportunity to override it in your project, in order to make it sort them for real.
As you can imagine this would be quite dumb (a method called sort_params that doesn't sort 😄).
Moreover, you'd still need to override this new method in your project.

Luckily, Ruby is an OOP language and as such it should already give you all the tools you need, so let me show you what I think it's the best solution in your case and, most probably, what Mislav was also referring to in his comment:

```ruby
class MyEncoder < Faraday::NestedParamsEncoder # or FlatParamsEncoder
  self.encode(params)
    super
    params.sort!
  end
end

# Then later when you initialise your Faraday Connection
conn = Faraday.new(url, request: { params_encoder: MyEncoder })
```

There you go, no change necessary in the Faraday gem itself, and just 6 lines of code in your project.
I haven't tested the code above, but I'm sure that will give you the idea of what I mean.

I'm closing this PR in the hope that the above helps you with what you're trying to do, but if you have any questions then feel free to follow-up.

wishdev:
I apologize but you misread the code. The current implementation sorts. I do not want a sort and therefore seek the change to allow the sort_params addition to be a passthrough.

There is no change at all to the default and current implemenation. This is simple a decomposition of the encode method to allow for a more granular override capability.

iMacTia:
I'm sorry @wishdev, I most definitely did!

To my surprise, we currently sort parameters indeed. I was unaware of that bit but it seems like it has been like that over the past 8 years 😮.

Based on everything I said above, your approach makes sense to me and it will allow devs to override that opinionated behaviour when necessary

iMacTia:
I was curious to see if tests were failing, but there seems to be an issue with one of the dependencies, I'll try to have a look when I get a chance

wishdev:
> ```ruby
> # The same can be done for Faraday::NestedParamsEncoder
> Faraday::FlatParamsEncoder.sort_params = false # default value is true
> ```
> 
> The change would be limited to the encoders (thus not touching the connection) and changing the behaviour would be as easy as to set the new option to `false`, with no need to override methods.
> What do you think of this revised, middle-way proposal @wishdev?

I believe it is an excellent idea @iMacTia. I started with the add nothing concept and I like middle grounds. I've pushed the change as you requested along with specs.

iMacTia:
@wishdev I did some digging about the Rubocop issue and I found this [comment](https://github.com/rubocop-hq/rubocop/issues/3471#issuecomment-244558731).
The reason seems to be that your lines `FlatParamsEncoder.sort_params = true` and `NestedParamsEncoder.sort_params = true` are inside the `module Faraday` block, causing Rubocop not to consider it a namespace (as it should be).

May I ask you one last change to your PR by moving those lines outside the `module Faraday` block?
That should fix the Rubocop issues 👍

wishdev:
My apologies for not understanding what the rubocop issue was concerning.

I believe this should move things along.

sergeymavrin:
Sorry for my nubian question, but how I can use option sort_params = false with such connection description?
```
@connection = Faraday.new(url: cfg['http_url']) do |req|
  req.request :url_encoded
  req.headers['Content-Type'] = 'application/x-www-form-urlencoded'
  req.adapter Faraday.default_adapter
end
```

iMacTia:
@sergeymavrin you simply need to set the `sort_params` property in the params encoder you're using.
If you're unsure which one you're using, it's probably the default one (`NestedParamsEncoder`), or you could use the `Faraday::Utils` helper to find out:

```ruby
# If you know which encored you're using
Faraday::NestedParamsEncoder.sort_params = false
# or
Faraday::FlatParamsEncoder.sort_params = false

# If you want to use the Faraday::Utils helper
Faraday::Utils.default_params_encoder.sort_params = false
```

sergeymavrin:
@iMacTia Thank You

davidbasalla:
I've run into a situation where I need to set `sort_params = false` but I've got multiple Faraday clients in my app. What's the best way of disabling sorting for a specific Faraday client? Iiuc `@sort_params` is a class variable (?) on the `FlatParamsEncoder` module, so running `Faraday::NestedParamsEncoder.sort_params = false` would change it for _all_ Faraday clients in my app? Can I subclass/include `FlatParamsEncoder` somehow and then set `@sort_params` to false?

iMacTia:
@davidbasalla mmmh I wish there was a nicer way to do this, but you're right, this is a class variable setting right now with no way to set it. The whole API of the param encoders are very "classy" right now (mostly class methods), so changing this on a per-client basis is not straightforward.

A possible way around this, as you already pointed out, is to subclass/include the `ParamsEncoder` you wish to use, and then pass it to a specific client instance as shown [here](paramshttps://lostisland.github.io/faraday/usage/customize#changing-how-parameters-are-serialized). However this solution requires you to have access to the initilization configuration of your connection, which may not be the case for you.

If that's still a problem, then things might get messy. I'd be happy to assist further but I'd suggest to move this into a separate issue so we can brainstorm the best way forward 👍

davidbasalla:
thanks for your reply @iMacTia. For the moment I've resorted to cloning the `FlatParamsEncoder` module on start up of my app and changing `sort_params` on the clone (which looks a bit non-standard but seems to work)

```
module Faraday
  UnsortedFlatParamsEncoder = FlatParamsEncoder.clone
  UnsortedFlatParamsEncoder.sort_params = false
end
```

It seems to solve the problem for me. I'll create a separate issue if I run into any more problems with it 👍

PR #1167, opened 2020-07-16
Trivial refactor: Range is guaranteed to be immutable

ujihisa:
* https://github.com/ruby/ruby/blob/8900a25581822759daca528d46a75e0b743fc22e/range.c#L76

## Description

Refactor internal implementation details to be less confusing and easier to maintain for future by removing unnecessary code and simplifying.

## Todos

N/A

## Additional Notes

According to the git blame, it seems to have been added when rubocop suggested don't use mutable values for constant. Since these values had been already immutable, we should have kept that unchanged actually :)

### DIFF
--- a/lib/faraday/response/raise_error.rb
+++ b/lib/faraday/response/raise_error.rb
@@ -6,8 +6,8 @@ module Faraday
     # client or server error responses.
     class RaiseError < Middleware
       # rubocop:disable Naming/ConstantName
-      ClientErrorStatuses = (400...500).freeze
-      ServerErrorStatuses = (500...600).freeze
+      ClientErrorStatuses = (400...500)
+      ServerErrorStatuses = (500...600)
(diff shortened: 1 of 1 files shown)


olleolleolle:
Thanks for taking a look at this!

The issue is not with Faraday, though, it's upstream, with the linting tool, and the report at: https://github.com/rubocop-hq/rubocop/issues/7192 discusses the details of this.

If and when that tool changes, we can update this code. (I was unable to see anything special in the list of releases since 0.84.0 which ran in CI. https://github.com/rubocop-hq/rubocop/releases - do you happen to know a version where this PR would pass? Can you update to that version, somehow?)

ujihisa:
That's true. In application level Rubyists can assume Range objects are immutable, but in library level we can't. I'm dropping my proposal :+1:

PR #1172, opened 2020-07-27
Handle verify_hostname in SSLOption

cosmo0920:
Signed-off-by: Hiroshi Hatake <cosmo0920.oucc@gmail.com>

## Description
This had been introduced in CRuby's openssl v2.0.0.
And it has been bundled since CRuby 2.4.

ref: https://docs.ruby-lang.org/en/2.4.0/NEWS.html

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Handle verify_hostname at least net/http and excon adapters 

## Additional Notes

fluent-plugin-elasticserach wants to handle verify_hostname option to handle wildcard certificate. But, without this option, ES plugin complains `does not match the server certificate (OpenSSL::SSL::SSLError)` error and fills up these errors in log. See: https://github.com/uken/fluent-plugin-elasticsearch/issues/770#issuecomment-658807393

### DIFF
--- a/spec/faraday/options/env_spec.rb
+++ b/spec/faraday/options/env_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe Faraday::Env do
   it 'handle verify_hostname when fetching' do
     ssl = Faraday::SSLOptions.new
     ssl.verify_hostname = true
-    expect(ssl.fetch(:verify_hostname, true)).to be_truthy
+    expect(ssl.fetch(:verify_hostname, false)).to be_truthy
   end
 
   it 'retains custom members' do
--- a/spec/faraday/options/env_spec.rb
+++ b/spec/faraday/options/env_spec.rb
@@ -27,6 +27,12 @@ RSpec.describe Faraday::Env do
     expect(ssl.fetch(:verify, true)).to be_falsey
   end
 
+  it 'handle verify_hostname when fetching' do
+    ssl = Faraday::SSLOptions.new
+    ssl.verify_hostname = true
+    expect(ssl.fetch(:verify_hostname, true)).to be_truthy
+  end
+
   it 'retains custom members' do
     env[:foo] = 'custom 1'
     env[:bar] = :custom_2
--- a/lib/faraday/adapter/excon.rb
+++ b/lib/faraday/adapter/excon.rb
@@ -48,6 +48,11 @@ module Faraday
         env[:body].respond_to?(:read) ? env[:body].read : env[:body]
       end
 
+      # To be public for tests.
+      def handles_ssl_verify_hostname?
+        Gem::Version.new(::Excon::VERSION) >= Gem::Version.new('0.76.0')
+      end
+
       private
 
       def opts_from_env(env)
@@ -83,6 +88,10 @@ module Faraday
         # https://github.com/geemus/excon/issues/106
         # https://github.com/jruby/jruby-ossl/issues/19
         opts[:nonblock] = false
+        # https://github.com/excon/excon/pull/722
+        if handles_ssl_verify_hostname?
+          opts[:ssl_verify_hostname] = !!ssl.fetch(:verify_hostname, false)
+        end
 
         OPTS_KEYS.each do |(key_in_opts, key_in_ssl)|
           next unless ssl[key_in_ssl]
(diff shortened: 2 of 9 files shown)


spec/faraday/options/env_spec.rb
```diff
@@ -27,6 +27,12 @@
     expect(ssl.fetch(:verify, true)).to be_falsey
   end
 
+  it 'handle verify_hostname when fetching' do
+    ssl = Faraday::SSLOptions.new
+    ssl.verify_hostname = true
+    expect(ssl.fetch(:verify, true)).to be_truthy
```
iMacTia, line 33: This should probably be `expect(ssl.fetch(:verify_hostname, false)).to be_truthy`.
At the moment this test passes because `:verify` is not set and you default the fetch to `true`.

cosmo0920: Ahhh... Thanks for pointing out silly mistake.

cosmo0920:
Thanks for the feedback. I'm continue to work for supporting `verify_hostname` option for SSLOption at least `net/http` and `excon` adapters. But, this option is quite new. Other HTTP backends seem not to be support for now.

iMacTia:
That's OK @cosmo0920, if a client doesn't support it then there's nothing you can do, but please add it to as many as possible 👍 
If `Net::HTTP` supports it then most probably [`Net::HTTP::Persistent`](https://github.com/lostisland/faraday/blob/master/lib/faraday/adapter/net_http_persistent.rb) does as well?

spec/faraday/options/env_spec.rb
```diff
@@ -27,6 +27,12 @@
     expect(ssl.fetch(:verify, true)).to be_falsey
   end
 
+  it 'handle verify_hostname when fetching' do
+    ssl = Faraday::SSLOptions.new
+    ssl.verify_hostname = true
+    expect(ssl.fetch(:verify_hostname, true)).to be_truthy
```
iMacTia, line 33: Sorry maybe I wasn't entirely clear in my previous comment.
I can see you fixed the property name, which is great, but it's not a good idea to test `be_truthy` when you default the `fetch` to `true`, as this might cause false-positives.

If you check the example above (for `ssl.verify`), the value is set to false and the condition is `be_falsey`. You could do that here as well, or if you prefer to keep the `true` value then please change the `fetch` to default to `false`.

cosmo0920: Oops. I got it.

cosmo0920:
> If Net::HTTP supports it then most probably Net::HTTP::Persistent does as well?

Yes. We can support this. The patch is here:
https://github.com/drbrain/net-http-persistent/pull/120

cosmo0920:
I'd added handling `verify_hostname` on excon adapter and sent a PR to handle `verify_hostname` on SSL context into net-http-persistent repository: drbrain/net-http-persistent#120

I'm waiting to be released patched version of net-http-persistent gem.

donv:
Hi!

We would like to get this merged.  What is the next step?  How can we help?

iMacTia:
Thank you @donv !

This PR was opened when Faraday bundled all the adapters together, so it will require some rebasing:
1. Remove changes done to `lib/faraday/adapter/excon.rb` and `lib/faraday/adapter/net_http.rb` (but save them for later!)
2. Rebase against main
3. Specs added to `spec/faraday/adapter/excon_spec.rb` and `spec/faraday/adapter/net_http_spec.rb` should be moved into the generic `spec/support/shared_examples/request_method.rb`, and wrapped into a new `on_feature` block ([see example](https://github.com/lostisland/faraday/blob/88d16b7e2d2a033fb344b053547625c95661b85c/spec/support/shared_examples/request_method.rb#L54)). This is because not all adapters will support this, so the test should only run when explicitly enabled.

Once this is done, the work should then move over to the already supported adapter gems: [faraday-net_http](https://github.com/lostisland/faraday-net_http) and [faraday-excon](https://github.com/lostisland/faraday-excon). For both these gems we should:
* Open a new PR that introduces the changes originally implemented on the adapters in this PR (see changes saved for later in point 1 above ☝️)
* Enable the feature introduced in point 3 above (see [list of features](https://github.com/lostisland/faraday-net_http/blob/main/spec/faraday/adapter/net_http_spec.rb#L4))

This may be a lot to take in, I'm happy to provide additional clarity if there's any confusion.
Also, I appreciate you might only be looking at making this work with `net_http`, so it's OK if you want to postpone the work on `faraday-excon` to later

iMacTia:
This was done in #1428 and released in Faraday v2.4.0 🎉 
Thank you @kazarin for the contribution 🙏

PR #1181, opened 2020-09-16
Include request info in exceptions raised by RaiseError Middleware

SandroDamilano:
## Description
When using the `RaiseError` middleware, the error raised contains only information about the response. This should not be a problem if we had just made a single request, or if we caught the error close to where it was raised, so that we can know for sure what the problematic request was. But things can get more tricky if the request was made along with other ones, or as part of a batch process.
When using an error tracker, it's very important to track as much information as we can about the error, so having information about the request as well as the response is critical.
Thus, considering that it would also be very easy for the middleware to include that information, I think it would be a very handy addition.

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -38,6 +38,14 @@ module Faraday
     #              :headers - String key/value hash of HTTP response header
     #                         values.
     #              :body    - Optional string HTTP response body.
+    #              :request - Hash
+    #                           :method   - Symbol with the request HTTP method.
+    #                           :url_path - String with the url path requested.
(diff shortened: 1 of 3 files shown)


iMacTia:
@SandroDamilano thanks for opening this PR. I agree in principle with the changes, however I noticed Rubocop linting is currently failing.
That's not your fault though and I've already fixed it in #1182 
Could you please pull/rebase from master and update your PR?

SandroDamilano:
@iMacTia Oh, that's great! I'm on it

iMacTia (approved):

PR #1185, opened 2020-10-01
Remove retry_change_requests from documentation

stim371:
## Description
The net-http-persistent gem removed the #retry_change_requests adapter configuration with https://github.com/drbrain/net-http-persistent/pull/100 in v4.0.0.


olleolleolle (commented):
This sounds good to do. 

If people are stuck on old versions, that could trip them up, but of course, the versioning of that gem is its issue, not ours.

iMacTia (approved):
Sorry for being late to the party.
No objections from me, that was just an example anyway to show how to customise adapter-specific configs, so happy to remove it 👍

PR #1183, opened 2020-09-18
Drop `git ls-files` in gemspec

utkarsh2102:
Hi @iMacTia,

Thanks for your work on this!

Whilst maintaining this in Debian, we found that this library relies on `git` to produce a list of files. Using git in gemspec files in problematic, in general.

It also adds the Packaging extension of RuboCop.
More about it can be found on https://docs.rubocop.org/rubocop-packaging/

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>

PR #1190, opened 2020-10-20
Do not package examples and specs in builds

nbibler:
## Description
It's common practice to only include production code in Rubygem builds. The default behavior of Bundler, for example, is to [ignore "test|spec|features"](https://github.com/rubygems/rubygems/blob/61447f3f1b19390a128d0239d0cbea250e265ae6/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt#L28) when capturing files to include in the builds.

After the #1183 PR in the 1.1.0 release, I noticed this was not being done – not that the PR changed the previous behavior, mind you – and that test files are being unnecessarily included in the released builds.

This PR removes examples and spec files from the Rubygems build.

### DIFF
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'multipart-post', '>= 1.2', '< 3'
   spec.add_dependency 'ruby2_keywords'
 
-  spec.files = Dir['CHANGELOG.md', '{examples,lib,spec}/**/*', 'LICENSE.md', 'Rakefile', 'README.md']
+  spec.files = Dir['CHANGELOG.md', 'lib/**/*', 'LICENSE.md', 'Rakefile', 'README.md']
   spec.require_paths = %w[lib spec/external_adapters]
   spec.metadata = {
     'homepage_uri' => 'https://lostisland.github.io/faraday',


olleolleolle:
The reason it now lives in the gem: the adapter test suite is borrowed by gems implementing adapters. Compare the [faraday-http](https://github.com/lostisland/faraday-http/) gem.

Perhaps there are other great ways of getting that.

iMacTia:
@nbibler thanks for raising this. Under normal circumstances, you'd be totally right and I'd completely agree with you.
In our case, as @olleolleolle already mentioned, this is actually a feature to allow adapters to share tests with Faraday and ensure they are "compliant" with its rules. You can see an example in the link he provided.

Indeed there's no way you could know that because we haven't properly documented this thing 🙈, so totally not your fault.
I'll consider adding a comment or a section in the README to clarify this.

I'm going to close this PR for now, but if you think there's a better way to achieve the same result then we're more than happy to listen 👍!

nbibler:
Not a problem. I just saw the gemspec and it flagged something in my head as a problem with a quick fix. If it's there by design, then I have no issue with it.

@iMacTia: If you're planning on documenting it, I'd probably suggest an inline comment around that line in the gemspec. It wouldn't cross my mind to check a README for why there's an odd behavior in a gemspec file. Something more localized to the "that's weird, but on purpose" makes more sense to me.

iMacTia:
Totally agree, thanks for the feedback and for caring about Faraday ❤️

PR #1193, opened 2020-10-21
docs: Adapters, how to create them

olleolleolle:
## Description

In order to explain what it takes to make a Faraday adapter, we're adding some guide text to the page on Adapters.

It's been a while since we added the website, but now could be the time to add a little more explanation.

## Todos
- [ ] Proof-reading
- [ ] Think about sections, such as the in_parallel section, is that good to be inline, or should it land in its own article?

## Additional Notes

Please, add more notes as comments on this PR, we need it.

### DIFF
--- a/docs/adapters/index.md
+++ b/docs/adapters/index.md
@@ -52,7 +52,14 @@ exist. Its only function is to make this example more concrete.
 
 ### An Adapter _is_ a Middleware
 
-There are only two things which are actually mandatory for an adapter middleware to function:
+When you subclass `::Faraday::Adapter`, you get helpful methods defined:
+
+```ruby
+class FlorpHttp < ::Faraday::Adapter
+end
+```
+
+Now, there are only two things which are actually mandatory for an adapter middleware to function:
 
 - a `#call` implementation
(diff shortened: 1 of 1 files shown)


docs/adapters/index.md
```diff
@@ -43,6 +43,136 @@ However, sometimes you need to access a feature specific to one of the adapters
...
+There are only two things which are actually mandatory for an adapter middleware to function:
+
+- a `#call` implementation
+- a call to `#save_response` inside `#call`, which will keep the Response around.
+
+These are the only two things.
+
+The rest of this text is about methods which make the authoring easier.
+
+### Helpful class: `::Faraday::Adapter`
```
iMacTia, line 64: Subclassing `Faraday::Adapter` is also necessary to access `save_response` and to be recognised as an adapter by the rack builder, so I'd make that a requirement.

iMacTia (commented):
Great work @olleolleolle 👏 
I've just added a small comment and I agree with you that the parallel manager can probably be moved into another page as it's quite an advanced and long topic.

PR #1194, opened 2020-10-22
Introduces on_request and on_complete methods in Faraday::Middleware.

iMacTia:
## Description
* Introduces `on_request` and `on_complete` methods in `Faraday::Middleware`
* Adds documentation on how to leverage the new methods

## Additional Notes
Unit tests for `on_complete` were a bit complicated to write as they rely on the callback system of the `Rack::Builder`.
It would be better to write integration tests for that, but given the limited amount of time I couldn't make it today.

### DIFF
--- a/docs/middleware/custom.md
+++ b/docs/middleware/custom.md
@@ -9,8 +9,7 @@ prev_name: Available Middleware
 prev_link: ./list
 ---
 
-Middleware are classes that implement a `call` instance method. They hook into
-the request/response cycle.
+Middleware are classes that implement a `call` instance method. They hook into the request/response cycle.
 
 ```ruby
 def call(request_env)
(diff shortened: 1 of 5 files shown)


docs/middleware/custom.md
```diff
@@ -43,3 +42,16 @@ later, response. Some keys are:
...
 ```
+
+### Faraday::Middleware
+
+There's an easier way to write middleware, and it's also the recommended one: make your middleware subclass `Faraday::Middleware`.
+`Faraday::Middleware` already implements the `call` method for you and looks for two methods in your subclass: `on_request(env) and `on_complete(env)`.
+`on_request` is called when the request is being built and is given the `env` representing the request.
+`on_complete` is called after the response has been received (that's right, it already supports parallel mode!) and receives the `env` of the response.
+
+### Do I need to override `call`?
```
olleolleolle, line 53: Minor: `#call`, `#on_request` etc could be names.

olleolleolle (approved):
🎂

docs/middleware/custom.md
iMacTia: yeah I like that, it makes it clear that they're method unlike `env` 👍

docs/middleware/custom.md
```diff
@@ -43,3 +42,16 @@ later, response. Some keys are:
...
+### Faraday::Middleware
+
+There's an easier way to write middleware, and it's also the recommended one: make your middleware subclass `Faraday::Middleware`.
+`Faraday::Middleware` already implements the `#call` method for you and looks for two methods in your subclass: `#on_request(env) and `#on_complete(env)`.
+`#on_request` is called when the request is being built and is given the `env` representing the request.
+`#on_complete` is called after the response has been received (that's right, it already supports parallel mode!) and receives the `env` of the response.
+
+### Do I need to override `#call`?
+
+For the majority of middleware, it's not necessary to override the `#call` method, as you can simply use `#on_request` and `#on_response`.
```
natematykiewicz, line 55: This on_response should be on_complete, right?

iMacTia: YES 🙈

iMacTia: Luckily, that was already spot and fixed 🙌 
https://github.com/lostisland/faraday/commit/4e9f4b06f1a1d46fac45663ab3a857aa60b79012

natematykiewicz: Oh nice! I was just reading through the PRs referenced in the release notes. I didn't think to diff the two versions. 😬

PR #1195, opened 2020-10-23
Replace the Net::HTTP adapter by a separate gem

JanDintel:
## Description

The Net::HTTP adapter is extracted into its own gem. The idea behind this change is to create an ecosystem of different packages that you can use with Faraday, so that Faraday doesn’t have to ship with the packages itself.

For now all the Net::HTTP tests are still in place to verify that everything is working correctly.

Link to related issues if any. (As `Fixes #XXX`)

## Todos
- [ ] Tests
- [ ] Documentation

## Additional Notes
Optional section


olleolleolle:
Guessing the implicit begin rescue end trio isn't available in the failing ruby version?

JanDintel:
@olleolleolle it was indeed the implicit begin rescue end, everything is green now.

@olleolleolle and @iMacTia is this what you had in mind regarding the componentisation of the adapters?

iMacTia:
Hi @JanDintel, apologies for the late reply!
YES, this is indeed what we were thinking about 😃.
The only thing we need to make sure of is that after this change existing code will still be working as expected.
I saw you didn't remove any `net_http` adapter tests, so I guess those are all still passing fine?

JanDintel:
Replaced by #1222

PR #1198, opened 2020-10-26
Adjust the retry documentation and spec to align with implementation

### LINKED ISSUE #1197: Retry documentation or implementation looks incorrect?
The retry documentation shows [an example](https://github.com/lostisland/faraday/blob/master/docs/middleware/request/retry.md#example-usage), which mentions that the retry delay range for the first retry is between 0.05 and 0.075, and the second retry is between 0.1 and 0.15. The second range is not correct, based on the [current implementation](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/retry.rb#L227).

The implementation calculates a `current_interval` and a `random_interval`, then adds them. The example uses a configuration with config values of max at 2, backoff_factor at 2, interval at 0.05, and interval_randomness at 0.5. For the first retry, the `current_interval` will be 0.05 (`0.05 * 2 ** 0`) and the `random_interval` will be a maximum of 0.025 (`1 * 0.5 * 0.05`). For the second retry, the `current_interval` will be 0.1 (`0.05 * 2 ** 1`) and the `random_interval` will be a maximum of 0.025 (`1 * 0.5 * 0.05`), again. The `random_interval` calculation never changes -- it always uses the configured interval value, not the calculated interval value. As such, the documented example is incorrect -- the range for the second retry would be 0.1 and 0.125.

I can quickly submit a PR to correct the documentation but wanted to check that it's just the documentation that's incorrect and not the implementation. The documentation seems to indicate that perhaps the algorithm was supposed to calculate the random_interval based on the current_interval.

nbeyer:
## Description
Fixes #1197 

The documentation on the range for the second retry iteration is a bit off and the spec related to the retry is as well. This PR corrects that.

### DIFF
--- a/docs/middleware/request/retry.md
+++ b/docs/middleware/request/retry.md
@@ -21,7 +21,7 @@ The middleware can also handle the [`Retry-After`](https://developer.mozilla.org
 ### Example Usage
 
 This example will result in a first interval that is random between 0.05 and 0.075
-and a second interval that is random between 0.1 and 0.15.
+and a second interval that is random between 0.1 and 0.125.
 
 ```ruby
 retry_options = {
--- a/lib/faraday/request/retry.rb
+++ b/lib/faraday/request/retry.rb
@@ -21,7 +21,7 @@ module Faraday
     #   end
     #
     # This example will result in a first interval that is random between 0.05
-    # and 0.075 and a second interval that is random between 0.1 and 0.15.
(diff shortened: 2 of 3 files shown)


olleolleolle (approved):
Thanks for the swift fix and good analysis!

This improves the chances of people not being confused in the future! 

@nbeyer 🎉

PR #1202, opened 2020-11-09
Remove EM-HTTP SSL patch duplication

### LINKED ISSUE #1201: EM-HTTP SSL patch duplication
This one is already merged https://github.com/igrigorik/em-http-request/pull/340

https://github.com/igrigorik/em-http-request/blob/v1.1.6/lib/em-http/http_connection.rb#L41

So instead of 
```
require 'faraday/adapter/em_http_ssl_patch'
```

in
* `lib/faraday/adapter/em_http.rb`
* `lib/faraday/adapter/em_synchrony.rb`

change to

```ruby
if EventMachine::HttpRequest::VERSION < '1.1.6'
  require 'faraday/adapter/em_http_ssl_patch'
end
```

Or remove patch completely...

kylekeesling:
## Description
This PR is meant to address the duplication of an SSL patch call. Fixes #1201 

Please let let me know if there are any changes/updates necessary!

### DIFF
--- a/lib/faraday/adapter/em_http.rb
+++ b/lib/faraday/adapter/em_http.rb
@@ -90,7 +90,23 @@ module Faraday
 
       include Options
 
-      dependency 'em-http'
+      dependency do
+        require 'em-http'
+
+        if Faraday::Adapter::EMHttp.loaded?
+          begin
+            require 'openssl'
+          rescue LoadError
+            warn 'Warning: no such file to load -- openssl. ' \
+              'Make sure it is installed if you want HTTPS support'
+          else
+            require 'em-http/version'
+            if EventMachine::HttpRequest::VERSION < '1.1.6'
+              require 'faraday/adapter/em_http_ssl_patch'
+            end
+          end
+        end
+      end
 
       self.supports_parallel = true
 
@@ -271,16 +287,5 @@ module Faraday
         end
       end
     end
-  end
-end
-
-if Faraday::Adapter::EMHttp.loaded?
-  begin
-    require 'openssl'
-  rescue LoadError
-    warn 'Warning: no such file to load -- openssl. ' \
-      'Make sure it is installed if you want HTTPS support'
-  else
-    require 'faraday/adapter/em_http_ssl_patch'
   end
 end
--- a/lib/faraday/adapter/em_synchrony.rb
+++ b/lib/faraday/adapter/em_synchrony.rb
@@ -12,6 +12,22 @@ module Faraday
         require 'em-synchrony/em-http'
         require 'em-synchrony/em-multi'
         require 'fiber'
+
+        require 'faraday/adapter/em_synchrony/parallel_manager'
+
+        if Faraday::Adapter::EMSynchrony.loaded?
+          begin
+            require 'openssl'
+          rescue LoadError
+            warn 'Warning: no such file to load -- openssl. ' \
+              'Make sure it is installed if you want HTTPS support'
+          else
+            require 'em-http/version'
+            if EventMachine::HttpRequest::VERSION < '1.1.6'
+              require 'faraday/adapter/em_http_ssl_patch'
+            end
+          end
+        end
       end
 
       self.supports_parallel = true
@@ -133,18 +149,5 @@ module Faraday
         client
       end
     end
-  end
-end
-
-require 'faraday/adapter/em_synchrony/parallel_manager'
-
-if Faraday::Adapter::EMSynchrony.loaded?
-  begin
-    require 'openssl'
-  rescue LoadError
-    warn 'Warning: no such file to load -- openssl. ' \
-      'Make sure it is installed if you want HTTPS support'
-  else
-    require 'faraday/adapter/em_http_ssl_patch'
   end
 end


lib/faraday/adapter/em_http.rb
```diff
@@ -280,7 +280,7 @@ def check_finished
   rescue LoadError
     warn 'Warning: no such file to load -- openssl. ' \
       'Make sure it is installed if you want HTTPS support'
-  else
+  elsif EventMachine::HttpRequest::VERSION < '1.1.6'
```
olleolleolle, line 283: Ah, `begin rescue else` does not support an `elsif`, it will need to be an `if` inside the `else`.

kylekeesling: Oops - I'll get that switched

olleolleolle:
And, thanks for contributing a PR!

kylekeesling:
Happy to help! Thanks for being a maintainer. This library is the backbone of so many gems/projects that I use, I figure the least I can do is submit a little patch 😉

olleolleolle:
You can learn a bit more about the ins and outs of begin-rescue-else-ensure-end in Jonan's article here:
https://blog.newrelic.com/engineering/weird-ruby-2-rescue-interrupt-ensure/

olleolleolle:
Perhaps we have yet to actually _load_ the dependency.

In this adapter, we have the call `dependency "em-http"`.

That method can _also_ be used like this:

```ruby
dependency do
   require 'em-http'
   # And: Do other things, like the thing you want to do
end

Here's the source code for `dependency`:
https://github.com/lostisland/faraday/blob/01c25e880c94ef65d488019da44c39eb7823772c/lib/faraday/dependency_loader.rb#L12

Perhaps this inspires other ways to get this working?

**Comment by kylekeesling** [comment]
I'm a bit out of my element on this (haven't done much work on gems) but as I was digging around inside the 'em-http-request' gem, it doesn't seem as though `version.rb` is included anywhere, which in my mind would indicate that it would then not be accessible when we require `em-http`

When I change the `dependency` call to a block, and explicity include the version file, I can get the sepcs to pass:
```
dependency do
  require 'em-http'
  require 'em-http/version'
end 
```

If that's the case, would the long-term fix actually be to include the version file in the `em-http-request` bootstrap file? IE appending this to it: `require 'em-http/decoders'`

https://github.com/igrigorik/em-http-request/blob/master/lib/em-http.rb

If so I'm happy to open a PR on that library as well to get it fixed.

File: lib/faraday/adapter/em_http.rb [hunk 2]
```diff
@@ -281,6 +281,8 @@ def check_finished
     warn 'Warning: no such file to load -- openssl. ' \
       'Make sure it is installed if you want HTTPS support'
   else
-    require 'faraday/adapter/em_http_ssl_patch'
+    if EventMachine::HttpRequest::VERSION < '1.1.6'
```
ayanko, line 284: Just require version. That's it.
```
require 'em-http-request/version'
if ....
```

ayanko:
> If so I'm happy to open a PR on that library as well to get it fixed.

No. That's not issue in library. Library itself is NOT responsible to define version.
Originally version is defined in gemspec file. So actually you should get it from gemspec API.
But thus it is long way there is kind of convention to have separated small file called version.rb 
So that it reused in gemspec and provides simple version access for developers.

lib/faraday/adapter/em_http.rb
kylekeesling: OK - that's what I had mentioned above. Just force-pushed an update that takes that approach

kylekeesling:
I've updated the conditional to require the version files if we get to that step. All tests seem to be passing.

lib/faraday/adapter/em_synchrony.rb
```diff
@@ -145,6 +145,8 @@ def call_block(block)
     warn 'Warning: no such file to load -- openssl. ' \
       'Make sure it is installed if you want HTTPS support'
   else
-    require 'faraday/adapter/em_http_ssl_patch'
+    if EventMachine::HttpRequest::VERSION < '1.1.6'
```
ayanko, line 148: And here require version as well. They are independent adapters.

kylekeesling: Sorry about that - just pushed this change

kylekeesling:
Ok I think we might have it this time - thanks for your patience as I work through this!

lib/faraday/adapter/em_http.rb
```diff
@@ -281,6 +281,9 @@ def check_finished
     warn 'Warning: no such file to load -- openssl. ' \
       'Make sure it is installed if you want HTTPS support'
   else
-    require 'faraday/adapter/em_http_ssl_patch'
+    require 'em-http/version'
```
olleolleolle, line 284: 👋 Hello, again!

Thanks for working on this issue!

One little detail about our _optional_ gem dependencies:

We need to avoid `require` of non-guaranteed-to-be-available gems in this context of the fix. In other words, this Faraday adapter file must be `require`:able without having the em-http gem available.

The `dependency` block is where are sure that we can `require` the `em-http` gem. When a Faraday user configures usage of the `:em_http` adapter, the `dependency` method is run.

Hope this helps, and that it's not discouraging to run into _details_! Thanks for helping out!

kylekeesling: Alright, I've now moved the `require 'em-http/version'` call to the dependency block in each of the adapter files

olleolleolle:
Cool, now we reached the place where the whole  patch logic (the whole test for "do we need it?" and the patch file require) needs to attempt to move into the dependency block - a time and place where we know we have access to the loaded em_http/version.rb file.

Faraday must avoid non-runtime dependencies on its optional gems, which leads to these shenanigans.

You're well on the way!

kylekeesling:
Thanks @olleolleolle - I just pushed an update that moves that logic into the dependency block and only requires the version file if we get to that step

olleolleolle:
@kylekeesling Awesome work! Let's get this merged!

ayanko:
Release?

Issue #1107, opened 2020-01-03
Infinite recursion (SystemStackError) on load when running with -rdebug with breakpoints


native-api:
## Basic Info

* Faraday Version: 0.17.3, 1.0.0
* Ruby Version: 2.5.3, 2.6.3

## Issue description

Whenever I run Ruby with `-rdebug` and set a breakpoint, the interpreter quits with `SystemStackError` upon loading Faraday.

This makes it impossible to debug any program that uses the library.

## Steps to reproduce

(For Ruby 2.5.3 and Faraday 0.17.3, the result is the same)

```
$ cat test.rb
require 'faraday'
puts "bla-bla-bla"

$ ruby -rdebug test.rb 
/home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/x86_64-linux/continuation.so: warning: callcc is obsolete; use Fiber instead
Debug.rb
Emacs support available.

test.rb:1:require 'faraday'
(rdb:1) b 2
Set breakpoint 1 at test.rb:2
(rdb:1) c
Traceback (most recent call last):
	2: from test.rb:1:in `<main>'
	1: from /home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
/home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- faraday (LoadError)
	9360: from test.rb:1:in `<main>'
	9359: from /home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:34:in `require'
	9358: from /home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:130:in `rescue in require'
	9357: from /home/vmuser/.rvm/rubies/ruby-2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:130:in `require'
	9356: from /home/vmuser/.rvm/gems/ruby-2.6.3/gems/faraday-1.0.0/lib/faraday.rb:21:in `<top (required)>'
	9355: from /home/vmuser/.rvm/gems/ruby-2.6.3/gems/faraday-1.0.0/lib/faraday.rb:126:in `<module:Faraday>'
[...]

PR #1203, opened 2020-11-10
Fix infinite recursion and NameError on load when running with `-rdebug`

native-api:
## Description
Fixes https://github.com/lostisland/faraday/issues/1107

When running with `ruby -rdebug` with a breakpoint set, `self.to_str` (which is not defined) is called repeatedly during module load, including the time after the `class << self` block is loaded but before further blocks that add more bits to `self` are.
As such, the dynamic method resolution machinery has to not break in this partially initialized state.

## Todos
- [x] Tests
If this needs to be autotested, some pointers would be welcome: I've no idea how to load the module under test in a separate Ruby invocation from within RSpec.
- [ ] Documentation

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -115,9 +115,7 @@ module Faraday
       # When running under debugger with a breakpoint set,
       # self.to_str is called repeatedly during module load,
       # including the time before the default_connection getter below is loaded
-      if name == :default_connection
-        return nil
-      end
+      return nil if name == :default_connection
 
       if default_connection.respond_to?(name)
         default_connection.send(name, *args, &block)
@@ -146,9 +144,7 @@ module Faraday
     # self.to_str is called repeatedly during module load,
     # including the time after this getter is loaded but
     # before the libs delay-loading below fires
-    unless const_defined?(:Connection)
-      return nil
-    end
+    return nil unless const_defined?(:Connection)
 
     @default_connection ||= Connection.new(default_connection_options)
   end
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -112,6 +112,13 @@ module Faraday
     # Internal: Proxies method calls on the Faraday constant to
     # .default_connection.
     def method_missing(name, *args, &block)
+      # When running under debugger with a breakpoint set,
+      # self.to_str is called repeatedly during module load,
+      # including the time before the default_connection getter below is loaded
+      if name == :default_connection
+        return nil
+      end
+
       if default_connection.respond_to?(name)
         default_connection.send(name, *args, &block)
       else
@@ -135,6 +142,14 @@ module Faraday
   #   access the Faraday constant directly, such as
   #   <code>Faraday.get "https://faraday.com"</code>.
   def self.default_connection
+    # When running under debugger with a breakpoint set,
+    # self.to_str is called repeatedly during module load,
+    # including the time after this getter is loaded but
+    # before the libs delay-loading below fires
+    unless const_defined?(:Connection)
+      return nil
+    end
+
     @default_connection ||= Connection.new(default_connection_options)
   end
 


lib/faraday.rb
```diff
@@ -112,6 +112,13 @@ def respond_to_missing?(symbol, include_private = false)
     # Internal: Proxies method calls on the Faraday constant to
     # .default_connection.
     def method_missing(name, *args, &block)
+      # When running under debugger with a breakpoint set,
+      # self.to_str is called repeatedly during module load,
+      # including the time before the default_connection getter below is loaded
+      if name == :default_connection
+        return nil
+      end
```
olleolleolle, line 120: Style opinion: These guards would look easier to read as a "suffix if", like

```ruby
return nil if name == :default_connection
```

iMacTia (changes requested):
Thanks @native-api for trying to solve this, and apologies if I never came back to it since you initially raised the issue.

I'm still not entirely sure of what's is going on, but this solution of introducing guards looks more like a patch to me and it's not really elegant.
I don't think filling your code with guards due to some sort of loading race-condition is a good solution and I'd much rather consider others.

It might be worth investigating the issue a little more: you mentioned `to_str` is being called, but why is it going to method_missing? Can we do any change to our code to avoid it breaking? Maybe we're not following a convention or something? Can you share your findings so far so we can brainstorm together?

Last time I had a look ([here](https://github.com/lostisland/faraday/issues/1107#issuecomment-572551237)), I did some progress by moving all the `self.*` methods into the `class << self` block. Although that didn't fix the problem entirely, it shows the kind of solution I would prefer. I feel like we're missing something 🤔

lib/faraday.rb
native-api: Sure. I don't really have a preference here. I first tried `name == :default_connection && return nil` but that turned out to be invalid syntax :frowning_face:

native-api:
> I'm still not entirely sure of what's is going on, but this solution of introducing guards looks more like a patch to me and it's not really elegant.
> I don't think filling your code with guards due to some sort of loading race-condition is a good solution and I'd much rather consider others.

**This happens because we initialize the nonexistent member resolution mechanism in parts rather than all at once:** first override `respond_to_missing?` and `missing_method` in the `class << self` block; then add `default_connection` getter, and finally load `faraday/connection` with the `Connection` definition.
A module code is run block by block upon import like any other script: definitions are parsed and added as corresponding entitites, raw code is executed as it's encountered. So I wouldn't call it a "race condition", it's just the way how code is executed.
A debugger is likely calling `<module>.to_str` at every trace point -- probably to find out if it needs to break execution.

As such, **an alternative solution is to initialize the entire mechanism all at once:** define all the overriding methods inside the `class << self` block and `require` all the submodules before it.
But **there's probably a reason why they are delay-loaded** so I didn't want to touch that potential can of worms and offered a solution that requires minimal changes to code structure. It looks quite reasonable actually to first check if a member exists and only then use it.

native-api:
Here's a trace of calls to the method resolution machinery during module load:

```
$ bundle exec ruby -rdebug -e "require 'faraday'; puts 'bla-bla-bla'" 2>&1 <<!
b Rakefile:1
c
!
<...>
respond_to_missing?
:to_str
method_missing
:default_connection
respond_to_missing?
:to_str
respond_to_missing?
:to_str
bla-bla-bla
```

iMacTia:
> As such, an alternative solution is to initialize the entire mechanism all at once: define all the overriding methods inside the class << self block and require all the submodules before it.

If I got it correctly, moving the all the methods into the `class << self` block and leaving `method_missing` / `respond_to_missing?` at the very bottom of the block might solve the issue, right?
I'd very much like to test that solution (potentially on a separate PR, if you prefer to keep this one) and see if that works.

Should that not work, then looking at your last comment it *seems* like we might get away just with changing the implementation of `respond_to_missing?` to simply return false for `to_str`?

native-api:
> If I got it correctly, moving the all the methods into the `class << self` block and leaving `method_missing` / `respond_to_missing?` at the very bottom of the block might solve the issue, right?

No. You also need to `require 'faraday/connection'` (and anything else that might be needed for `Connection.new()` to work) before the `class <<self` block. I don't know if that will work without other changes (e.g. the submodules might expect something to be already initialized in the main module upon their load and malfunction somewhere down the line if that is not so) thus I don't want to pursue that way.

> we might get away just with changing the implementation of `respond_to_missing?` to simply return false for `to_str`?

Yes. But I don't want to rely on the knowledge that it's specifically `to_str` being called -- since that may change in the future or in different circumstances (e.g. with a different debugger). I would very much prefer a universal solution that works regardless of what is called.

native-api:
> But I don't want to rely on the knowledge that it's specifically `to_str` being called

Not to mention that we might want `to_str` to be successfully resolved through `default_connection` as usual after the module's load.

native-api:
> I'd very much like to test that solution (potentially on a separate PR, if you prefer to keep this one) and see if that works.

Go ahead, knock yourself out :slightly_smiling_face: : https://github.com/lostisland/faraday/pull/1204

As I suspected, some submodules expect the main module to already be initialized and break on load.

iMacTia:
Sorry for being a pain 🙏 , that's far from what I really want, which is finding the right solution to the problem (if at all possible). I've left a comment in #1204 and opened #1205 
If we're still unable to find an alternative solution, I'll reconsider this PR, so keep it warm for now 👍

iMacTia:
Fixed in #1205 🎉 
Thanks again @native-api for all the research 🙏

PR #1204, opened 2020-11-12
Fix rdebug recursion alt solution

native-api:
## Description
An alternative solution to https://github.com/lostisland/faraday/pull/1203

May break something due to submodules being loaded early.

## Todos
- [ ] Tests

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -115,9 +115,7 @@ module Faraday
       # When running under debugger with a breakpoint set,
       # self.to_str is called repeatedly during module load,
       # including the time before the default_connection getter below is loaded
-      if name == :default_connection
-        return nil
-      end
+      return nil if name == :default_connection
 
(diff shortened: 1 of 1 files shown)


iMacTia:
Thank you @native-api 🙏 (especially for being patient with me 😄!)
I think this is a great starting point, and that we're highlighting a few legacy issues Faraday have been dragging due to the fact it was written first more than 10 years ago!
I've given a quick refresh to the version and the methods constants definitions and pushed a new PR (#1205).

I was unable though to understand how to test that with ruby-debug, to be entirely honest I've not been using it before in favour of bye bug or pry instead (or Rubymine debugger).
Would you be able to give it a try (and share with me how you did it) and let me know if the original issue is now solved?

iMacTia:
Fixed in #1205 🎉 
Thanks again @native-api for all the research 🙏

PR #1208, opened 2020-11-27
test against ruby head

grosser:
Fixes #1189
replaces https://github.com/lostisland/faraday/pull/1191

net-http-persistent does not support gemspec builds, so using a branch, see https://github.com/drbrain/net-http-persistent/pull/101

PR #1209, opened 2020-11-28
avoid argument warnings / prepare for ruby 3

grosser:
builds on top of https://github.com/lostisland/faraday/pull/1208
reverts https://github.com/lostisland/faraday/pull/1153 since 2.7 no longer has warnings and 3.0 works

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,15 @@ jobs:
   build:
     needs: [linting]
     runs-on: ubuntu-latest
+    name: build ${{ matrix.ruby }}
     strategy:
+      fail-fast: false
       matrix:
-        ruby: ['2.4', '2.5', '2.6', '2.7', 'head']
+        ruby: ['2.4', '2.5', '2.6', '2.7']
+        experimental: [false]
+        include:
+        - ruby: head
+          experimental: true
 
     steps:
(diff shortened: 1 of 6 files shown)


olleolleolle:
In order to go forward, I resolved the conflict in favor of the merged PR #1211

grosser:
thx, there are some real bugs in here, (ran tests on 2.4 / 2.7 and no combination made both work)
so I think we need to either keep the dependency or do what it does in a simpler way :(

olleolleolle:
@grosser Does this change your mind on #1211?

olleolleolle:
@grosser Now I'm just messing with this, perhaps we can take a smaller step forwards trying to find working code. Grr, now I made a commit which predictably works on 2.7, but now the others. As you say, perhaps keeping the ruby2_keywords gem to support Olde Versions.

In addition, I noted that the coveralls gem is old and wonky, and I think I recall someone's actually shepherding it in a GitHub repo, somewhere. So, one _could_ use that gem, to avoid red output at the end of the test run.

grosser:
Could also try https://github.com/grosser/single_cov :D

iMacTia:
We support Ruby 3.0, 3.1 and 3.2 pretty well now so this should not be necessary anymore 👍

PR #1213, opened 2020-12-01
EmHttp adapter: drop superfluous loaded? check

olleolleolle:
This PR removes a conditional, which is always truthy.

## Description

This PR removes a check that I think is superfluous at this point 

We are loaded in this section, and loaded meant "has no load errors".

Follow-up of #1202. 

## Todos

Review.


iMacTia (approved):
Totally agree! LGTM

PR #1221, opened 2020-12-23
Fix #1219 Net::HTTP still uses env proxy

### LINKED ISSUE #1219: Net::HTTP Adapter still doesn't ignore env proxy
Hi there

I found out that Faraday is being used in the Inspec framework https://github.com/inspec/inspec for making HTTP requests.

My question is: How can I make Faraday ignore the global proxy settings (the `http_proxy` env var).

Since I just found out about Faraday, I'm fairly new to Ruby and my use case is to make working Inspec test so I don't even know how to test it otherwise; I made a custom Inspec resource (copy&paste of <https://github.com/inspec/inspec/blob/master/lib/inspec/resources/http.rb#L142>).
> NOTE: The initial `module Inspec::Resources` and the last `end` needs to be removed from `http.rb` for Inspec to register it as new resource. Also new class name and `name` should be given to that resource.

I tried setting `Faraday.ignore_env_proxy = true` according to https://github.com/lostisland/faraday/issues/752 but that didn't help.

I also tried `Faraday.new(url, :proxy => '')` and `conn.proxy=''` but that didn't work either.

The only thing that works is to remove the ENV variable for whole Inspec by `http_proxy= inspec exec /tests`.

Could you please advise what am I missing? I'd really want to be able to disable the proxy for specific calls instead.

_Originally posted by @meridius in https://github.com/lostisland/faraday/discussions/1218_

iMacTia:
## Description
Ensures that the `proxy_address` is passed as `nil` to `Net::HTTP` to avoid that it picks the one from the environment.
Also, since Net::HTTP::Proxy is obsolete (https://ruby-doc.org/stdlib-2.3.2/libdoc/net/http/rdoc/Net/HTTP.html#method-c-Proxy), it has now been replaced with a different call to `.new`.
Fixes #1219

### DIFF
--- a/spec/support/shared_examples/request_method.rb
+++ b/spec/support/shared_examples/request_method.rb
@@ -3,6 +3,7 @@
 shared_examples 'proxy examples' do
   it 'handles requests with proxy' do
     res = conn.public_send(http_method, '/')
+
     expect(res.status).to eq(200)
   end
 
@@ -233,7 +234,7 @@ shared_examples 'a request method' do |http_method|
 
   context 'when a proxy is provided as option' do
     before do
-      conn_options[:proxy] = 'http://google.co.uk'
+      conn_options[:proxy] = 'http://env-proxy.com:80'
     end
 
     include_examples 'proxy examples'
--- a/spec/support/shared_examples/request_method.rb
+++ b/spec/support/shared_examples/request_method.rb
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+shared_examples 'proxy examples' do
+  it 'handles requests with proxy' do
+    res = conn.public_send(http_method, '/')
+    expect(res.status).to eq(200)
+  end
+
+  it 'handles proxy failures' do
+    request_stub.to_return(status: 407)
+
+    expect { conn.public_send(http_method, '/') }.to raise_error(Faraday::ProxyAuthError)
+  end
+end
+
 shared_examples 'a request method' do |http_method|
   let(:query_or_body) { method_with_body?(http_method) ? :body : :query }
   let(:response) { conn.public_send(http_method, '/') }
@@ -223,16 +236,7 @@ shared_examples 'a request method' do |http_method|
       conn_options[:proxy] = 'http://google.co.uk'
     end
 
-    it 'handles requests with proxy' do
-      res = conn.public_send(http_method, '/')
-      expect(res.status).to eq(200)
-    end
-
-    it 'handles proxy failures' do
-      request_stub.to_return(status: 407)
-
-      expect { conn.public_send(http_method, '/') }.to raise_error(Faraday::ProxyAuthError)
-    end
+    include_examples 'proxy examples'
   end
 
   context 'when http_proxy env variable is set' do
@@ -244,15 +248,14 @@ shared_examples 'a request method' do |http_method|
       end
     end
 
-    it 'handles requests with proxy' do
-      res = conn.public_send(http_method, '/')
-      expect(res.status).to eq(200)
-    end
+    include_examples 'proxy examples'
 
-    it 'handles proxy failures' do
-      request_stub.to_return(status: 407)
+    context 'when the env proxy is ignored' do
+      around do |example|
+        with_env_proxy_disabled(&example)
+      end
 
-      expect { conn.public_send(http_method, '/') }.to raise_error(Faraday::ProxyAuthError)
+      include_examples 'proxy examples'
     end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples/request_method.rb
```diff
@@ -218,17 +218,41 @@
     end
   end
 
-  it 'handles requests with proxy' do
-    conn_options[:proxy] = 'http://google.co.uk'
+  context 'when a proxy is provided as option' do
+    before do
+      conn_options[:proxy] = 'http://google.co.uk'
```
olleolleolle, line 236: Maybe this value is a bit not-self-documenting? "Does Google act as a proxy? How?" the reader (me) might think.

spec/support/shared_examples/request_method.rb
```diff
@@ -218,17 +218,41 @@
...
+    end
+
+    it 'handles requests with proxy' do
+      res = conn.public_send(http_method, '/')
+      expect(res.status).to eq(200)
```
olleolleolle: Minor, which I woudl like: Separate the "acts" from the "assert" with a newline

spec/support/shared_examples/request_method.rb
```diff
@@ -218,17 +218,41 @@
...
+
+    around do |example|
+      with_env 'http_proxy' => proxy_url do
+        example.run
+      end
+    end
+
+    it 'handles requests with proxy' do
+      res = conn.public_send(http_method, '/')
+      expect(res.status).to eq(200)
```
olleolleolle: (A newline, please. Thanks!)

spec/support/shared_examples/request_method.rb
```diff
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+shared_examples 'proxy examples' do
+  it 'handles requests with proxy' do
+    res = conn.public_send(http_method, '/')
+    expect(res.status).to eq(200)
```
olleolleolle, line 6: A newline, yeah?

spec/support/shared_examples/request_method.rb
```diff
@@ -1,5 +1,18 @@
 # frozen_string_literal: true
 
+shared_examples 'proxy examples' do
```
olleolleolle, line 3: Really, I like this shared one.

olleolleolle (approved):
Thanks for improving this!

walterking:
faraday is a transitive dependency for us, is the expectation now that faraday will ignore env variables always? we were depending on that behavior

iMacTia:
@walterking Not at all! The problem was that using Net::HTTP adapter and setting `Faraday.ignore_env_proxy = true` would still result in the request being proxied.

If you don't set that config (which defaults to false), Faraday will continue to automatically pick the proxy from the environment for you

walterking:
hm, we dont set that variable that i can see.  we updated to 1.2 and started to get connection refused errors, which we usually see when the proxy isnt being used, and this pr seemed like the most likely to cause that change

iMacTia:
Mmmh thanks for the example, I'll have a closer look and try to understand what's wrong, but yes it seems like 1.2 did change something...

That said, http_proxy should be used for http requests, while https_proxy is the one for https, so I'll try a few combinations. Right now my guess is that Net::HTTP might have used them incorrectly 🤔

walterking:
interesting. in our app we set both, but when i switch my local to https it seems to do the right thing

iMacTia:
So, I did some extra testing and I believe the issue is with `Net::HTTP`.

```ruby
# ~ HTTP_PROXY=http://127.0.0.1:1234 irb
> require 'net/http'
> Net::HTTP.new('http://www.google.com').get('/')
=> Errno::ECONNREFUSED (Failed to open TCP connection to 127.0.0.1:1234 (Connection refused - connect(2) for "127.0.0.1" port 1234))
> Net::HTTP.new('https://www.google.com').get('/')
=> Errno::ECONNREFUSED (Failed to open TCP connection to 127.0.0.1:1234 (Connection refused - connect(2) for "127.0.0.1" port 1234))

# ~ HTTPS_PROXY=http://127.0.0.1:1234 irb
> require 'net/http'
> Net::HTTP.new('http://www.google.com').get('/')
=> SocketError (Failed to open TCP connection to http://www.google.com:80 (getaddrinfo: nodename nor servname provided, or not known))
> Net::HTTP.new('https://www.google.com').get('/')
=> Errno::ECONNREFUSED (Failed to open TCP connection to 127.0.0.1:1234 (Connection refused - connect(2) for "127.0.0.1" port 1234))
```

So as you can see `HTTPS_PROXY` is correctly ignored for `http` calls, however `HTTP_PROXY` is used for both `http` and `https`, which doesn't really makes sense... Faraday relies on `URI#find_proxy` instead (https://ruby-doc.org/stdlib-2.6.3/libdoc/uri/rdoc/URI/Generic.html#method-i-find_proxy) which behaves correctly.

PR #1225, opened 2020-12-29
CHANGELOG: add releases after 1.0

olleolleolle:
## Description

We released code, and didn't add it to this file, which we distribute.

## Todos

- [ ] Decide if we want to make the links in there work
- [ ] Decide if we want to have release dates for each release


iMacTia (approved):
You're right @olleolleolle, I remember we had this discussion and decided to keep the CHANGELOG updated as well, but since it's not something I'm used to I completely forgot to do so for the last releases, thanks for doing it and hopefully I'll remember next time 😅

PR #1226, opened 2020-12-29
Adds Ruby 3.0 to CI Matrix

iMacTia:
## Description
Adds Ruby 3.0 to CI Matrix

## Additional Notes
* Should we keep Ruby head as experimental for future Ruby releases?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.4', '2.5', '2.6', '2.7']
+        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0']
         experimental: [false]
         include:
         - ruby: head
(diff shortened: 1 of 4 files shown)


olleolleolle:
Timeouts are reported in a wonky way: on 3.0

https://github.com/lostisland/faraday/pull/1226/checks?check_run_id=1622851358#step:6:4461

iMacTia:
@olleolleolle problem with `Net::HTTP::Persistent` timeouts it fixed 🙌
What do you think about keeping the ruby-head in the matrix? Is it still useful now that we added 3.0?

olleolleolle:
Oh, I think we can throw it out for the time being. Useful trick to have when big changes coming ahead. But not for all days. 

Great news!

iMacTia:
Thanks, makes sense.
Also, I just realised you already identified and attempted to fix this issue in #1154 !
In this branch, I raise a `Faraday::Timeout` exception to keep it consistent with the older behaviour, but does that mean we can now support `Net::HTTP::Persistent` also for older rubies 🤔?

So at the moment, we can see all tests are passing, but v4.0 is used only for Ruby 3.0 (see Gemfile)
I'll update the Gemfile to allow v4.0 also for lower rubies and give it a try 👍

PR #1205, opened 2020-11-12
Native api fix rdebug recursion alt

iMacTia:
## Description
Adds commit on top of #1204 

## Todos
List any remaining work that needs to be done, i.e:
- [x] Test with ruby-debug

PR #1227, opened 2020-12-31
Don't assign to ::Timer

bpo:
## Description
The change in 1595e6fc moved the definition of Timer out of the Faraday module definition so that it was being defined in the root namespace. I don't think that was the intention of #1205 where that patch was introduced.

This changes the definition from `::Timer` to `::Faraday::Timer`.

I discovered this when a separately defined `::Timer` in my app was overridden in Faraday v1.2.0


iMacTia (approved):
You're absolutely right @bpo!
Thanks for catching and fixing this 🎉

PR #1230, opened 2021-01-02
Refactor CI to remove duplicated line

tricknotes:
## Description
"libcurl4-openssl-dev" is already installed in previous step.
https://github.com/lostisland/faraday/blob/v1.3.0/.github/workflows/ci.yml#L54-L56

## Todos
- [x] Pass CI 

## Additional Notes
This line was added in https://github.com/lostisland/faraday/pull/1100, but it seems unintentional.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
 
     - name: Build
       run: |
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
Good eye! Thanks, @tricknotes!

Issue #1214, opened 2020-12-04
URI validation fails when it contains reserved characters 


luizkowalski:
## Basic Info

* Faraday Version: 1.1.0
* Ruby Version: 2.7.1

## Issue description
One of the service providers I'm using has URLs like this: `https://service.com/service:search`. This is how we setup Faraday:

```ruby
class Provider::Client
 ...
 delegate :post, :patch, :delete, :put, to: :client

 def client
  Faraday.new(url: 'https://service.com/') do |faraday|
    faraday.headers = headers.merge(content_type: content_type)
    faraday.response(:json, content_type: CONTENT_TYPE)
    faraday.basic_auth(USER, PASSWORD)
    faraday.adapter(Faraday.default_adapter)
  end
 end
end
```
and when I invoke it

```ruby
data = Provider::Client.post(
  "service:search?limit=#{LIMIT}&offset=#{offset}", search_params
)
```

but when I do this with the URL in question, I get this: 

```ruby
URI::InvalidURIError: query conflicts with opaque
from /Users/luiz/.rbenv/versions/2.7.1/lib/ruby/2.7.0/uri/generic.rb:832:in `query='
```

I assume it was something related to net/http adapter since it is the default adapter so I did the same call manually

```ruby
require "uri"
require "net/http"

url = URI("https://service.com/service:search?limit=50&offset=400")

https = Net::HTTP.new(url.host, url.port);
https.use_ssl = true
request = Net::HTTP::Post.new(url)
request["Accept"] = "custom_header"
request["Authorization"] = "Basic xxx="
request["Content-Type"] = "application/json"
request.body = "long_json"
response = https.request(request)

puts response.read_body
```

and this actually works so I suspect that something in Faraday is validating this URL in a different way than net/http

iMacTia:
Hi @luizkowalski, the exception is coming from `URI` and is raised in the first lines of the `query=` method:

```ruby
def query=(v)
  return @query = nil unless v
  raise InvalidURIError, "query conflicts with opaque" if @opaque
  ...
end
```

Faraday does some internal magic on the request url by splitting and recombining it to help with some common transformations, and I presume at some point it may cause the `@opaque` to be set thanks to the presence of the `:` in the URL.
I've traced the issue down to this line: https://github.com/lostisland/faraday/blob/c26df87b8653db4f270e3bcdc7a15bcdd2dd5cae/lib/faraday/connection.rb#L525
[...]

PR #1237, opened 2021-01-12
escape colon in path segment

yarafan:
## Description
Considering example
```ruby
f =  Faraday.new(url: 'https://service.com/') do |faraday|
    faraday.adapter(Faraday.default_adapter)
  end
f.post('service:search')
```
`URI.parse` parses relative path `service:search` as opaque_part (according to https://tools.ietf.org/html/rfc2396#section-3) (eg `mailto:foo@example.com`), so `Faraday::Connection#build_exclusive_url` produces wrong URI as result and passing query params leads to `InvalidURIError`

```
irb(main):004:0> URI.parse('mailto:foo@example.com').opaque
=> "foo@example.com"
irb(main):005:0> URI.parse('service:search').opaque
=> "search"
```
Fixes https://github.com/lostisland/faraday/issues/1214

## Todos
- [x] Rubocop (not sure what to do with ClassLength)

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -524,7 +524,6 @@ module Faraday
       end
       url = url && URI.parse(url.to_s).opaque ? url.to_s.gsub(':', '%3A') : url
       uri = url ? base + url : base
-      uri = URI.parse(CGI.unescape(uri.to_s))
       if params
         uri.query = params.to_query(params_encoder || options.params_encoder)
       end
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -522,7 +522,9 @@ module Faraday
         base = base.dup
         base.path = "#{base.path}/" # ensure trailing slash
       end
+      url = url && URI.parse(url.to_s).opaque ? url.to_s.gsub(':', '%3A') : url
       uri = url ? base + url : base
+      uri = URI.parse(CGI.unescape(uri.to_s))
       if params
         uri.query = params.to_query(params_encoder || options.params_encoder)
       end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
 
     - name: Build
       run: |
-        sudo apt-get install libcurl4-openssl-dev
         gem install bundler -v '<2'
         bundle install --jobs 4 --retry 3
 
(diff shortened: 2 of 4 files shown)


lib/faraday/connection.rb
```diff
@@ -522,7 +522,9 @@ def build_exclusive_url(url = nil, params = nil, params_encoder = nil)
         base = base.dup
         base.path = "#{base.path}/" # ensure trailing slash
       end
+      url = url && URI.parse(url.to_s).opaque ? url.to_s.gsub(':', '%3A') : url
       uri = url ? base + url : base
+      uri = URI.parse(CGI.unescape(uri.to_s))
```
yarafan, line 527: Not sure whether final uri should be unescaped

iMacTia: That's a good point, this may cause issues in case the URL contains escaped characters like spaces? Escaped columns may also be part of the URL, so I'm not sure unescaping here is a good idea at all.
E.g. `Faraday.get('https://service.com/path%20with%20spaces')

iMacTia (changes requested):
I like the approach and even more the test coverage! But unfortunately the escaping solution might not be the best one 🤔. I'm curious to see if other tests are passing (I'd expect at least one to fail, if we're testing spaces!), but Rubocop is in the way...

The `ClassLength` cop is a tricky one, so feel free to increase the limit in the `.rubocop_todo.yml` file for this time. Since you added 2 lines that should be 236 now, I guess?

iMacTia:
@yarafan this looks good now, but is your service call working as expected even with the `:` escaped?
@luizkowalski maybe you want to have a quick look as well?

yarafan:
@iMacTia I don't know any service with such uri. I tried to access to uri in issue and got 302.
Would be great if @luizkowalski also check this case

luizkowalski:
@yarafan @iMacTia wow I disabled email from comments from GitHub (cause I get too many) so I missed this one, guys, so sorry! I will check tomorrow morning first thing and will come back to you!

Thanks for the PR!

luizkowalski:
Hey guys! 

I can confirm it works and this is how the URL appeared on the logs

```ruby
INFO  Rails : request: POST https://subdomain.service.com/api/loans%3Asearch?limit=50&offset=2900
```

iMacTia (approved):

PR #1239, opened 2021-01-18
add ciphers to ssl options

apneadiving:
## Description
As I explain in https://github.com/lostisland/faraday-net_http/pull/4 there is sometimes a need to configure `ciphers`.
I need to add this option to make the adapter specs pass.

## Todos
Seems the options list is not tested. Yet I feel like it's fine as is.


apneadiving:
closed in favor of adapter block config: https://lostisland.github.io/faraday/adapters/net-http
Thanks!

olleolleolle:
Thank you, for adding the workaround!

PR #1243, opened 2021-02-02
Fix === for subclasses of deprecated classes

mervync:
## Description

Fix the `===` implementation for deprecated classes.

An example of code that behaves incorrectly with the current implementation:

```
class SomeError < Faraday::Error::ClientError; end
class AnotherError < Faraday::Error::ClientError; end

begin
  raise AnotherError('a')
rescue SomeError
  puts "oops"
end

# => oops
```

## Todos
List any remaining work that needs to be done, i.e:
- [x] Tests
- [ ] Documentation

## Additional Notes
This PR is intended for the 0.17.x release

PR #1244, opened 2021-02-03
Adds fail-fast = false to ci workflow

iMacTia:
## Description
Check if the 0.1x branch CI build is still working
(Linked to #1243)

## Todos
- [x] Always run all matrix combinations to see which Ruby versions are still good and which ones are not
- [x] Decide if removing support for some rubies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         ruby: ['2.0', '2.1', '2.2', '2.3', '2.4', '2.5', '2.6', '2.7']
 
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
This is good information to get.

I believe the GitHub Actions infrastructure only includes in-support Ruby versions. Or, I believe that that is "the promise" they make. **Update:** Ruby 2.4 and up.

iMacTia:
@olleolleolle happy to keep supporting 2.0+, but since GH won't build them, how do we ensure they're actually working?
I think a good middle ground could be to remove them form the CI pipeline (keep only 2.4+) and then do our best to address any issue that arises. The only problem with this approach is that it won't be possible to spot issues before releasing them, but fixes should be small and limited to bugfix versions only.

We could also try to run tests locally against 2.0 before a release

olleolleolle (approved):
Looks super!

PR #1250, opened 2021-03-19
Replacing Net::HTTP::Persistent with faraday-net_http_persistent

MikeRogers0:
## Description

Using https://github.com/lostisland/faraday-net_http_persistent over storing the code with in this repo :)

Part of **v2 Wishlist**: https://github.com/lostisland/faraday/issues/953

## Todos

- [x] Tests
- [ ] Documentation

## Additional Notes

Optional section


MikeRogers0:
> Your bundle requires gems that depend on each other, creating an infinite loop. Please remove either gem 'faraday' or gem 'faraday-net_http_persistent' and try again.

I need to remove `faraday` as a dependency - I'll update the other repo & come back to this :)

olleolleolle:
> I need to remove `faraday` as a dependency - I'll update the other repo & come back to this :)

Now that PR has been merged ✨ - thanks a lot, @MikeRogers0!

MikeRogers0:
@iMacTia What is the release procedure for https://github.com/lostisland/faraday-net_http_persistent/blob/master/.github/workflows/publish.yml - Do I just push up a tag & add the release notes?

olleolleolle:
@MikeRogers0 I _think_ the Released event is created in the GitHub API events universe when hitting the "GitHub Releases" create button. A big green one, in the UI.

iMacTia:
@MikeRogers0 As @olleolleolle said, there's no need to create a tag locally and pushing it, just use the GitHub UI to release a new version (remember to bump the VERSION file first, I always forget!) and that will automatically create a tag for you and trigger the Action to deploy on Rubygems 👍

MikeRogers0:
👀  Ruby 3.0 is failing due to:

```
Bundler could not find compatible versions for gem "net-http-persistent":
  In Gemfile:
    net-http-persistent (>= 3.0)

    faraday was resolved to 1.3.0, which depends on
faraday-net_http_persistent (~> 1.0) was resolved to 1.0.1, which depends
on
        net-http-persistent (~> 3.1)
```

I will update the requirements in the `faraday-net_http_persistent` gem to be `>= 3.0` to match what is in the Gemfile this weekend & see if that solves the issue :)

iMacTia:
@MikeRogers0 actually, I think we should remove `Net::HTTP::Persistent` direct dependency from `Faraday` Gemfile, as that should be decided by `faraday-net_http_persistent`.

I think the problem you're having may be with @grosser's adding support for Ruby 3 to `Net::HTTP::Persistent`
If you check our Gemfile, we actually have a special directive to manage Ruby 3.0 as a special case (https://github.com/lostisland/faraday/blob/master/Gemfile#L28). So those 2 lines could be moved from Faraday's Gemfile to your gem's one.

MikeRogers0:
Woo @iMacTia I think that worked perfectly!

Gemfile
```diff
@@ -24,9 +24,6 @@ group :test, :development do
   gem 'excon', '>= 0.27.4'
   gem 'httpclient', '>= 2.2'
   gem 'multipart-parser'
-  # TODO: remove this once v4 is released
-  options = (RUBY_VERSION.start_with?('3') ? { github: 'grosser/net-http-persistent', branch: 'grosser/spec' } : {})
-  gem 'net-http-persistent', '>= 3.0', **options
```
olleolleolle, line 29: So very nice to move this out of the way!

olleolleolle (approved):
The changes look 💯.

👍 Thanks for joining this effort, @MikeRogers0! 

iMacTia (approved):
Thank you so much @MikeRogers0 🙏!
One more out of the way 🥳🎉🍾!

Issue #1254, opened 2021-03-26
CI: Coveralls encountered an exception: NoMethodError undefined method 'coverage' Did you mean?  coverage_data


olleolleolle:
## Basic Info

* Faraday Version: unreleased
* Ruby Version: 2.7, 3.0.0 (perhaps others)

## Issue description
```
Coveralls encountered an exception: 
NoMethodError
undefined method `coverage' for #<SimpleCov::SourceFile:0x00005638ba843c50 @filename="/home/runner/work/faraday/faraday/lib/faraday.rb",
```

and then later, after lots of output: `Did you mean?  coverage_data`

<details>

```
Coveralls encountered an exception:
NoMethodError
[...]

PR #1255, opened 2021-03-27
Fix NoMethodError undefined method 'coverage'

Maroo-b:
## Description
This is a fix for https://github.com/lostisland/faraday/issues/1254
As mentioned in coveralls-ruby issue: https://github.com/lemurheavy/coveralls-ruby/issues/161, using coveralls-ruby-rebon should fix the problem.

Here is a screenshot from the CI logs as a proof :) 
coveralls-logs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ group :lint, :development do
 end
 
 group :test, :development do
-  gem 'coveralls', require: false
+  gem 'coveralls_reborn', require: false
   gem 'em-http-request', '>= 1.1', require: 'em-http'
   gem 'em-synchrony', '>= 1.0.3', require: %w[em-synchrony em-synchrony/em-http]
   gem 'excon', '>= 0.27.4'


Gemfile
```diff
@@ -18,7 +18,7 @@ group :lint, :development do
 end
 
 group :test, :development do
-  gem 'coveralls', require: false
+  gem 'coveralls_reborn', '~> 0.21.0', require: false
```
olleolleolle, line 21: I think we can allow _any_ version of coveralls_reborn.

olleolleolle (commented):
Thanks for setting this up!

Gemfile
Maroo-b: The fix was only introduced for v "0.21.0"  but I removed the explicit version as you recommended.

iMacTia (approved):
Thanks for providing the solution! Together with #1256, we should be able to get coveralls back up 🎉

Maroo-b:
@iMacTia I'm glad that this small fix was helpful. Feel free to share any issue up-for-grabs in this project and I can dedicate some free time to it.

iMacTia:
We now get the following

```
Coverage report generated for RSpec to /home/runner/work/faraday/faraday/coverage. 1946 / 2095 LOC (92.89%) covered.
[Coveralls] Submitting to https://coveralls.io/api/v1
[Coveralls] Couldn't find a repository matching this job.
Coverage is at 92.89%.
Coverage report sent to Coveralls.
```

So it seems reports are now sent correctly to Coveralls 👍!
We can get this merged and look into that "Couldn't find a repository matching this job" separately

iMacTia:
@Maroo-b Thanks! I'll send anything that pops in my mind, but for now I'd suggest to look at the "help wanted" and "good first issue" labels in the repo issues 🎉

PR #1257, opened 2021-04-12
Replace Excon adapter with Faraday::Excon gem, and fix autoloading issue with Faraday::NetHttpPersistent

iMacTia:
## Description
Use [Faraday::Excon gem](https://github.com/lostisland/faraday-excon) instead of internal adapter.
Also: fixes autoloading issues with Faraday::NetHttpPersistent.


olleolleolle:
I see you did 2 things, so I added the 2nd thing to the title. Makes the commit message truer.

OK, cop.

```
faraday.gemspec:20:3: C: Gemspec/OrderedDependencies: Dependencies should be sorted in an alphabetical order within their section of the gemspec. Dependency faraday-excon should appear before faraday-net_http_persistent.
  spec.add_dependency 'faraday-excon', '~> 1.0'
```

olleolleolle (approved):
❤️ This has been a long time coming, and it is happening!

@iMacTia thank you!

gurgeous:
Thank you. This speeds up faraday load time by ~20%.

rainerborene:
shouldn't `faraday-excon` and `faraday-net_http_persistent` be optional dependencies? what is the reasoning behind shipping two HTTP clients?

iMacTia:
@rainerborene totally agree! That's the plan for Faraday v2.0 🎉!
The only reason we keep shipping them in the 1.x releases is for backwards compatibility, but they will all go away in v2.0 and you'll simply add the one you want to use 👌

tisba:
I was wondering why `excon` made it into my dependencies then I found this (we don't use `excon`). I think it should be mentioned in the Changelog explicitly as this is quite confusing.

iMacTia:
@tisba Ah I think I see what @rainerborene also meant now. The adapters need to be included, as they were included with Faraday before, but `excon` and `net_http_persistent` were not, people had to manually add them to the Gemfile before.
They now are part of the adapters dependencies.

This clearly doesn't work, by the time we move out all adapters people will have their dependencies full of clients they might not be using. I'll get this fixed and release a v1.4.1 to address that, thanks both for raising it 👍 !

PR #1262, opened 2021-04-13
CONTRIBUTING: add a policy on inclusive language

olleolleolle:
## Description

This policy wants to make the Code of Conduct's note on inclusive language **actionable** by adding resources, and a statement that we intend to do this.


iMacTia (approved):
Fantastic addition, thank you @olleolleolle!
This is good as it is, but maybe it's worth adding also a course of action of how users can raise concerns regarding non-inclusive language with the core team? An email to a point of contact, or some sort of channel where issues can be raised with us and we can intervene.

While looking into ways of searching for already existing non-inclusive works in our repos, I found this: https://github.com/aergonaut/rubocop-inclusivity/
We can add it in on a different PR to ensure we stay compliant, at least in the codebase 👍

PR #1264, opened 2021-04-13
Drop RDoc support file .document

olleolleolle:
## Description

~This is an experimental change, which removes references to files which do not exist in this repo.~

This PR removes an RDoc artifact, which we don't use, since we have this repo configured for YARD instead.


iMacTia (commented):
I think this file is a legacy from RDoc, but since we switched to YARD we should now rely on the `.yardopts` file instead?
See https://rdoc.info/gems/yard/0.9.23/YARD/CLI/Yardoc

olleolleolle:
@iMacTia Right, perfect, this PR should _drop_ this file instead.

iMacTia (approved):

PR #1267, opened 2021-04-14
Add rubocop-inclusivity

olleolleolle:
## Description

This PR adds a linter addition: avoid some non-inclusive terms.


olleolleolle (commented):
This gem exists, and this will be run with the CI.

PR #1273, opened 2021-04-25
update usage docs (#1270)

gurgeous:
Took a stab at rewriting the Usage page. Changes include:

1. showing method signatures for HTTP verbs
2. give an overview of adapters & middleware
3. explain a bit more about the shortcut methods vs. creating a connection
4. use httpbingo.org instead of the defunct sushi.com
5. quick note about the default connection and url encoding

I have no ego about this change, feedback is appreciated.

### DIFF
--- a/docs/usage/index.md
+++ b/docs/usage/index.md
@@ -14,20 +14,20 @@ requiring the Faraday gem and using `Faraday.get`:
 ```ruby
 require 'faraday'
 
-resp = Faraday.get 'http://httpbingo.org'
+response = Faraday.get 'http://httpbingo.org'
 ```
 
 This returns a `Faraday::Response` object with the response status, headers, and
 body.
 
 ```ruby
-resp.status
+response.status
 # => 200
 
-resp.headers
+response.headers
 # => {"server"=>"Fly/c375678 (2021-04-23)", "content-type"=> ...
 
-resp.body
+response.body
 # => "<!DOCTYPE html><html> ...
 ```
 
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
These are wonderful changes, thank you!

When I read "resp" I thought I'd like the full word to be in the variable name, and when I read "bacon" I thought that we could spell it out fully as "chunky bacon" in order to honour _why the lucky stiff properly.

(But those are minor quibbles, easily changed later, if ever.)

I am glad we got to here: you filed an issue which became a discussion, which opened our eyes to starting troubles which could trip first-time users. And, now you have them covered, with this very-valuable PR.

gurgeous:
Aww, thank you for the kind words! Happy to help. We all gotta stick together in our little community.

olleolleolle:
@gurgeous Thanks again! 

gurgeous:
Thanks for your help with the usage page. Very pleased with the new content. I'd be happy to work on the Middleware page too. Some ideas:

1. List more middleware. Specifically thinking of cookie_jar, encoding, http-cache, etc. Any other important middleware we should list? https://rubygems.org/search?query=faraday. Perhaps a table?
1. Actually I think maybe the Available Middleware content should move onto the middleware page. This is a huge selling point for Faraday. Maybe the most important feature. Let's put it front and center! 😄 
1. `use` vs `request` vs `response`
1. Touch on a few things from the new usage page - unlocking the power of faraday via middleware, the default stack, etc.

PR #1283, opened 2021-05-20
default proxy to http:// if necessary, fixes #1282

### LINKED ISSUE #1282: Proxy silently ignored without `http:`
## Basic Info

* Faraday Version: 1.4.1
* Ruby Version: 2.7

## Issue description

Proxy silently ignored without `http:`. For example, if you set proxy to `something:1234` it will be silently ignored. The correct proxy setting would be `http://something:1234`. This is an important issue because it's undocumented, difficult to debug, and can have serious consequences for crawling projects.

I propose either:

1. In ProxyOptions.from, reject strings that lack a scheme.
or
1. In ProxyOptions.from, default to `http://` for strings that lack a scheme.

Let me know if you agree with one of these options and I'l put together a PR. Thanks!

## Steps to reproduce

```ruby
> Faraday::ProxyOptions.from("something:1234").host
# => nil

> Faraday::ProxyOptions.from("http://something:1234").host
# => "something"
```

gurgeous:
Previously, the proxy setting was silently ignored if you accidentally left off the scheme (`something:1234`). This change fills in the `http://` if it seems to be missing. Fixes #1282.

### DIFF
--- a/lib/faraday/options/proxy_options.rb
+++ b/lib/faraday/options/proxy_options.rb
@@ -19,6 +19,13 @@ module Faraday
           value[:uri] = Utils.URI(uri)
         end
       end
+
+      # URIs without a scheme should default to http (like 'example:123'). This
+      # fixes #1282 and prevents a silent failure in some adapters.
+      if value && !value[:uri].to_s.include?('://')
+        value[:uri] = Utils.URI("http://#{value[:uri]}")
+      end
+
       super(value)
     end
 
--- a/lib/faraday/options/proxy_options.rb
+++ b/lib/faraday/options/proxy_options.rb
@@ -11,6 +11,9 @@ module Faraday
     def self.from(value)
       case value
       when String
+        # URIs without a scheme should default to http (like 'example:123').
+        # This fixes #1282 and prevents a silent failure in some adapters.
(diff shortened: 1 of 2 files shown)


lib/faraday/options/proxy_options.rb
```diff
@@ -19,6 +19,13 @@ def self.from(value)
           value[:uri] = Utils.URI(uri)
         end
       end
+
+      # URIs without a scheme should default to http (like 'example:123'). This
+      # fixes #1282 and prevents a silent failure in some adapters.
+      if value && !value[:uri].to_s.include?('://')
+        value[:uri] = Utils.URI("http://#{value[:uri]}")
+      end
```
iMacTia, line 27: What do you think of having this check only in the `case when String` branch?
That looks like the most common use case, and if a `URI` or `Options` object is provided then probably the caller knows what they're doing (leaving the door open to override this automatic fix, if necessary)

gurgeous: Good idea - how about this?

iMacTia (approved):

iMacTia:
Thanks @gurgeous, this looks great! 🎉

PR #1288, opened 2021-06-23
Always dup url_prefix in Connection#build_exclusive_url

### LINKED ISSUE #1287: Connection url is shared between requests causing concurrency issues
## Basic Info

* Faraday Version: 1.3, 1.4
* Ruby Version: any

## Issue description
In a certain setup(provided below) requests made with the same connection are using shared URI instance which in my case had caused a race condition when two simultaneous requests with different query params were actually executed with the same query params resulting in very unexpected behavior.

Example:
```rb
conn = Faraday.new("http://google.com/users")

response1 = conn.get(nil, { id: 1 })
response2 = conn.get(nil, { id: 2 })

response1.env[:url] #=> "http://google.com/users?id=2"
response1.env[:url] === response2.env[:url] #=> true
```

alexeyds:
## Description
Fixes #1287

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -520,9 +520,8 @@ module Faraday
     # @return [URI]
     def build_exclusive_url(url = nil, params = nil, params_encoder = nil)
       url = nil if url.respond_to?(:empty?) && url.empty?
-      base = url_prefix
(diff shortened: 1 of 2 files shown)


iMacTia (approved):
Thanks for investigating and fixing this! LGTM

PR #1290, opened 2021-07-04
Use external patron adapter

iMacTia:
## Description
Use external patron adapter


olleolleolle (approved):
Thanks for also repairing the linting!

PR #1274, opened 2021-04-26
Use external em_http and em_synchrony adapters

iMacTia:
## Description
Replace the internal `Em::HTTP` and `EM::Synchrony` adapters with the new gems

PR #1294, opened 2021-07-10
restore jruby support

ahorek:
## Description
JRuby doesn't support em-synchrony (existing tests are already skipped)

unfortunately, after https://github.com/lostisland/faraday/pull/1274 faraday can't boot up anymore even with a different adapter.

this PR skips loading these runtime dependencies on JRuby.

I've also increased the version of jruby-openssl which is quite outdated.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 
 ruby RUBY_VERSION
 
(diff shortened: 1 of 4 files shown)


iMacTia:
Thanks @ahorek, none of the core maintainers are JRuby users/experts, so we really appreciate help on this even though JRuby is not officially supported

iMacTia (approved):

PR #1298, opened 2021-08-01
Add strict_mode to Test::Stubs

### LINKED ISSUE #1291: Feature request: how about making `verify_stubbed_calls` verify the parameters of stubbed requests?
`verify_stubbed_calls` is a useful method to stub whether a request is exactly called within a production code.
However, I found out it does not verify the parameters of a request though it does for the request path.

For example, I'm expecting this test to **fail** because the new parameter `foo` is added to `sushi`, but the stubbed Faraday connection is expecting that the only `abc` exists in its params. However, the current implementation seems to pass the test.

```ruby
class Client
  def initialize(conn)
    @conn = conn
  end

  def sushi(jname, **params)
    res = @conn.get("/#{jname}", **params)
    data = JSON.parse(res.body)
    data['name']
  end
end

RSpec.describe Client do
  let(:stubs)  { Faraday::Adapter::Test::Stubs.new }
  let(:conn)   { Faraday.new { |b| b.adapter(:test, stubs) } }
  let(:client) { Client.new(conn) }

  it 'verifies the all parameter values are identical' do
    stubs.get('/ebi?abc=123') do # <-- This expects only `abc` added, not including `foo`. Currently, this test unexpectedly succeeds!
      [
        200,
        { 'Content-Type': 'application/javascript' },
        '{"name": "shrimp"}'
      ]
    end

    expect(client.sushi('ebi', abc: 123, foo: 'Kappa')).to eq('shrimp')
    stubs.verify_stubbed_calls
  end
end
```

I succinctly came up with an idea to check all parameter keys are identical to the request parameters like this, but it would cause the existing tests to fail because of such a **strict** stubbing.

```diff
diff --git a/lib/faraday/adapter/test.rb b/lib/faraday/adapter/test.rb
index 0230b85..739b7ec 100644
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -184,6 +184,8 @@ module Faraday
         end

         def params_match?(request_params)
+          return false unless Set.new(params.keys) == Set.new(request_params.keys)
+
           params.keys.all? do |key|
             request_params[key] == params[key]
           end
```

What do you think? This idea might bring breaking changes but would be useful when it comes to verifying the parameters.

yykamei:
## Description

Close #1291

Sometimes, users might want to check whether stubbed requests exactly
match the parameters and headers. In this case, they expect
`Test::Stubs` to raise an error when they add a new parameter in a
request. This patch introduces `strict_mode` for the purpose of the
expectation. With `strict_mode`, `Test::Stubs` tries to check all a
path, parameters, and headers exactly.

## Todos

- [x] Tests
- [x] Documentation


iMacTia (approved):
Nice work implementing this and great test coverage!
Apologies about the mess with commits, but we recently started the works on Faraday v2.0 and I got confused 😅. I've reverted those changes and changed the base branch to the correct one (`1.x`) 👍.

LGTM 🎉 !

yykamei:
Oh, thank you! I should have opened a PR into `1.x` 😅 
Anyway, thank you for reviewing so quickly!

PR #1311, opened 2021-08-16
Add latest rubocop and use Ruby 2.6+

olleolleolle:
## Description

This introduces the latest RuboCop version, and sets lower bound of Ruby support to 2.6.

This does the absolute minimum, to get it configured.

- We landed _some_ new rules, perhaps we'll go with "all new rules", soon, I don't know.
- The TODO file grew a small amount.

See #1309 

## Todos

- [x] Add the new set of defaults
- [x] Re-run the auto-generated todo file

## Additional Notes

We will see if there are many separate steps to take, if we need a big checklist & issues etc.


Gemfile
```diff
@@ -19,7 +19,7 @@ group :development, :test do
 end
 
 group :development, :lint do
-  gem 'rubocop', '~> 0.90.0'
+  gem 'rubocop', '~> 0.90'
```
mattiagiuffrida-st, line 22: This won't install Rubocop 1.0+ though, should we take this opportunity to go `~> 1.19`?

.rubocop_todo.yml
```diff
@@ -1,32 +1,91 @@
...
+# Offense count: 11
+# Configuration parameters: AllowComments, AllowEmptyLambdas.
+Lint/EmptyBlock:
+  Exclude:
+    - 'spec/faraday/connection_spec.rb'
+    - 'spec/faraday/rack_builder_spec.rb'
+    - 'spec/faraday/response_spec.rb'
+
+# Offense count: 1
+# Cop supports --auto-correct.
```
iMacTia, line 28: This and other cops added to the `.rubocop_todo.yml` support `--auto-correct`.
Is there a reason why we don't simply let Rubocop auto-correct them?

iMacTia (commented):
Fantastic work @olleolleolle and well done keeping the todo file under control 👏 
I just have a question about auto-correctable cops

.rubocop_todo.yml
olleolleolle: I just wanted to stay green, while doing it, let's see if we can get this done before closing the lid for today!

olleolleolle: @iMacTia There! Now those are out of the file, and handled.

.rubocop_todo.yml
```diff
@@ -1,32 +1,57 @@
...
+  Max: 33
 
-# Offense count: 9
+# Offense count: 8
 # Configuration parameters: IgnoredMethods.
 Metrics/PerceivedComplexity:
   Max: 14
+
+# Offense count: 3
+Style/DocumentDynamicEvalDefinition:
```
olleolleolle, line 53: I want to talk a bit about this rule, since the fix will have visual impact, and we'd like to pick the nicest alternative.

iMacTia (approved):
This is good as is IMO, we can discuss the pending points offline and address in a follow-up

PR #1315, opened 2021-08-21
Automatically refresh website Team page on commit to default branch

iMacTia:
## Description
According to [the doc](https://docs.github.com/en/rest/reference/repos#a-word-about-caching), the `stats` endpoint is quite heavy and it's content is cached.
We rely on this endpoint to render our [team page](https://lostisland.github.io/faraday/team/), however the cache is invalidated after each commit to `main` and it takes several minutes to rebuild it.
This process is only kicked off on the first visit to the team page, rendering it unavailable for some users.

Solution, as part of our CI build, only in case of a push to `main`, we should make an API call to trigger the cache rebuild.
This will increase the chances of the cache being available when someone visits the team page.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,6 @@ env:
   COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
 
 jobs:
-  refresh_stats:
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Looks very clear on a first read.

PR #1316, opened 2021-08-28
Respect the `params_encoder` in `Faraday::Adapter::Test`

### LINKED ISSUE #1312: Feature request: support `Faraday::FlatParamsEncoder` in `Faraday::Adapter::Test`
#1291 introduced `strict_mode` in `Faraday::Adapter::Test::Stubs`, and I want more features for the mode: supporting `Faraday::FlatParamsEncoder`. For example, if I want to call an HTTP request with the same query parameter `a` multiple times, I expect the test stub to check whether the same parameters with different values are all correctly passed.

```ruby
class Client
  def initialize(conn)
    @conn = conn
  end

  def sushi(jname, params: {})
    res = @conn.get("/#{jname}", params)
    data = JSON.parse(res.body)
    data['name']
  end
end

RSpec.describe Client do
  let(:stubs) { Faraday::Adapter::Test::Stubs.new(strict_mode: true) }
  let(:conn)   { Faraday.new(request: { params_encoder: Faraday::FlatParamsEncoder }) { |b| b.adapter(:test, stubs) } }
  let(:client) { Client.new(conn) }

  it 'handles the same multiple URL parameters' do
    # The same parameter `a` is expected to appear multiple times with different values.
    stubs.get('/ebi?a=x&a=y&a=z') { [200, { 'Content-Type' => 'application/json' }, '{"name": "shrimp"}'] }

    # Test should pass
    expect(client.sushi('ebi', params: { a: %w[x y z] })).to eq('shrimp')

    # Test should not pass due to the lack of "x" and "y" for `a`
    # expect(client.sushi('ebi', params: { a: %w[z] })).to eq('shrimp')
    stubs.verify_stubbed_calls
  end
end
```

What do you think? I want to open a pull request if it makes sense.

yykamei:
Close #1312 

Faraday provides a feature to let its callers configure
`params_encoder`, but `Faraday::Adapter::Test` had always created test
stubs with `#parse_nested_query` (`NestedParamsEncoder`). So, we had not
able to test such a connection with `Faraday::Adapter::Test`. This might
be inconvenient for users who want to use non-default `params_encoder`,
such as `Faraday::FlatParamsEncoder`.

This patch changed the adapter to use the same `params_encoder` as its
connection in order to solve the problem.

## Todos

- [x] Tests
- [x] Documentation (I thought the update of documentation is not required in this PR)

### DIFF
--- a/examples/client_test.rb
+++ b/examples/client_test.rb
@@ -13,8 +13,8 @@ class Client
     @conn = conn
   end
 
-  def sushi(jname)
-    res = @conn.get("/#{jname}")
+  def sushi(jname, params: {})
+    res = @conn.get("/#{jname}", params)
     data = JSON.parse(res.body)
     data['name']
   end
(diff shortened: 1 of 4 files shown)


examples/client_test.rb
```diff
@@ -70,6 +70,45 @@ def test_sushi_exception
...
+      ]
+    end
+
+    cli = client(stubs)
+    assert_equal 'shrimp', cli.sushi('ebi', params: { abc: 123 })
+
+    # uncomment to raise Stubs::NotFound
+    # assert_equal 'shrimp', cli.sushi('ebi', params: { abc: 123, foo: 'Kappa' })
+    stubs.verify_stubbed_calls
+  end
```
yykamei, line 89: I forgot to add this test in #1298 😅

lib/faraday/adapter/test.rb
```diff
@@ -239,26 +233,19 @@ def configure
...
                                  "#{normalized_path} #{env[:body]}"
         end
 
-        env[:params] = if (query = env[:url].query)
-                         params_encoder.decode(query)
-                       else
-                         {}
-                       end
+        params_encoder = env.request.params_encoder || Faraday::Utils.default_params_encoder
+        env[:params] = params_encoder.decode(env[:url].query) || {}
```
iMacTia, line 248: Mmmmh I really like how much simpler the `matches?` methods signature are now, thanks to the `env` being passed, but I notice there's a lot a of repeated code between here and there. These two lines for example, and line 242 as well.

I wonder if instead of reusing `env` it might make sense to introduce a new `Struct`, internal and only used in the Test adapter, that holds this information so we only need to calculate it once. What do you think?

iMacTia (changes requested):
Nice work @yykamei, I left a comment with a suggestion on how to keep the code cleaner and more maintainable. I'm scared that the duplication introduced might cause issues in the future when only one side is updated and not the other.

I'm also open to different solutions that solve the same issue in different ways, mine is only an idea

lib/faraday/adapter/test.rb
yykamei: Thank you for your quick review! It makes sense 👍 I'll try it.

yykamei: I tried to avoid similar codes with 91689f0. How about the change?

iMacTia (approved):
Thanks for the changes! Yes the code is definitely cleaner 👏 
LGTM 🎉

Issue #1317, opened 2021-08-31
WARNING: `Faraday::Connection#basic_auth` is deprecated


mscoutermarsh:
## Basic Info

* Faraday Version: `1.7.1`
* Ruby Version: `3.0.2`

## Issue description
Saw the deprecation warning, updated code to match the [documented basic auth instructions](https://lostisland.github.io/faraday/middleware/authentication). The documented code throws a number of arguments error. 

Not sure if the docs are wrong or it's an issue with the gem.

## Steps to reproduce
```ruby
http = Faraday.new do |conn|
  conn.request(:authorization, :basic, 'username', 'password')
end

http.post("https://example.com")
ArgumentError: wrong number of arguments (given 4, expected 3)
from /Users/mscoutermarsh/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.7.1/lib/faraday/request/authorization.rb:43:in `initialize'
```

peterberkenbosch:
The docs are wrong :) I just opened a PR to fix that. 

In the meantime you need to use it like this:

```ruby
http = Faraday.new do |conn|
  conn.request(:basic_auth, 'username', 'password')
end
```

olleolleolle:
Hello! The context for the documentation website not rhyming with the 1.x deprecation message is: "all the website documents is the upcoming 2.0". That version where the old way is removed. 

The context is only ever mentioned only in the README - this is a weakness, which we'd love to get concrete help with on the website front.

iMacTia:
@peterberkenbosch is correct, that's the right way to use the middleware in Faraday 1.x, while the documentation is showing the correct usage for the upcoming v2.0.

Sorry about the confusion, this is not the first time an issue like this has been raised. I'll be looking at a quick solution to make things clearer

mscoutermarsh:
Thank you! All good now. ❤️

xiaoxipang:
Hello @iMacTia,

I still have problem on this issue.
I'm using Faraday 1.7.1 and Ruby 2.7.2. 

I originally created a connection like this
```
client = Faraday.new(url: url) do |faraday|
    faraday.headers["Accept"] = "application/json;"
    faraday.headers["X-Version"] = "2"
    faraday.headers["Authorization"] = "token #{token}"
    faraday.request(:json)
    faraday.response(:json, content_type: /\bjson$/)
    faraday.adapter(Faraday.default_adapter)
  end
```

[...]

PR #1318, opened 2021-08-31
Update documentation using basic auth middleware

peterberkenbosch:
Documentation on using basic auth is incorrect. The deprecation warning shows the right information though. This PR updates the documentation so it matches the deprecation warning and works as expected.

When using the current documented basic auth middleware it will render the following error: 

```
ArgumentError: wrong number of arguments (given 4, expected 3)
```

### DIFF
--- a/docs/middleware/request/authentication.md
+++ b/docs/middleware/request/authentication.md
@@ -34,6 +34,6 @@ The middleware will automatically Base64 encode your Basic username and password
 
 ```ruby
 Faraday.new(...) do |conn|
-  conn.request :authorization, :basic, 'username', 'password'
+  conn.request :basic_auth, 'username', 'password'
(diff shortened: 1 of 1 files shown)


iMacTia:
@peterberkenbosch thanks for trying to fix this and apologies about the confusion 🙏 
As @olleolleolle explained in his [comment](https://github.com/lostisland/faraday/issues/1317#issuecomment-909951983), the documentation website is currently built based on the `main` branch, which already incorporates the v2.0 changes.
That's why the example is incorrect for people using Faraday 1.x.

This is not the first report and we're exploring how to effectively tackle this issue as our current tools don't allow us to have version switching in the docs.

I'll be closing this but will try to put something temporary in place to avoid further confusion

PR #1322, opened 2021-09-09
Backport authorization procs to 1.x branch

jarl-dk:
## Description

This is a backport  of authorization using `Proc`s. It is a simple cherry-pick from `main` branch.

(cherry picked from commit fe600c72adc79180ad7e545191edd5dd2fe61990)

## Todos
List any remaining work that needs to be done, i.e:
- [x] Tests
- [x] Rubocop
- [x] Documentation

## Additional Notes
This is a backport  of authorization using `Proc`s. It is a simple cherry-pick from `main` branch.
(cherry picked from commit fe600c72adc79180ad7e545191edd5dd2fe61990)

PR #1326, opened 2021-09-17
Proc auth value backport cleanup

iMacTia:
## Description
This PR is a follow-up to #1322 to adapt the implementation to the 1.x branch

cc @jarl-dk, who worked on the original PR 🙏

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        sudo apt-get udpate
+        sudo apt-get update
         sudo apt-get install libcurl4-openssl-dev
 
     - name: Build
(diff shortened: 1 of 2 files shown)


iMacTia:
Not sure if this is a blip from Ubuntu packages or what, but I'm getting the following:

> E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/c/curl/libcurl4-openssl-dev_7.68.0-1ubuntu2.6_amd64.deb
> 404  Not Found [IP: 52.147.219.192 80]

I'll try waiting a bit and then re-run tests. Good news is that they're all passing locally 🎉

iMacTia:
And it's green 🟢 🎉 !

olleolleolle (approved):
This adapts the solution to be 1.x-ready. Thanks, everyone involved!

jarl-dk:
Regarding ubuntu; It could be that it was a one-off server down issue.

iMacTia:
Thanks for the review @olleolleolle 🎉!

@jarl-dk in case this might be useful, we were missing `apt-get update` before `install`, so it likely was an outdate package source 🤦‍♂️.

PR #1335, opened 2021-10-15
Improve request info in exceptions raised by RaiseError Middleware

### LINKED ISSUE #1249: Faraday::Error does not provide original request URL or params
## Basic Info

* Faraday Version: 1.3.0
* Ruby Version: 3.0.0

## Issue description

When the `raise_error` middleware is used, the resulting `Faraday::Error` object provides no access to the original request URL that triggered the error. This makes it hard to troubleshoot errors and attribute an error to a problematic URL.

## Steps to reproduce

```ruby
require "faraday"

faraday = Faraday.new("https://example.com/") do |config|
  config.response :raise_error
end

begin
  faraday.get("wut", "q" => "hello")
rescue Faraday::Error => e
  puts e.response[:request].inspect
end
```

Output:

```ruby
{:method=>:get, :url_path=>"/wut", :params=>nil, :headers=>{"User-Agent"=>"Faraday v1.3.0"}, :body=>nil}
```

Note that `:params` is `nil`, even though I passed `"q" => "hello"` as params in the `.get` request. Also there is a `:url_path` but no full URL that includes the query string.

willianzocolau:
## Description
A few sentences describing the overall goals of the pull request's commits.
Link to related issues if any. (As `fixes #XXX `)

Fixes #1249

Add URI object with the url requested and fix query string params. In [test adapter](https://github.com/lostisland/faraday/blob/main/lib/faraday/adapter/test.rb#L240), the params already was treated which means the specs was biased. With default adapter, the params problem occurred as described in Issue #1249.

## Todos
- [X] Tests
- [X] Documentation

## Additional Notes
Optional section

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -52,6 +52,7 @@ module Faraday
     #              :body    - Optional string HTTP response body.
     #              :request - Hash
     #                           :method   - Symbol with the request HTTP method.
+    #                           :url      - URI object with the url requested.
     #                           :url_path - String with the url path requested.
     #                           :params   - String key/value hash of query params
     #                                     present in the request.
--- a/lib/faraday/response/raise_error.rb
+++ b/lib/faraday/response/raise_error.rb
@@ -44,12 +44,18 @@ module Faraday
           body: env.body,
           request: {
             method: env.method,
+            url: env.url,
             url_path: env.url.path,
-            params: env.params,
+            params: query_params(env),
             headers: env.request_headers,
             body: env.request_body
           }
         }
+      end
+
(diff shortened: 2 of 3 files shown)


iMacTia (approved):
Hi @willianzocolau and thanks for fixing this!

I just want to point out that this PR will be merged on the `main` branch, and that currently contains the code that will go in the next major release (v2.0).

If you'd like to have this change released for Faraday v1 as well, then you'll need open another PR against the `1.x` branch. This is absolutely not a requirement though, so feel free to ignore if you're planning to use v2.0 anyway.

PR #1341, opened 2021-11-13
Add Truffleruby to CI

gogainda:


olleolleolle (approved):
Thanks for adding this, happy to see it green.

iMacTia:
I'm OK having it in for now, considering it's green and all tests pass. But I'm not a truffleruby user, so it will be hard to maintain compatibility without some help, therefore let's consider it experimental support for the time being 🙂. If things get hard, we might make this an optional matrix run.

Issue #1248, opened 2021-03-15
Is there a way to remove Faraday's User-Agent header as a default option for all connections?


dmytro-savochkin:
## Basic Info

* Faraday Version: 1.3.0
* Ruby Version: 2.7.2

## Issue description
Is there a way to prevent Faraday from sending its User-Agent header (`Faraday v1.3.0`)? I need it because one of the external servers we are dependent upon has banned it (https://github.com/lostisland/faraday/issues/1247#issuecomment-799345338).
Well, I know how to rewrite this header with something else per connection. Also I found how to rewrite it with something else as a default but only when there are no headers added when creating a connection (the new headers overwrite the default ones):

```
Faraday.default_connection_options = { headers: { user_agent: 'Something 1.2.3' } }
headers = { 'CustomHeader' => 'CustomValue' }
connection = Faraday.new('https://www.google.com', headers: headers) do |f|
  f.response :logger, nil, { headers: true }
end
connection.get

I, [2021-03-15T13:41:49.910701 #7925]  INFO -- request: GET https://www.google.com/
I, [2021-03-15T13:41:49.910993 #7925]  INFO -- request: CustomHeader: "CustomValue" User-Agent: "Faraday v1.3.0"
```

But I cannot find a way as how to remove that header completely (except for monkey-patching of course). Is there one? 

Thank you!

iMacTia:
Hi @dmytro-savochkin, you're right this is not currently possible because passing the `headers` option to the connection will override any `headers` options set in `Faraday.default_connection_options`.
So as you said, a temporary solution would be to override the header on each connection, but I appreciate this is not ideal when you have many of them (although it's rarely a good idea to have many connections, to be honest).

In order to solve this on a global level, I can see 2 possible solutions:

1. We change the way we merge connection options with default_connection_options by performing a `deep_merge` (there's a helper in `Faraday::Utils` so we don't need to pull `ActiveSupport` in). This will solve the problem as `headers` won't be replaced, but merged gracefully. I'd very much prefer this solution, but I'm unsure on the backwards-compatibility impact
2. We introduce a new `Faraday.default_user_agent` setting that will default to the current value, but can also be overridden and, if necessary, even set to `nil` to remove the header completely.

@olleolleolle your input on the above?

dmytro-savochkin:
Just FYI, I ended monkey-patching Faraday for now to stop it from setting a User-Agent value.
[...]

PR #1343, opened 2021-11-29
Replace Hash#merge with Utils#deep_merge for connection options

xkwd:
## Description
This is an attempt to close the https://github.com/lostisland/faraday/issues/1248 issue by following the proposed solution `#1` in [this comment](https://github.com/lostisland/faraday/issues/1248#issuecomment-799362164). 

Prior to this change:
```ruby
Faraday.default_connection_options = { headers: { user_agent: 'My Agent 1.2' } }
Faraday.new('http://example.co', headers: { 'CustomHeader' => 'CustomValue' }).headers 
# => { 'CustomHeader' => 'CustomValue', 'User-Agent' => 'Faraday v2.0.0.alpha-2' }
Faraday.new(url: 'http://example.co', headers: { 'CustomHeader' => 'CustomValue' }).headers 
# => { 'CustomHeader' => 'CustomValue', 'User-Agent' => 'Faraday v2.0.0.alpha-2' }
```
After this change:
```ruby
Faraday.default_connection_options = { headers: { user_agent: 'My Agent 1.2' } }
Faraday.new('http://example.co', headers: { 'CustomHeader' => 'CustomValue' }).headers 
# => { 'CustomHeader' => 'CustomValue', 'User-Agent' => 'My Agent 1.2' }
Faraday.new(url: 'http://example.co', headers: { 'CustomHeader' => 'CustomValue' }).headers 
# => { 'CustomHeader' => 'CustomValue', 'User-Agent' => 'My Agent 1.2' }
```

The only issue with the proposed solution was that the `deep_merge` from `Faraday::Utils` didn't work out of the box for an `Options Struct` like `RequestOptions` and `SSLOptions` from `Faraday::ConnectionOptions`.

### DIFF
--- a/spec/faraday/connection_spec.rb
+++ b/spec/faraday/connection_spec.rb
@@ -639,6 +639,24 @@ RSpec.describe Faraday::Connection do
 
       it_behaves_like 'default connection options'
     end
+
+    context 'preserving a user_agent assigned via default_conncetion_options' do
+      context 'when url is a Hash' do
+        let(:conn) { Faraday.new(url: 'http://example.co', headers: { 'CustomHeader' => 'CustomValue' }) }
+
+        before { Faraday.default_connection_options = { headers: { user_agent: 'My Agent 1.2' } } }
+
+        it { expect(conn.headers).to eq('CustomHeader' => 'CustomValue', 'User-Agent' => 'My Agent 1.2') }
+      end
+
+      context 'when url is a String' do
+        let(:conn) { Faraday.new('http://example.co', headers: { 'CustomHeader' => 'CustomValue' }) }
+
+        before { Faraday.default_connection_options = { headers: { user_agent: 'My Agent 1.2' } } }
+
+        it { expect(conn.headers).to eq('CustomHeader' => 'CustomValue', 'User-Agent' => 'My Agent 1.2') }
+      end
+    end
   end
 
   describe 'request params' do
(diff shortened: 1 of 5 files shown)


iMacTia (approved):
Thank you @xkwd, this change is very much welcome!
The biggest issue I had with this solution, as I mentioned in my comment on the issue, was that this is effectively backwards-incompatible and might cause surprise for users.

Good news though is that we're in alpha testing for Faraday 2.0 🎉! So I'm happy to merge this is in and simply drop a comment on the `UPGRADING.md` guide 🙌 

The only downside is that if you need this in one of your projects, you'll need to upgrade to Faraday 2.0

spec/faraday/connection_spec.rb
```diff
@@ -639,6 +639,24 @@ def decode(params)
 
       it_behaves_like 'default connection options'
     end
+
+    context 'preserving a user_agent assigned via default_conncetion_options' do
+      context 'when url is a Hash' do
+        let(:conn) { Faraday.new(url: 'http://example.co', headers: { 'CustomHeader' => 'CustomValue' }) }
+
+        before { Faraday.default_connection_options = { headers: { user_agent: 'My Agent 1.2' } } }
```
iMacTia, line 647: I just realised there no rollback of the original value after these test have run 😄.
I'll be fixing this on the main branch, so no worries at all, but please keep this in mind next time you write tests for a global configuration 👍

Issue #1344, opened 2021-12-10
Ability to pass a callable to authorization middleware


sled:
The authorization middleware currently checks for a `Proc`:
https://github.com/lostisland/faraday/blob/746ab737b5d5bf3bcd28f377044f7bec03f8286d/lib/faraday/request/authorization.rb#L40

Wouldn't it make sense to ease this restriction and just require a callable? 

Example use case:

```rb
class ComplexAuthorizer
  def call
     # do magic and return a token
  end
end

authorizer = ComplexAuthorizer.new(....)

Faraday.new(url: ....) do |conn|
  conn.request :authorization, 'Bearer', authorizer
end

```

iMacTia:
That would make sense indeed, I'd happily review a PR that enhances the middleware to accept either a `proc` or a callable 👍

sled:
> That would make sense indeed, I'd happily review a PR that enhances the middleware to accept either a `proc` or a callable 👍

Here you go: https://github.com/lostisland/faraday/pull/1345

PR #1345, opened 2021-12-12
Callable authorizers

sled:
## Description
Allows callable authorizers to be passed as proposed in #1344

### DIFF
--- a/lib/faraday/request/authorization.rb
+++ b/lib/faraday/request/authorization.rb
@@ -37,7 +37,7 @@ module Faraday
           raise ArgumentError, "Unexpected params received (got #{params.size} instead of 1)"
         else
           value = params.first
-          value = value.call if value.is_a?(Proc)
+          value = value.call if value.respond_to?(:call)
           "#{type} #{value}"
         end
       end
--- a/lib/faraday/request/authorization.rb
+++ b/lib/faraday/request/authorization.rb
@@ -8,10 +8,11 @@ module Faraday
 
       # @param app [#call]
       # @param type [String, Symbol] Type of Authorization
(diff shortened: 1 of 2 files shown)


lib/faraday/request/authorization.rb
```diff
@@ -37,7 +38,7 @@ def header_from(type, *params)
           raise ArgumentError, "Unexpected params received (got #{params.size} instead of 1)"
         else
           value = params.first
-          value = value.call if value.is_a?(Proc)
+          value = value.call if value.is_a?(Proc) || value.respond_to?(:call)
```
sled, line 41: The `||` is on purpose, because checking the ancestors of an object should be faster than looking up a method. This way, there should be no performance penalty on existing code bases which use a `proc` already.

spec/faraday/request/authorization_spec.rb
```diff
@@ -63,6 +63,15 @@
       include_examples 'does not interfere with existing authentication'
     end
 
+    context 'when passed as callable' do
```
olleolleolle, line 66: Minor: this is about _a_ callable, right?

```suggestion
    context 'when passed a callable' do
```

olleolleolle (approved):
Thanks for the neat test-case and implementation!

olleolleolle (approved):
Thanks for putting this together!

lib/faraday/request/authorization.rb
iMacTia: Love this attention to performance, thanks for caring and explaining!

iMacTia (approved):
LGTM 💯 🎉

PR #1351, opened 2021-12-20
Default value for exc error

DariuszMusielak:
## Description
I would like to introduce default value for `exc` in `def initialize` for `Faraday::Error`. 

In some cases it is difficult to find the reason on a first spot when doing `raise Faraday::ConnectionFailed` because it raises `ArgumentError`. In other cases like `raise Faraday::TimeoutError` it works because `Faraday::TimeoutError` sets default value or `exc`. 

Link to related issues if any. (As `Fixes #XXX`) - there was no related issue. 
I found out that someone has been asking why this is happening on Stackoverflow, I spent some time today looking for an answer for myself. 

From my perspective introducing default value does not break anything and simplify codebase. 

## Todos
- [x] Tests
- [x] Documentation

## Additional Notes
Optional section
Stackoverflow: https://stackoverflow.com/questions/59176256/unexpected-behaviour-with-faradayconnectionfailed/70427415#70427415

### DIFF
--- a/docs/adapters/testing.md
+++ b/docs/adapters/testing.md
@@ -33,7 +33,7 @@ conn = Faraday.new do |builder|
 
     # test exceptions too
     stub.get('/boom') do
-      raise Faraday::ConnectionFailed, nil
+      raise Faraday::ConnectionFailed
     end
   end
 end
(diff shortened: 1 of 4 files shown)


iMacTia (approved):
I couldn't find a reason for why the `exc` parameter was enforced. We may argue it's not a good idea to raise an exception without useful information in it, but going as far as making it manadatory makes testing more complicated than necessary (as this PR shows fixing our own tests).

LGTM 💯

PR #1357, opened 2022-01-02
Remove multipart middleware and all its documentation and tests

iMacTia:
## Description
Remove multipart middleware and all its documentation and tests.
Multipart middleware has been moved to a [separate repo](https://github.com/lostisland/faraday-multipart).

## Todos
- [x] Tests
- [x] Documentation


olleolleolle:
Perhaps this could be in UPGRADE.md or something, how to include the new gem in the Gemfile, and how to include the file required.

olleolleolle (approved):
Removal looks ace, I made a note about "perhaps this one is a thing to teach somewhere, for people upgrading".

iMacTia:
@olleolleolle good point, I agree they deserve their own section for clarity. I've just added that 👍

olleolleolle (approved):
LGTM!

ahmedakef:
@iMacTia @olleolleolle it is listed in the change log of `v 2.0` while it is merged into `v 1.9.0` release https://github.com/lostisland/faraday/releases/tag/v1.9.0

it this expected ?

iMacTia:
@ahmedakef that's correct, this change was applied to both version branches.
It was backported to 1.9.0 as part of #1367, while this PR made the change to the `main` branch (v2.0)

ahmedakef:
@iMacTia but isn't this a breaking change ?

iMacTia:
@ahmedakef oh no we'd never do that!

In #1367 we removed the middleware code from Faraday, but we added both `faraday-multipart` and `faraday-retry` as direct dependencies, which basically "replace" the remove code and provide backwards compatibility.
The idea is that you can upgrade to 1.9.0 and everything should still work exactly the same.

The real breaking change is in 2.0 where we remove those dependencies, which mean you'll now need to manually add them to the Gemfile if you want to use them.

In fact, the changelog for 1.9.0 says "Use external multipart and retry middleware", while the one for 2.0 is "Remove multipart middleware and all its documentation and tests".

iMacTia:
Ah thanks for reporting this @ahmedakef.
That seems a problem on the `google-api-ruby-client` gem though, because there's no need for requiring middleware thanks for Faraday's dependency autoloading feature.
Removing the require should make the code working across both versions. You can try this on your fork and then open a PR on their gem if it works as expected (as it should, unless something else is causing that not to work for any reason).

iMacTia:
Ah, I see they're referencing it directly here: https://github.com/googleapis/google-api-ruby-client/blob/v0.8-main/lib/google/api_client/request.rb#L326

That feels really wrong, they're completely bypassing the middleware this way and calling it explicitly, it doesn't really make sense...

PR #1370, opened 2022-01-06
Update adapter dependencies in Gemspec

iMacTia:
## Description
Adapters v2 have hard dependencies, hence can't be used with Faraday v1


olleolleolle (approved):
Good catch!

PR #1371, opened 2022-01-06
Re-add support for Ruby 2.4+

### LINKED ISSUE #1369: Faraday 1.9 broke the ruby 2.5 compatibility
## Basic Info

* Faraday Version: 1.9.0
* Ruby Version: 2.5.x

## Issue description
Version 1.9.0 bumped the minimum ruby constraint breaking the compatiblity with ruby 2.5
I understand that 2.5 is EOL, but such end of support is equivalent to a breaking change IMHO.
Was the bump of the constraint important/mandatory?
If not, do you think it would be possible to yank/republish a 1.9 version (or a 1.9.1) with 2.5 support?

Thanks in advance

iMacTia:
## Description
Re-add support for Ruby 2.4+
Fixes #1369

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.6', '2.7', '3.0', '3.1']
+        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0', '3.1']
 
     steps:
     - uses: actions/checkout@v1
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ require:
 AllCops:
   DisplayCopNames: true
   DisplayStyleGuide: true
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 2.4
 
 Inclusivity/Race:
   Enabled: true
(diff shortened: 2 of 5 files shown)


olleolleolle (approved):
👍

iMacTia:
Will need to re-release `faraday-multipart` and `faraday-retry` to support Ruby 2.4+ or Ruby 2.5+, but I'll leave this for later as it's not as urgent

Justin-W:
> Will need to re-release `faraday-multipart` and `faraday-retry` to support Ruby 2.4+ or Ruby 2.5+, but I'll leave this for later as it's not as urgent

@iMacTia The removal of support for Ruby v2.4 from `faraday-retry` is currently breaking some of our automation. We'd like to see that regression fixed ASAP.

ccdredzik:
@iMacTia It is interesting that some do update their faraday library, but can't do so for the ruby version they're running. 2.4 is dead, so is 2.5. Don't resurrect them.

iMacTia:
I'd normally agree but since I messed up a bit with the releases yesterday and that it didn't cost me much in the end, I re-waded both in a fix release for 1.9.

That said, Faraday 2.0 will not support anything below 2.6 and we will drop EOL'ed support, even on minor version bumps, if it becomes a problem to support them

PR #1372, opened 2022-01-06
Add alias with legacy name to gemified middleware

iMacTia:
## Description
Add alias with legacy name to gemified middleware
Fixes https://github.com/lostisland/faraday/issues/1369#issuecomment-1006769658


olleolleolle (approved):
Another one! Good catch!

connorshea (approved):
Thank you!

PR #1379, opened 2022-01-12
Introduce mutex Monitor in Faraday::Test::Stubs

### LINKED ISSUE #1365: Faraday::Adapter::Test::Stubs is not thread safe
## Basic Info

* Faraday Version: 0.13.1 (via Debian 10 "Buster")
* Ruby Version: 2.5.5

(I'm aware this version is *quite* far behind the most recent version, but the sections of code I will try to highlight below appear to still be relevant, and the thread-unsafety should be verifiable by inspection.)

## Issue description
Encountered while investigating a low repro rate unit test failure (< 0.1%) in our build system.
The test would fail on rare occasion with a "no stubbed request for".. failure despite the mocked stubs being accurate for the test.

From code inspection, there appears to be a delete vs. iteration race that is possible when multiple threads are running against the same Test::Stubs, with the delete in match() with the "each" iteration in matches?:
```
         stub, meta = matches?(stack, path, headers, body)
          if stub
            consumed << stack.delete(stub)
            return stub, meta
          end
          matches?(consumed, path, headers, body)
```

```
        def matches?(stack, path, headers, body)
          stack.each do |stub|
            match_result, meta = stub.matches?(path, headers, body)
            return stub, meta if match_result
          end
          nil
        end
```

When the issue is hit it appears to be able to cause a request that should be matched to not have a Stub match.

## Steps to reproduce
Below is a snippet of code that I believe can be used to reproduce this issue. (Apologies for any sloppiness - I don't write in Ruby quite that often.) The snippet is meant to be similar to the unit test that was failing, with the project specific information simplified out.

The sizes of the arrays and the number of threads were increased to make the issue more likely in this example, and the sleep time was added for this purpose as well, but it still might need to be run a dozen times in order for the issue to be reproduced.

```
require 'faraday'

BASE_PATH = "/path/".freeze

def test
  @stubs = Faraday::Adapter::Test::Stubs.new
  @conn = Faraday.new do |builder|
    builder.adapter :test, @stubs
  end

  test_input = {
    'a' => Array(1..100),
    'b' => Array(101..200),
    'c' => Array(201..300),
    'd' => Array(301..400),
    'e' => Array(401..500),
    'f' => Array(501..600),
    'g' => Array(601..700),
    'h' => Array(701..800),
  }

  (1..800).each do |i|
    @stubs.get(BASE_PATH + i.to_s) {|env| [200, {}, {}]}
  end

  threads = []
  test_input.each_pair do |letter, numbers|
    threads << Thread.new do
      sleep(1)
      numbers.each do |n|
        begin
          r = @conn.get(BASE_PATH + n.to_s, {}, nil)
          unless r.status == 200
            raise "Unexpected response code"
          end
        rescue => e
          puts "Failure for #{n}: #{e}"
        end
      end
      puts "Done."
    end
  end

  threads.each {|thread| thread.join}
end
```

iMacTia:
## Description
Introduce mutex `Monitor` in `Faraday::Test::Stubs`
Fixes #1365

## Additional Notes
cc @ioquatix @tchutw

### DIFF
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -160,7 +160,7 @@ module Faraday
           end
           nil
         end
-        
+
         # Executes the given block in a mutex context to avoid multi-thread race conditions
         def stubs_mutex(&block)
           @stubs_mutex ||= Monitor.new
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -72,7 +72,7 @@ module Faraday
 
           stub, meta = matches?(stack, env)
           if stub
-            consumed << stack.delete(stub)
+            stubs_mutex { consumed << stack.delete(stub) }
             return stub, meta
           end
           matches?(consumed, env)
@@ -159,6 +159,12 @@ module Faraday
             return stub, meta if match_result
           end
           nil
+        end
+        
+          # Executes the given block in a mutex context to avoid multi-thread race conditions
+        def stubs_mutex(&block)
+          @stubs_mutex ||= Monitor.new
+          @stubs_mutex.synchronize(&block)
         end
       end
 
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -55,6 +55,7 @@ module Faraday
           @stack = {}
           @consumed = {}
           @strict_mode = strict_mode
+          @stubs_mutex = Monitor.new
           yield(self) if block_given?
         end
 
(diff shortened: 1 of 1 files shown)


lib/faraday/adapter/test.rb
```diff
@@ -160,6 +160,12 @@ def matches?(stack, env)
           end
           nil
         end
+
+        # Executes the given block in a mutex context to avoid multi-thread race conditions
+        def stubs_mutex(&block)
+          @stubs_mutex ||= Monitor.new
```
olleolleolle, line 166: We could create this one in the initializer, to always have it on hand, knowing it would be created.

iMacTia: And actually, avoid race conditions on its creation 😄! Good point 🙌

olleolleolle (approved):
Enjoying the simplicity! 👍

Issue #1073, opened 2019-11-12
Adapter's extra options within assignment default_adapter


aleksb86:
Hi,

It's seems that missing support adapter extra options within assignment default_adapter. I want to assign default_adapter once (in initialization) and use it in a few connections. Because typhoeus adapter need some extra options (`http_version: :httpv2_0`), I should to add it in each connection, that produces some duplications. It would be great - to assign default_adapter and set its options in the same place.

iMacTia:
Hi @aleksb86, thanks for opening this.

That sounds like a reasonable feature, you'd be looking at something like this:

```ruby
Faraday.default_adapter = :typhoeus
Faraday.default_adapter_options = {...}

conn = Faraday.new(...)
# conn will use default adapter with default options
```

Would this work for you or did you have something different in mind?

aleksb86:
Thanks for fast reply. It is looking great, but not working for me:
```
NoMethodError: undefined method `default_adapter_options=' for #<Faraday::Connection:0x00007fcbd79f6b18>
Did you mean?  default_parallel_manager=
/Users/aleks_overdrive/dev/gettransfer/config/config.rb:97:in `included'
/Users/aleks_overdrive/dev/gettransfer/app.rb:6:in `include'
/Users/aleks_overdrive/dev/gettransfer/app.rb:6:in `<class:Application>'
/Users/aleks_overdrive/dev/gettransfer/app.rb:5:in `<module:GetTransfer>'
/Users/aleks_overdrive/dev/gettransfer/app.rb:3:in `<top (required)>'
/Users/aleks_overdrive/dev/gettransfer/environment.rb:49:in `require'
/Users/aleks_overdrive/dev/gettransfer/environment.rb:49:in `block in <top (required)>'
/Users/aleks_overdrive/dev/gettransfer/environment.rb:52:in `<top (required)>'
config.ru:10:in `require'
config.ru:10:in `block in <main>'
config.ru:in `new'
config.ru:in `<main>'
/Users/aleks_overdrive/dev/gettransfer/Rakefile:390:in `block in <top (required)>'
Tasks: TOP => console
(See full trace by running task with --trace)
```
my current Faraday version: 0.12.2. Maybe update need

iMacTia:
Great, I've labelled it as a new feature so we'll pick it for development at some point.
Since this is not a critical part of the library, we're happy to review a contributor PR if anyone wants to work on it

PR #1382, opened 2022-01-14
Add default adapter options

hirasawayuki:
## Description
Add option for the default adapter.

Fixes https://github.com/lostisland/faraday/issues/1073

### DIFF
--- a/lib/faraday/rack_builder.rb
+++ b/lib/faraday/rack_builder.rb
@@ -73,7 +73,7 @@ module Faraday
     def build
       raise_if_locked
       block_given? ? yield(self) : request(:url_encoded)
-      adapter(Faraday.default_adapter, Faraday.default_adapter_options) unless @adapter
+      adapter(Faraday.default_adapter, **Faraday.default_adapter_options) unless @adapter
     end
 
     def [](idx)
--- a/lib/faraday/rack_builder.rb
+++ b/lib/faraday/rack_builder.rb
@@ -73,7 +73,7 @@ module Faraday
     def build
       raise_if_locked
       block_given? ? yield(self) : request(:url_encoded)
-      adapter(Faraday.default_adapter) unless @adapter
+      adapter(Faraday.default_adapter, Faraday.default_adapter_options) unless @adapter
     end
 
     def [](idx)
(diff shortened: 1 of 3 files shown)


hirasawayuki:
oh... Can't pass a hash as keyword argument since Ruby3.0.
https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

lib/faraday/rack_builder.rb
```diff
@@ -73,7 +73,7 @@ def initialize_dup(original)
     def build
       raise_if_locked
       block_given? ? yield(self) : request(:url_encoded)
-      adapter(Faraday.default_adapter) unless @adapter
+      adapter(Faraday.default_adapter, Faraday.default_adapter_options) unless @adapter
```
iMacTia, line 76: Have you tried the double splat operator?

```suggestion
      adapter(Faraday.default_adapter, **Faraday.default_adapter_options) unless @adapter
```

spec/faraday/rack_builder_spec.rb
```diff
@@ -151,6 +151,34 @@ class Banana < Handler
...
+        end
+      end
+    end
+
+    let(:cat) { subject.adapter.build }
+
+    it 'adds a handler to construct adapter with named options' do
+      Faraday.default_adapter = :cat_adapter
+      Faraday.default_adapter_options = { name: 'Chloe' }
+      Faraday::Adapter.register_middleware(cat_adapter: cat_adapter)
```
iMacTia, line 174: It would be better not to register the dummy adapter, as that will remain registered for other tests as well. This should work as well:
```suggestion
      Faraday.default_adapter = cat_adapter
      Faraday.default_adapter_options = { name: 'Chloe' }
```

iMacTia (commented):
HI @hirasawayuki, thanks for opening this PR.
I'm not against this addition, but we obviously need to make the CI pass first 😄.
I've left a couple of comments on how to achieve that. Unfortunately, I can't test them myself now but I hope they'll at least nudge you in the right direction 👍

lib/faraday/rack_builder.rb
hirasawayuki: Thank you! I've just fixed it!（ ec72df9 ）

spec/faraday/rack_builder_spec.rb
hirasawayuki: I fixed it in e82f273 . I've confirmed that all tests have passed 😃 .
Thank you!!

iMacTia (approved):
Looks good 🙌 💯

PR #1385, opened 2022-01-19
Remove Faraday 2.0 alpha warning from README

haines:
## Description

Faraday 2.0 isn't upcoming or alpha any more, so the warning in the README is misleading.


iMacTia (approved):
Totally forgot about that, thanks for pointing it out 🙌 !

PR #1394, opened 2022-02-03
CHANGELOG: add 2.2.0 section

olleolleolle:
## Description

Even though the CHANGELOG is a secondary effort, it was easier to add than not to.


iMacTia (approved):
Thanks for keeping this up to date ❤️

PR #1414, opened 2022-05-03
docs: Correct default default_adapter value

kuahyeow:
Since https://github.com/lostisland/faraday/pull/1366, the default value for `default_adapter` is `:net_http`.

Fixes https://www.rubydoc.info/gems/faraday/Faraday#default_adapter-class_method


olleolleolle (approved):
Thanks, that's very kind!

Let's get this merged.

PR #1415, opened 2022-05-06
Allow application/x-www-form-url_encoded POST requests to use file objects as the request body

catlee:
## Description
Currently if you try to do something like

```ruby
Faraday.post("https://httpbingo.org/post", StringIO.new("hello"), {'content-length': '5'})
```

you will get an `ArgumentError` raised from `url_encoded.rb` since it's trying to pass the `StringIO` object to `Hash`.

This PR skips over bodies that respond to the `:read` method in addition to bodies that respond to the `:to_str` method.

## Todos
- [x] Tests
- [ ] Documentation


iMacTia (approved):
I didn't know this was possible, TIL 😄!
Thanks for the useful addition 👏

Issue #1346, opened 2021-12-14
NoMethod Error Given When URL is not formatted properly


gardnerapp:
## Basic Info

* Faraday Version: 1.8
* Ruby Version: 3.0.2

## Issue description

When an improper URL is passed to an HTTP verb method the following error is raised: 
```NoMethodError (undefined method `include?' for nil:NilClass```

## Steps to reproduce
```ruby
Faraday.get("google.com") # => returns NoMethodError 
Faraday.get("http://google.com") # => this works 
```

The HTTP verb methods should raise an exception upon being given an improper URL as an argument. This exception should specify what URL format the method expects and what invalid URL formatting looks like as well. 
What files should I look at to start fixing this issue, thanks!

gardnerapp:
Sorry this error actually comes from Net::HTTP
net/http.rb:1679:in `addr_port': undefined method `include?' for nil:NilClass

iMacTia:
Thanks for double-checking @gardnerapp. I also tried this locally with Faraday v1.8 and I get a `Faraday::ConnectionFailed` error back instead of `NoMethodError`, so I'm not sure what's going on your side.
I'm not against adding some extra checks and useful messages though, so I'd welcome a PR that addresses that.

HTTP verb methods are dynamically defined in the [`connection`](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb) class and they internally all call the [`run_request`](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L432) method.

JAugusto42:
I was investigating this, in the `lib > faraday > adapter.rb` file, in the `connection` method i added a `raise`, but I think I need some help with the message and the best way to make sure this implementation is correct. I'll send you a PR so you can review.

fchatterji:
Hey @iMacTia, i looked at this issue as it seemed a good place to start contributing. Like you however I can't reproduce it, I just get a connection error. I think you can close it.

iMacTia:
@fchatterji thanks for the reminder! Closing it now on the basis that we were not able to reproduce it. Happy to reopen if someone can provide additional information!

bautrey37:
I get this same error on faraday 2.7. I think the problem is when faraday is configured like this:


```
Faraday.new(
      url: nil,
	...
)
```

This happened to my code because of a configuration bug when the URL changes based off of environment. The error code that is received comes Net::HTTP, but it would make sense for faraday to catch this configuration problem and return a error message.

[...]

PR #1433, opened 2022-07-16
#1346 - NoMethod Error Given When URL is not formatted properly

JAugusto42:
## Description
Adding a raise to handling exception when the URL is incomplete, by adding a message reporting the error.
related issue #1346

## Todos
- [ ] Tests
- [ ] Documentation

## Additional Notes
this is just a proposal to start thinking about the best way to solve this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       BUNDLE_WITHOUT: development:test
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Ruby 2.7
         uses: ruby/setup-ruby@v1
@@ -43,7 +43,7 @@ jobs:
         ruby: [ '2.6', '2.7', '3.0', '3.1', truffleruby-head ]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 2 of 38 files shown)


iMacTia (changes requested):
@JAugusto42 thanks for taking a stab at this.
I had a quick look and unfortunately, the PR requires some changes, I'll break them down into sections to make it easier for you to address them 👍

## Solution

I don't think this check should happen in the `connection` method for 2 main reasons:
1. this method is not actually used by all adapters
2. even when used, the method is called just before the request is sent. A malformed URL could have gone through other middleware before the adapter and caused issues.

My suggestion would be to add this check in `Faraday::Connection` just before a request is sent down the middleware rack app.
For example [this method](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L471) could be a great place, although I didn't have time to have a closer look right now.

## Test Coverage

When I first looked into the issue, I was unable to reproduce it.
Before we actually put a fix in place, we should add a test that shows the issue and then show that the fix does indeed fix the test (TDD 🎉!)
I'd really like to see such a test before the PR is accepted (and ideally, before any fix is coded).

## Commits history

Also, looking at the commits history, there's something wrong.
I suspect this may be because you based this off an old branch you used to work on documentation.
It's likely that the branch was squash-merged upstream into Faraday, causing your branch to go out of sync.
Before we can merge this, you'll need to reset your fork and add the changes on top of the current `main` branch in Faraday. (my suggestion would be to also avoid pushing your commits to your fork's `main` branch and instead use a separate branch, as that will keep your `main` clean and allow you to merge changes from upstream 👍)

JAugusto42:
Okay, thanks for the advices and the review, and what about the exception message? any suggestion? I'll take a look at it and see what I can do.

iMacTia:
About the error message itself, `https` is a great default, but `http` should also be supported, so the error message is a bit misleading.
As for the check, I'm not sure that checking for the presence of a host is going to be enough to ensure the URL is a valid http/https URL 🤔, but that's why I'd like to see some tests since I could be completely wrong on this one 😄

PR #1436, opened 2022-07-27
Support Proc type for stubbed request body

yykamei:
## Description

Previously, the Faraday testing adapter compared the request body to the stubbed body just by calling `#==` if the stubbed body is present. Sometimes, I want to check the equality between request and stubbed body in more advanced ways. For example, there is a case that I want to check only the parts of the body are actually passed to Faraday instance like this:

```ruby
stubs = Faraday::Adapter::Test::Stubs.new do |stub|
  stub.post('/foo', '{"name:"YK","created_at":"ANY STRING IS OK"}') { [200, {}, ''] }
end
connection.post('/foo', JSON.dump(name: 'YK', created_at: Time.now))
stubs.verify_stubbed_calls
```

In this case, it's difficult to make tests always pass with `"created_at"` because the value is dynamic. So, I came up with an idea to pass a proc as a stubbed value and compare bodies, inside the proc:

```ruby
stubs = Faraday::Adapter::Test::Stubs.new do |stub|
  check = -> (request_body) { JSON.parse(request_body).slice('name') == { 'name' => 'YK' } }
  stub.post('/foo', check) { [200, {}, ''] }
end
connection.post('/foo', JSON.dump(name: 'YK', created_at: Time.now))
stubs.verify_stubbed_calls
```

I believe this would be flexible but compatible with the previous behavior.

### DIFF
--- a/docs/adapters/testing.md
+++ b/docs/adapters/testing.md
@@ -64,6 +64,14 @@ initialized. This is useful for testing.
 stubs.get('/uni') { |env| [ 200, {}, 'urchin' ]}
 ```
 
+You can also stub the request body with a string or a proc.
+It would be useful to pass a proc if it's OK only to check the parts of the request body are passed.
+
+```ruby
+stubs.post('/kohada', 'where=sea&temperature=24') { |env| [ 200, {}, 'spotted gizzard shad' ]}
+stubs.post('/anago', -> (request_body) { JSON.parse(request_body).slice('name') == { 'name' => 'Wakamoto' } }) { |env| [200, {}, 'conger eel'] }
+```
+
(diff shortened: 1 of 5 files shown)


iMacTia (approved):
Thank you for the contribution @yykamei!
This looks great, I've left just a couple of comments and Rubocop is complaining, but once those are fixed I'd be happy to merge 👍

yykamei:
Thank you 😄  I applied your suggestions as well as fixed problems RuboCop detected. I will rebase this PR if it's needed.

iMacTia (approved):
Fantastic all-round addition 🙌 
Thanks for updating the docs and examples too 💯

PR #1437, opened 2022-07-29
Add encoders `#sort_params` to documentation

### LINKED ISSUE #1429: Query parameters doesn't respect hash
## Basic Info

* Faraday Version: 2.3.0
* Ruby Version: 3.0

## Issue description

When performing a GET request with query parameters, the parameters seems to be re-ordered in alphabetical order.
Which is an issue as some web site doesn't parse all parameters at once and behavior will differ due to parameter ordering

## Steps to reproduce
~~~
Faraday.get('http://httpbin.org/get', :b => 1, :a => 1)
~~~
Anwser show that the request was: `http://httpbin.org/get?a=1&b=1`
Instead of expected: : `http://httpbin.org/get?b=1&a=1`

iMacTia:
## Description
Add encoders `#sort_params` to documentation
Fixes #1429

### DIFF
--- a/docs/usage/customize.md
+++ b/docs/usage/customize.md
@@ -77,6 +77,20 @@ serializes POST bodies.
 
 The default encoder is `Faraday::NestedParamsEncoder`.
 
+### Order of parameters
+
+By default, paramters are sorted by name while being serialized.
+Since this is really useful to provide better cache management and most servers don't really care about parameters order, this is the default behaviour.
+However you might find yourself dealing with a server that requires parameters to be in a specific order.
+When that happens, you can configure the encoder to skip sorting them.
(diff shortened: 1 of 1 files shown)


docs/usage/customize.md
```diff
@@ -77,6 +77,20 @@ serializes POST bodies.
 
 The default encoder is `Faraday::NestedParamsEncoder`.
 
+### Order of parameters
+
+By default, paramters are sorted by name while being serialized.
+Since this is really useful to provide better cache management and most servers don't really care about parameters order, this is the default behaviour.
```
olleolleolle, line 83:
```suggestion
Since this is really useful to provide better cache management and most servers don't really care about parameter order, this is the default behaviour.
```

olleolleolle (approved):
Super! Much clearer!

Issue #1410, opened 2022-04-04
Disable upgrade warnings by default


iMacTia:
This seems to be an issue again on the jump from v1 to v2. I'm getting a bunch of these:

```
WARNING: `Faraday::Connection#basic_auth` is deprecated; it will be removed in version 2.0.
While initializing your connection, use `#request(:basic_auth, ...)` instead.
See https://lostisland.github.io/faraday/middleware/authentication for more usage info.
WARNING: `Faraday::Connection#basic_auth` is deprecated; it will be removed in version 2.0.
[...]
```

Did you move away from the `FARADAY_DEPRECATE` idea? To be clear, I don't even directly depend on faraday-- it comes through another dependency.

_Originally posted by @kyrofa in https://github.com/lostisland/faraday/issues/1089#issuecomment-1086227010_

hyuraku:
@iMacTia 
We would like to add `Faraday#DeprecatedClass` to `Faraday ver1.xx` like https://github.com/lostisland/faraday/commit/6e746fceefe0d1f75535b40a62d846a6a95bf85f , How about?

iMacTia:
@hyuraku yep that would work!

iMacTia:
Oh right, `Faraday#DeprecatedClass` deprecates an entire class, but I thought you wanted to use [`Faraday::Deprecate#deprecate`](https://github.com/lostisland/faraday/commit/6e746fceefe0d1f75535b40a62d846a6a95bf85f#diff-a256f6821a10eccdab286b51f2b9e6b178492c404e2a33b08940fe7a0fd75205R70) from that PR. That should allow you do deprecate individual methods with a similar configurable behaviour.

I'd rather not extend or mess with public standard classes like Warning as that could have unexpected repercussions on projects and affects things outside of Faraday's control.

iMacTia:
This was fixed in #1438 🎉 
Thank you @hyuraku 🙏

PR #1438, opened 2022-08-04
add `Faraday#Deprecate` to `1.x`

hyuraku:
## Description
add Faraday#Deprecate to hide deprecate method's warning message

Fixes https://github.com/lostisland/faraday/issues/1410

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -54,7 +54,6 @@ module Faraday
         else :skip
         end
       end
-      @skip = :warn
       @skip == :skip
     end
 
@@ -65,7 +64,7 @@ module Faraday
     # Temporarily turn off warnings. Intended for tests only.
     def skip_during
       original = Faraday::Deprecate.skip
-      Faraday::Deprecate.skip, = true
+      Faraday::Deprecate.skip = true
       yield
     ensure
       Faraday::Deprecate.skip = original
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+module Faraday
+  # @param new_klass [Class] new Klass to use
+  #
+  # @return [Class] A modified version of new_klass that warns on
+  #   usage about deprecation.
+  # @see Faraday::Deprecate
+  module DeprecatedClass
+    def self.proxy_class(origclass, ver = '1.0')
+      proxy = Class.new(origclass) do
+        const_set('ORIG_CLASS', origclass)
+
+        class << self
+          extend Faraday::Deprecate
+
+          def ===(other)
+            (superclass == const_get('ORIG_CLASS') && other.is_a?(superclass)) || super
+          end
+        end
+      end
+      proxy.singleton_class.send(:deprecate, :new, "#{origclass}.new", ver)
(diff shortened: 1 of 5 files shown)


lib/faraday/deprecate.rb
```diff
@@ -0,0 +1,110 @@
...
+  #     end
+  module Deprecate
+    def self.skip # :nodoc:
+      @skip ||= begin
+        case ENV['FARADAY_DEPRECATE'].to_s.downcase
+        when '1', 'warn' then :warn
+        else :skip
+        end
+      end
+      @skip = :warn
```
iMacTia, line 57: This line looks like a mistake 🤔

lib/faraday/deprecate.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    def self.skip=(value) # :nodoc:
+      @skip = value ? :skip : :warn
+    end
+
+    # Temporarily turn off warnings. Intended for tests only.
+    def skip_during
+      original = Faraday::Deprecate.skip
+      Faraday::Deprecate.skip, = true
```
iMacTia, line 68: There's an extra `,` here (I noticed this was also in the original commit! But should probably remove it now)

iMacTia (changes requested):
Thank you for bringing this back @hyuraku.
There are some changes that I wouldn't expect in this PR, please see my comment below.
Do you also think you could add some test coverage to check the warnings are correctly output when the skip is disabled?

lib/faraday/deprecate.rb
hyuraku: This is a code for debugging. I remove it soon.

iMacTia:
Thank you for addressing all the comments @hyuraku 🙌!
Are you planning to add unit tests as well as I mentioned above?

> Do you also think you could add some test coverage to check the warnings are correctly output when the skip is disabled?

hyuraku:
@iMacTia 
> check the warnings are correctly output when the skip is disabled?

I will add the test when the skip is disabled? to `spec/faraday/connection_spec.rb`

lib/faraday/error.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faraday/deprecate'
```
iMacTia, line 3: This is not necessary anymore 🙌 !

iMacTia:
Or better, a generic test around the `Faraday::Deprecate#deprecate` method 👍!
If we know that works as expected (e.g. with an anonymous class defined in the specs), we don't need to test it everywhere we use it

iMacTia (commented):
Just realised we can now remove the `require` from `error.rb` as well

hyuraku:
@iMacTia 
I added the spec for `Faraday::Deprecate#deprecate`; could you review this PR?

iMacTia (approved):
Fantastic work @hyuraku and great test coverage 💯👏
The `rubocop` offence is unrelated to the PR itself, I'll test this locally and merge if all tests are passing 👍

iMacTia:
All tests pass locally ✅ 🎉

kyrofa:
Thank you both!

kenichi-sonoda (commented):
Sorry, mistake

PR #1439, opened 2022-08-08
Improve streaming interface to provide response info

### LINKED ISSUE #1426: `on_data` callback should be passed in the response's status code
If you request a URL with redirects enabled, I don't think there's a way to distinguish between chunks from the initial request or redirected requests.

```ruby
require 'webmock'
require 'faraday'
require 'faraday/net_http'
require 'faraday/follow_redirects'

include WebMock::API
WebMock.enable!

stub_request(:get, 'https://www.example.com')
  .to_return(body: 'You are being redirected',
             status: [302, 'Content moved'],
             headers: { 'Location' => 'https://example.com' })

stub_request(:get, 'https://example.com')
  .to_return(body: 'You have arrived')

client = Faraday.new do |f|
  f.use Faraday::FollowRedirects::Middleware
end

client.get('https://www.example.com') do |req|
  req.options.on_data = proc do |chunk, _|
    puts "got: #{chunk}"
  end
end
```

outputs

```
got: You are being redirected
got: You have arrived
```

And there is no way within `on_data` to distinguish between chunks from the initial response, and chunks from the redirected response.

Passing in the response and/or status code, and/or headers, and/or env to the the callback would help handle this case.

iMacTia:
## Description

* Add `bake-test-external` and test `faraday-net_http` as part of the CI
* Introduce the new streaming API with `stream_response` helper

Fixes #1426 

## Todos
- [x] Tests
- [x] Documentation

## Additional Notes

* As part of this PR, we're now running `faraday-net_http` tests as part of CI thanks to [bake-test-external](https://github.com/ioquatix/bake-test-external) 🙌 
* However, it seems like `bake-test-external` is incompatible with Ruby 2.6, so we only run them for Ruby 2.7+
* Rubocop-packaging is currently incompatible with the latest Rubocop release. The fix is on the main branch already, but they haven't released a new version yet, so I've temporarily updated the Gemfile to point to that
* `truffleruby-head` and `ruby-head` are back in the CI matrix, but they've been marked as experimental

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,13 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
 
-      - name: Test
+      - name: RSpec
         continue-on-error: ${{ matrix.experimental }}
         run: bundle exec rake
+
+      - name: Test External Adapters
+        if: ${{ matrix.ruby != '2.6' }}
+        continue-on-error: ${{ matrix.experimental }}
+        run: bundle exec bake test:external
+
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '2.6', '2.7', '3.0', '3.1', truffleruby-head ]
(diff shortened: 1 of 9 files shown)


olleolleolle (approved):
Thanks @iMacTia!

The TODOs are good, they work as a sign-post.

iMacTia:
I'd forget about documentation all the time without it 😂

PR #1446, opened 2022-08-20
deprecate three methods for version 2.0

hyuraku:
## Description
apply `Faraday#deprecate` to three deprecated methods for version 2.0.

- `Request#method`
- `Connection#token_auth`
- `Connection#authorization`

related with https://github.com/lostisland/faraday/pull/1438

## Todos
- [x] Tests
- [ ] Documentation

## Additional Notes
Optional section

### DIFF
--- a/lib/faraday/deprecate.rb
+++ b/lib/faraday/deprecate.rb
@@ -78,7 +78,7 @@ module Faraday
     # @param repl [#to_s, :none] the replacement to use, when `:none` it will
     #   alert the user that no replacement is present.
     # @param ver [String] the semver the method will be removed.
-    def deprecate(name, repl, ver)
+    def deprecate(name, repl, ver, custom_message = nil)
       class_eval do
         gem_ver = Gem::Version.new(ver)
         old = "_deprecated_#{name}"
@@ -95,7 +95,8 @@ module Faraday
           msg = [
             "NOTE: #{target_message} is deprecated",
             repl == :none ? ' with no replacement' : "; use #{repl} instead. ",
-            "It will be removed in or after version #{gem_ver}",
+            "It will be removed in or after version #{gem_ver} ",
+            custom_message,
             "\n#{target}#{name} called from #{Gem.location_of_caller.join(':')}"
           ]
           warn "#{msg.join}." unless Faraday::Deprecate.skip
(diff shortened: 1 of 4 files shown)


iMacTia:
Thank you @hyuraku, one thing I'm not entirely happy about is that the existing warnings not only communicate which method has been deprecated and from where it was called, but they also provide a hint as to how to address them, and a link to the relevant documentation.
Simply replacing them with `Faraday::Deprecate` will make that information disappear 😢

Do you think it would be possible to extend `#deprecate` so that it accepts an additional (optional) parameter that we can use to provide a "custom message" on some deprecation warnings, and use that new parameter in these cases?

hyuraku:
@iMacTia  
> Do you think it would be possible to extend `#deprecate` so that it accepts an additional (optional) parameter that we can use to provide a "custom message" on some deprecation warnings, and use that new parameter in these cases?

I  guess it's better to add `custom_message` as the args to `Faraday::Deprecate#deprecate`, How about?

- Before change
```ruby
    def deprecate(name, repl, ver)
         --< omitted >--
          msg = [
            "NOTE: #{target_message} is deprecated",
            repl == :none ? ' with no replacement' : "; use #{repl} instead. ",
            "It will be removed in or after version #{gem_ver}",
            "\n#{target}#{name} called from #{Gem.location_of_caller.join(':')}"
          ]
```


- After change
```ruby
def deprecate(name, repl, ver, custom_message)
       --< omitted >--
      msg = [
            "NOTE: #{target_message} is deprecated",
            repl == :none ? ' with no replacement' : "; use #{repl} instead. ",
            "It will be removed in or after version #{gem_ver}",
            #{custom_message}
            "\n#{target}#{name} called from #{Gem.location_of_caller.join(':')}"
          ]
          
end
```

iMacTia:
Yeah, that would be perfect to share the link to the documentation or any other free-text valuable context on the deprecation 🙏!

hyuraku:
@iMacTia 
Could you review this PR again?

lib/faraday/deprecate.rb
```diff
@@ -54,6 +54,7 @@ def self.skip # :nodoc:
         else :skip
         end
       end
+      @skip = 1
```
olleolleolle, line 57: Just noting this line here, it's for debugging while writing this PR, I take it. Should it be removed?

hyuraku: Yes, the line is for only debugging. I'll remove this soon.

iMacTia (approved):
Thank you @hyuraku for the changes and @olleolleolle for spotting the leftover line. LGTM!

PR #1353, opened 2021-12-29
Improve documentation for v2

iMacTia:
## Description

* Revamps the "adapters" page to prepare for the v2.0 release
* Highlight the recommended way to use Faraday in the "usage" page
* Explain that adapters now need to be explicitly added to the project
* Move documentation on how to build an adapter to a dedicated page
* Move adapters documentation to gems READMEs

PR #1447, opened 2022-08-25
docs: fix usage page to match latest version's default adapter specification

mi-wada:
## Description

Since https://github.com/lostisland/faraday/pull/1366, the default value for default_adapter is :net_http.

## Additional Notes

The added Adapters section was referenced [here](https://github.com/lostisland/faraday/pull/1353/files#diff-15c410571e14efe1076710d52f0e43cef4068175812ffde74ca941b62e494a44L131).

### DIFF
--- a/docs/adapters/index.md
+++ b/docs/adapters/index.md
@@ -10,6 +10,28 @@ a Faraday response object. Adapters are typically implemented with common Ruby
 HTTP clients, but can have custom implementations. Adapters can be configured
 either globally or per Faraday Connection through the configuration block.
 
+For example, consider using `httpclient` as an adapter. Note that [faraday-httpclient](https://github.com/lostisland/faraday-httpclient) must be installed beforehand.
+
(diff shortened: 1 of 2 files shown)


olleolleolle:
I like this, thank you!

As a follow-up, when looking at the Adapters page, I think as a reader, I would like a snippet of code in context, to take the mystery out of it. Setting a different default_adapter and so on.

https://lostisland.github.io/faraday/adapters/

@iMacTia Do you agree that something more concrete on that page could guide a half-knowledgeable reader to the water?

mi-wada:
@olleolleolle 
> As a follow-up, when looking at the Adapters page, I think as a reader, I would like a snippet of code in context, to take the mystery out of it. Setting a different default_adapter and so on.
> 
> https://lostisland.github.io/faraday/adapters/

Sure looks better that way!

07f78ac

I have added a doc about this and would appreciate a review.

PR #1451, opened 2022-09-26
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,7 +4,8 @@ on:
   release:
     types: [published]
 
-permissions: {}
+permissions:
+  contents: read  # to checkout the code (actions/checkout)
 jobs:
   build:
     name: Publish to Rubygems
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,7 @@ on:
   release:
     types: [published]
 
+permissions: {}
 jobs:
   build:
     name: Publish to Rubygems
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -4,6 +4,7 @@ on:
   release:
     types: [published]
 
+permissions: {}
```
iMacTia, line 7: If I understand correctly, this should provide the `contents: read` permissions as well since we use the `actions/checkout@v3` action.

iMacTia (changes requested):
Thank you @sashashura, I'm always in when it comes to improving security practices.
I left a comment about the `publish` action since I think we're missing a permission there, but would welcome your feedback

iMacTia:
Nope, sorry!
I rebased, made the workflow re-run and then forgot about it 🙄
Thanks for the nudge!

iMacTia (approved):

PR #1459, opened 2022-11-10
Convert `self[key]` to a String with `#<<` on `#add_parsed`

yykamei:
## Description

I want to handle [this problem](https://github.com/lostisland/faraday/pull/1456#issuecomment-1309567153).

As I said in #1456, an instance of `Faraday::Utils::Headers` may have non-string values. So, an error might happen when the instance calls `#parse` with such existing values. For example, this code will raise an error before this PR is merged.

```
irb(main):001:0> h = Faraday::Utils::Headers.new
=> {}
irb(main):002:0> h[:foo] = 123
=> 123
irb(main):003:0> h.parse 'foo: 89'
/Users/yutaka.kamei/git/Fork/faraday/lib/faraday/utils/headers.rb:135:in `<<': no implicit conversion of String into Integer (TypeError)
	from /Users/yutaka.kamei/git/Fork/faraday/lib/faraday/utils/headers.rb:135:in `add_parsed'
	from /Users/yutaka.kamei/git/Fork/faraday/lib/faraday/utils/headers.rb:124:in `block in parse'
	from /Users/yutaka.kamei/git/Fork/faraday/lib/faraday/utils/headers.rb:124:in `each'
	from /Users/yutaka.kamei/git/Fork/faraday/lib/faraday/utils/headers.rb:124:in `parse'
	from (irb):3:in `<main>'
	from ./bin/console:15:in `<main>'
```

## Todos

- [x] Tests
- ~[ ] Documentation~

### DIFF
--- a/lib/faraday/utils/headers.rb
+++ b/lib/faraday/utils/headers.rb
@@ -132,7 +132,12 @@ module Faraday
 
       # Join multiple values with a comma.
       def add_parsed(key, value)
-        self[key] ? self[key] << ', ' << value : self[key] = value
+        if key?(key)
+          self[key] = self[key].to_s
+          self[key] << ', ' << value
+        else
+          self[key] = value
+        end
       end
     end
   end
--- a/spec/faraday/utils/headers_spec.rb
+++ b/spec/faraday/utils/headers_spec.rb
@@ -57,10 +57,10 @@ RSpec.describe Faraday::Utils::Headers do
   end
 
   describe '#parse' do
-    before { subject.parse(headers) }
-
     context 'when response headers leave http status line out' do
       let(:headers) { "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\n\r\n" }
+
+      before { subject.parse(headers) }
 
       it { expect(subject.keys).to eq(%w[Content-Type]) }
       it { expect(subject['Content-Type']).to eq('text/html') }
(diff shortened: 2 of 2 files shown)


olleolleolle (approved):
This allows the conversion to happen just before it is needed. Good stuff!

iMacTia (commented):
@yykamei I'm trying to understand how this relates to #1456 🤔

Just to be sure we're on the same page: `Faraday::Utils::Headers#parse` is used to parse *response* headers, while the issues you are experiencing in #1456 has to do with *request* headers.

When Faraday is parsing response headers, it should be doing so against a new instance of `Faraday::Utils::Headers`, so I personally don't see how this error could possibly happen under normal circumstances.

yykamei:
#1456 might be irrelevant. I just found a TypeError when dealing with #1456.

Yes, you're right. This case doesn't happen in usual use cases. However, `Headers#[]=` could be called anywhere (e.g., the setup code for testing), so I don't think it's always safe to call `#parse` without type conversion.

It's OK to close this PR if we don't have to consider the case I found 😇

iMacTia (approved):
I don't really have anything against this change: `parse` is a public API of `headers` and although it's not supposed to be used on a pre-initialised instance, there's nothing really blocking people from doing it 😄.

I just wanted to be sure we agree on the connection between this and #1456 👍

yykamei:
Thank you!

PR #1460, opened 2022-11-10
Feature Request: Allows logging of errors that raised in adapter or other middleware

epaew:
## Description
This patch introduces logging of errors that raised in adapter or other (deeper) middlewares.

I am very much looking forward to merged this patch. Thanks.
## Todos
- [x] Tests
- [x] Documentation

## Additional Notes
Optional section

### DIFF
--- a/docs/middleware/response/logger.md
+++ b/docs/middleware/response/logger.md
@@ -42,12 +42,21 @@ end
 ### Include and exclude headers/bodies
 
 By default, the `logger` middleware logs only headers for security reasons, however, you can configure it
-to log bodies as well, or disable headers logging if you need to. To do so, simply provide a configuration hash
-when you add the middleware to the stack:
+to log bodies and errors as well, or disable headers logging if you need to.
+To do so, simply provide a configuration hash when you add the middleware to the stack:
 
 ```ruby
 conn = Faraday.new(url: 'http://httpbingo.org') do |faraday|
-  faraday.response :logger, nil, { headers: true, bodies: true }
+  faraday.response :logger, nil, { headers: true, bodies: true, errors: true }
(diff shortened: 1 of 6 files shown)


iMacTia (changes requested):
@epaew this looks really good 👏 !
I particularly like the idea of having an `on_error` hook available to all middleware as the base for this implementation.
The code looks great and is well test-covered.

The only thing I'd ask is to add one more piece to the docs ([in this section](https://github.com/lostisland/faraday/pull/1460/files#diff-15d532e8f1e05dec44aa4498ffe3094821c9b0f7f32d06161b0608e327175780R42)) to illustrate how the configuration block now accepts also a `error` option 👍

iMacTia:
Ah sorry, the link is not working as expected (probably because the line is outside of the changed code).
I'm referring to this section:

```
### Include and exclude headers/bodies
By default, the `logger` middleware logs only headers for security reasons, however, you can configure it
to log bodies as well, or disable headers logging if you need to. To do so, simply provide a configuration hash
when you add the middleware to the stack:

conn = Faraday.new(url: 'http://httpbingo.org') do |faraday|
  faraday.response :logger, nil, { headers: true, bodies: true }
end

Please note this only works with the default formatter.
```

which should be updated to mention that errors will also be logged if you pass `true`.
You also make a good point that we should probably illustrate the following syntax as well:

```ruby
faraday.response :logger, nil, { headers: { error: true }, bodies: { error: false } }
```

but to be fair, we're not doing that for the current available options, so I won't consider this a blocker for this PR

epaew:
I added commits. 👍

iMacTia (approved):

Issue #1462, opened 2022-11-18
undefined method `pretty_inspect` in 2.7.0


SujayPrabhu96:
## Basic Info

* Faraday Version: 2.7.0
* Ruby Version: 3.1.2

## Issue description
With latest version of faraday (2.7.0) logging is breaking with `undefined method pretty_inspect` method

## Steps to reproduce
```
conn = Faraday.new(ENDPOINT) do |client|
   client.response :logger, nil, { headers: true, bodies: true }
end
```

## Solution
- `require 'pp'` is must when using `pretty_inspect` which has been removed in https://github.com/lostisland/faraday/pull/1457
- Reference [Doc]: https://docs.ruby-lang.org/en/master/Kernel.html#method-i-pretty_inspect
- Reference: https://github.com/ruby/ruby/blob/v3_1_2/lib/pp.rb#L612-L617

## CHECKLIST (delete before creating the issue)
* If you're not reporting a bug/issue, you can ignore this whole template.
* Are you using the latest Faraday version? If not, please check the [Releases](https://github.com/lostisland/faraday/releases) page to see if the issue has already been fixed.
* Provide the Faraday and Ruby Version you're using while experiencing the issue.
* Fill the `Issue description` and `Steps to Reproduce` sections.
* Delete this checklist before posting the issue.

PR #1463, opened 2022-11-18
fix: require 'pp' to have access to #pretty_inspect

olleolleolle:
## Description

This fixes the well-described #1462 issue.


iMacTia:
Thank you @olleolleolle 🙏!
I wonder why this wasn't caught by tests 🤔

AlexWayfer:
> I wonder why this wasn't caught by tests 🤔

I'm not surprised if it's required by RSpec or something similar, from test bundle.

Issue #1469, opened 2022-12-14
Difference between `timeout` and `open_timeout`


Yu-Chieh-Henry-Yang:
We have these comment in the codebase: https://github.com/lostisland/faraday/blob/1a5b794b88ba897afca24e914b9b19297bc80473/lib/faraday/options/env.rb#L18-L19

`:timeout` says `open/read timeout Integer in seconds`
`:open_timeout` says `read timeout Integer in seconds`

Question 1: Does this mean `timeout` includes `open_timeout`? For example, if everything times out at the limit and we configured to have 4 seconds `open_timeout` and 6 seconds `timeout`, do we actually wait only 6 seconds (which includes the 4 seconds) or do we actually wait 4+6 = 10 seconds?

Question 2: `:open_timeout` is called "open" timeout, why does it says `read timeout Integer in seconds` instead of something like `timeout Integer in seconds for opening connection`?

Thank you very much.

iMacTia:
Hi @Yu-Chieh-Henry-Yang, apologies if this is confusing.

`timeout` is a setting for the whole request and thus is calculated for both opening the connection and reading the response (open/read). If the request hasn't finished after this threshold (even if some data have been received in the meantime), then it will timeout.

OTOH `open_timeout` is only used to time the opening of the connection (i.e. time to first byte received). If the server replies with some data before this threshold, then no timeout will be raised (unless obviously the requests goes over the `timeout` threshold.

iMacTia:
You're right though the description for `open_timeout` seems incorrect!
It should be `:open_timeout - open timeout Integer in seconds`

iMacTia:
There are 2 more types of timeout, [I've added a comment](https://github.com/lostisland/faraday/pull/1470#issuecomment-1359271143) to explain them all to your PR 👍 
Thank you for addressing this 🙏 !

Yu-Chieh-Henry-Yang:
Thanks for the explanation @iMacTia , I will continue the conversation in my PR. 🙏

PR #1470, opened 2022-12-16
Clarify diff between connection settings `timeout` and `open_timeout` 

Yu-Chieh-Henry-Yang:
## Description

This PR aims to clarify the difference between the config settings `timeout` and `open_timeout`.

I am very confused about the difference between `timeout` and `open_timeout`.

Originally, the comment for these 2 config settings says:

```rb
  #   - `:timeout`  open/read timeout Integer in seconds
  #   - `:open_timeout` - read timeout Integer in seconds
```

- `open_timeout` is described as `read timeout Integer in seconds` but the previous line has the word `open` in it, which suggests that `read` is something other than `open` which is confusing because I expect `open_timeout` to represent `open`.
- `timeout` is described as `open/read` which is slightly ambiguous because I find it hard to tell which one of the following it means:
  - open plus read
  - open or read

Also when I was looking at this, I don't really know about TCP connection phase and reading phase and whether they are exclusive of each so searching online to clarify this is difficult as well. I found it difficult to clarify whether `timeout` actually includes the time in `open_timeout` and I have asked a question in here: https://github.com/lostisland/faraday/issues/1469.

The reason that I think `timeout` includes `open_timeout` is because:

1. I saw `faraday` gem using the `typhoeus` adapter in this file: https://github.com/lostisland/faraday/blob/e7bbb4e7500590b13c47010546e8f75e19c19eb4/lib/faraday/adapter/typhoeus.rb#L12
2. `faraday`'s `timeout` is represented by `timeout_ms` in `typhoeus` and `faraday`'s `open_timeout` is represented by `connecttimeout_ms` in `typhoeus` according to [this file](https://github.com/typhoeus/typhoeus/blob/da478030a750475a63570432f444e3d782a135d0/lib/typhoeus/adapters/faraday.rb#L165-L166)
3. There is [documentation regarding timeout in the `typhoeus` gem README.md](https://github.com/typhoeus/typhoeus/blob/da478030a750475a63570432f444e3d782a135d0/README.md?plain=1#L435-L436). It says

> `timeout` is the time limit for the entire request in seconds.
> `connecttimeout` is the time limit for just the connection phase, again in seconds.

### DIFF
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -15,8 +15,14 @@ module Faraday
   #   @return [Hash] options for configuring the request.
   #   Options for configuring the request.
   #
-  #   - `:timeout`      - total timeout for both the connection-opening phase and the reading phase after it added together (Integer in seconds)
-  #   - `:open_timeout` - timeout for only the connection-opening phase (Integer in seconds)
+  #   - `:timeout`      - generic timeout (Integer in seconds) that represents
+  #   different things depending on which Faraday adapter is used. For instance,
+  #   it may represent the timeout for the entire request including both
+  #   time to establish the connection and time to read data from server or it
+  #   may represent the default timeout value for various timeouts like timeout
+  #   for establishing connection and timeout for reading data from server
+  #   - `:open_timeout` - timeout (Integer in seconds) for establishing the
+  #   connection between the client and the server
   #   - `:on_data`      - Proc for streaming
   #   - `:proxy`        - Hash of proxy options
   #       - `:uri`        - Proxy Server URI
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -15,19 +15,19 @@ module Faraday
   #   @return [Hash] options for configuring the request.
   #   Options for configuring the request.
   #
-  #   - `:timeout`      - generic timeout (Integer in seconds) that represents
-  #   different things depending on which Faraday adapter is used. For instance,
-  #   it may represent the timeout for the entire request including both
-  #   time to establish the connection and time to read data from server or it
-  #   may represent the default timeout value for various timeouts like timeout
-  #   for establishing connection and timeout for reading data from server
-  #   - `:open_timeout` - timeout (Integer in seconds) for establishing the
-  #   connection between the client and the server
-  #   - `:on_data`      - Proc for streaming
-  #   - `:proxy`        - Hash of proxy options
-  #       - `:uri`        - Proxy Server URI
-  #       - `:user`       - Proxy server username
-  #       - `:password`   - Proxy server password
+  #   - `:timeout`       - time limit for the entire request (Integer in
+  #                        seconds)
+  #   - `:open_timeout`  - time limit for just the connection phase (e.g.
+  #                        handshake) (Integer in seconds)
+  #   - `:read_timeout`  - time limit for the first response byte received from
+  #                        the server (Integer in seconds)
+  #   - `:write_timeout` - time limit for the client to send the request to the
+  #                        server (Integer in seconds)
+  #   - `:on_data`       - Proc for streaming
+  #   - `:proxy`         - Hash of proxy options
+  #       - `:uri`         - Proxy Server URI
+  #       - `:user`        - Proxy server username
+  #       - `:password`    - Proxy server password
   #
   # @!attribute request_headers
   #   @return [Hash] HTTP Headers to be sent to the server.
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -15,7 +15,7 @@ module Faraday
   #   @return [Hash] options for configuring the request.
   #   Options for configuring the request.
   #
-  #   - `:timeout`      - timeout for both the connection-opening phase and the reading phase after it (Integer in seconds)
+  #   - `:timeout`      - total timeout for both the connection-opening phase and the reading phase after it added together (Integer in seconds)
   #   - `:open_timeout` - timeout for only the connection-opening phase (Integer in seconds)
   #   - `:on_data`      - Proc for streaming
   #   - `:proxy`        - Hash of proxy options
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -25,7 +25,7 @@ module Faraday
   #                        server (Integer in seconds)
   #   - `:on_data`       - Proc for streaming
   #   - `:proxy`         - Hash of proxy options
-  #       - `:uri`         - Proxy Server URI
+  #       - `:uri`         - Proxy server URI
   #       - `:user`        - Proxy server username
   #       - `:password`    - Proxy server password
   #
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -15,8 +15,8 @@ module Faraday
   #   @return [Hash] options for configuring the request.
   #   Options for configuring the request.
   #
-  #   - `:timeout`  open/read timeout Integer in seconds
-  #   - `:open_timeout` - read timeout Integer in seconds
+  #   - `:timeout`      - timeout for both the connection-opening phase and the reading phase after it (Integer in seconds)
+  #   - `:open_timeout` - timeout for only the connection-opening phase (Integer in seconds)
   #   - `:on_data`      - Proc for streaming
   #   - `:proxy`        - Hash of proxy options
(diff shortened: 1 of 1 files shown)


yykamei (commented):
I think `Env#request` comes from this code, so it seems to be an instance of [`RequestOptions`](https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/options/request_options.rb#L5).

https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/request.rb#L127-L128

In the first place, the description of `Env#request` doesn't tell the all attributes of `RequestOptions`, such as `read_timeout` or `write_timeout`. Ideally, I think [`RequestOptions`](https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/options/request_options.rb#L5) should have its own attribute documents, and it might be good for this `request` to just tell `return [RequestOptions]`.

However, it might require more work to add full explanations to `RequestOptions`. How about adding `read_timeout` and `write_timeout` as well?

Note `open_timeout`, `read_timeout`, and `write_timeout` will be referred by [`#request_timeout`](https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/adapter.rb#L85), and it depends on adapters how they will be used. For example, faraday-net_http sets `open_timeout` [here](https://github.com/lostisland/faraday-net_http/blob/0f9970b095ef7ab847883809ad500f453094c88f/lib/faraday/adapter/net_http.rb#L162-L164), but `#request_timeout` will fall back to `timeout` when `open_timeout` is not set. I could be wrong, but "total timeout for both the connection-opening phase and the reading phase" might not be correct.

Yu-Chieh-Henry-Yang:
Thank you for looking into this and the comment.

> I could be wrong, but "total timeout for both the connection-opening phase and the reading phase" might not be correct.

When I created this PR, I am not certain about what it does as well. If this might not be correct, do you know what would be correct instead? To be specific, let's say we have the following phase when we make a request using `faraday`:

- Phase 1, the connection phase
- After phase 1, phase 2, the reading phase

Does `timeout` only represent the phase 2 and not (phase 1 + phase 2)?

Does `open_timeout` only represent phase 1? (If yes, why does it says `read timeout Integer in seconds` instead of `read timeout Integer in seconds` in [the original code comment](https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/options/env.rb#L19)?)

Please let me know if have a mistake in understanding anywhere.

yykamei:
> Does timeout only represent the phase 2 and not (phase 1 + phase 2)?

`timeout` is just the default value for `open_timeout`, `read_timeout`, and `write_timeout`. If only `timeout` is set, `open_timeout`, `read_timeout`, and `write_timeout` will be set with the value of `timeout`.

This is where `timeout` is used in Faraday:

https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/adapter.rb#L90

In faraday-net_http, `read_timeout`, `write_timeout`, and `open_timeout` are set with `#request_timeout`.

https://github.com/lostisland/faraday-net_http/blob/0f9970b095ef7ab847883809ad500f453094c88f/lib/faraday/adapter/net_http.rb#L152-L170

> Does open_timeout only represent phase 1? (If yes, why does it says read timeout Integer in seconds instead of open timeout Integer in seconds in [the original code comment](https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/options/env.rb#L19)?)

Yes in faraday-net_http, but I'm not sure why the code comment says "`read timeout Integer in seconds`". I think `"open timeout Integer in seconds"` is correct 👍

Yu-Chieh-Henry-Yang:
For `typhoeus` adapter:

1. `faraday`'s `timeout` is represented by `timeout_ms` in `typhoeus` and `faraday`'s `open_timeout` is represented by `connecttimeout_ms` in `typhoeus` according to [this file](https://github.com/typhoeus/typhoeus/blob/da478030a750475a63570432f444e3d782a135d0/lib/typhoeus/adapters/faraday.rb#L165-L166)
2. There is [documentation regarding timeout in the `typhoeus` gem README.md](https://github.com/typhoeus/typhoeus/blob/da478030a750475a63570432f444e3d782a135d0/README.md?plain=1#L435-L436). It says

> `timeout` is the time limit for the entire request in seconds.
> `connecttimeout` is the time limit for just the connection phase, again in seconds.

so I think the understanding around `open_timeout` is pretty consistent whereas the use and meaning of `timeout` varies between different adapters.

I will update the code for my PR to reflect that.

Yu-Chieh-Henry-Yang:
In the future, it's probably better if we can expose `open_timeout`, `read_timeout`, `write_timeout` and `total_timeout` and `default_timeout` as configurable options so that adapters can use the correctly named options.

But for now I have expanded the explanation comment to point out that there is an uncertainty on how the `timeout` option is used.

yykamei (approved):
The explanation looks good 👍

iMacTia:
OK so it looks like we have some outdated documentation in this file.
We actually support 4 types of timeouts 😄 

* `timeout` is the time limit for the entire request
* `open_timeout` is the time limit for just the connection phase (e.g. handshake)
* `read_timeout` is the time limit for the first response byte received from the server
* `write_timout`is the time limit for the client to send the request to the server

The revised documentation should note all 4 timeout types and then make clear that `open_timeout`, `read_timeout` and `write_timeout` are only supported in some adapters.

Yu-Chieh-Henry-Yang:
Thanks for the explanation @iMacTia 

According to the comment in https://github.com/lostisland/faraday/pull/1470#issuecomment-1357172732, it seems like the `timeout` value may be used differently in different adapter like `faraday-net_http`. For `faraday-net_http`, it seems to use `timeout` as the default value of `open_timeout`, `read_timeout` and `write_timeout`.

Would you say that it is using the `timeout` incorrectly? (I'm asking this just trying to understand `timeout` correctly)

faraday's `#request_timeout` method falls back to `timeout` value as we can see in here:

https://github.com/lostisland/faraday/blob/20a5d573f95f5cc1fb39d55629932f1ca90dbf78/lib/faraday/adapter.rb#L90

When both `open_timeout` and `read_timeout` are set via `#request_timeout` method, this cause the total timeout from `open_timeout` and `read_timeout` added together to exceed the value of `timeout`, should we perhaps remove the ability for this `#request_timeout` method to fall back to `timeout`? (Maybe in a future major release, since this is breaking change)

Yu-Chieh-Henry-Yang:
I have updated the comment again according to our discussion, if the formatting or explanation is incorrect, please let me know. I'm happy to change them.

iMacTia:
I haven't touched that code in a while, so a couple of questions come up on the Net::HTTP internals:
1. Does it support a global timeout config? If not, then what we're doing is the most sensible thing and closest to what the user might possibly want.
2. When does the read timeout begin to count? If it counts from the connection opening, then yeah we could potentially double the timeout. But maybe it's counting from the request start?

iMacTia:
If Net::HTTP does indeed support a global timeout, then I agree we should set that instead, but I'd be surprised we're not using that already 🤔

iMacTia (approved):
Changes to the docs are good 🙌!
We can continue discussing adapters implementation and the need to fix them, but this is the expected behaviour

Yu-Chieh-Henry-Yang:
I see, thanks for the response, I don't know much about how the Net::HTTP works but this has answered my question.

I don't have the permission to merge this PR but I'm happy for this PR to be merged

lib/faraday/options/env.rb
```diff
@@ -15,13 +15,19 @@ module Faraday
...
+  #                        seconds)
+  #   - `:open_timeout`  - time limit for just the connection phase (e.g.
+  #                        handshake) (Integer in seconds)
+  #   - `:read_timeout`  - time limit for the first response byte received from
+  #                        the server (Integer in seconds)
+  #   - `:write_timeout` - time limit for the client to send the request to the
+  #                        server (Integer in seconds)
+  #   - `:on_data`       - Proc for streaming
+  #   - `:proxy`         - Hash of proxy options
+  #       - `:uri`         - Proxy Server URI
```
olleolleolle, line 28:
```suggestion
  #       - `:uri`         - Proxy server URI
```

Issue #1474, opened 2022-12-26
URL not included in response hash


aaronstillwell:
## Basic Info

* Faraday Version: 2.6.0 (though issue is present in latest afaik)
* Ruby Version: 3.1.2

## Issue description
The URL isn't available on the response object after being converted to a hash. When using certain middleware (e.g https://github.com/tisba/faraday-follow-redirects) the URL on the response isn't necessarily the same as on the request. This information can be therefore valuable to have to check the final URL used to fetch the response.

## Steps to reproduce

```
res = Faraday.get 'https://httpbin.org/status/200'
puts res.env.url
# => #<URI::HTTPS https://httpbin.org/status/200>

dumped = Marshal::dump(res)
loaded = Marshal::load(dumped)

puts loaded.env.url
# => nil
```

## Remediation

Assuming there are no unintended consequences of adding the url to the `to_hash` object, this should be an easy modification to make. (I am happy to follow up with a PR after discussion here).

https://github.com/lostisland/faraday/blob/0e16de8e5ad3d29810c3388ea95836e0e4689635/lib/faraday/response.rb#L61-L66

iMacTia:
Fixed in #1475 🎉

PR #1475, opened 2022-12-26
Add URL to to_hash in Faraday::Response (#1474)

aaronstillwell:
## Description
This PR adds the `url` to the `to_hash` method of Faraday::Response as discussed in #1474 

## Todos
Review, commentary & discussion

## Additional Notes
The url is provided to the hash as-is, which means it'll still be a URI object.

See https://github.com/aaronstillwell/faraday/blob/29ac4eec651f367d4be57985297db20f10796f9d/lib/faraday/response.rb#L65

In the spec, we create a URI object to test with: https://github.com/aaronstillwell/faraday/blob/29ac4eec651f367d4be57985297db20f10796f9d/spec/faraday/response_spec.rb#L6-L9

We currently do not explicitly test that the deserialized object has a URI object (rather than the URL as a string), but manual inspection suggests this works as I'd expect as an end user (that we'd get the URI object back and not a string).

Is there any expectation that `to_hash` should not contain an actual URI object?

### DIFF
--- a/lib/faraday/response.rb
+++ b/lib/faraday/response.rb
@@ -61,7 +61,8 @@ module Faraday
     def to_hash
       {
         status: env.status, body: env.body,
-        response_headers: env.response_headers
+        response_headers: env.response_headers,
+        url: env.url
       }
     end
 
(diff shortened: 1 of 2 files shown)


iMacTia (approved):
I checked the codebase and `to_hash` is currently only used for serialization.
`marshal_dump` should have no issues supporting a `URI` object, and if this is turned into json the URI will automatically be converted into a string.

Issue #1349, opened 2021-12-19
Faraday tries to modify frozen URI (FrozenError)


sled:
Ruby Version = 3.0.2
Faraday Version = 1.8.0, 2.0.0

If a frozen URI is passed, Faraday tries to modify the URL and a `FrozenError:  can't modify frozen URI::HTTP: #<URI::HTTP http://example.org/api/v1>` is raised.

The affected code sections

v1.8.0 https://github.com/lostisland/faraday/blob/v1.8.0/lib/faraday/connection.rb#L448-L459
v2.0.0 https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L448-L459

Minimal code sample:

```rb
endpoint = URI('http://example.org/api/v1').freeze
connection = Faraday.new(endpoint)

ruby/3.0.2/lib/ruby/3.0.0/uri/generic.rb:780:in `set_path': can't modify frozen URI::HTTP: #<URI::HTTP http://example.org/api/v1> (FrozenError)
  from ruby/3.0.2/lib/ruby/3.0.0/uri/generic.rb:807:in `path='
  from ruby/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.8.0/lib/faraday/connection.rb:455:in `path_prefix='
  from ruby/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.8.0/lib/faraday/connection.rb:430:in `url_prefix='
  from ruby/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.8.0/lib/faraday/connection.rb:84:in `initialize'
  from ruby/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.8.0/lib/faraday.rb:112:in `new'
  from ruby/3.0.2/lib/ruby/gems/3.0.0/gems/faraday-1.8.0/lib/faraday.rb:112:in `new'
```

Is behaviour intended?

iMacTia:
Thanks for reporting this @sled, no I would say this is not intended! It seems like a simple overlooking.
When `url` is either a `URI` or `string`, we set `url_prefix` to it passing through [`Utils.URI`](https://github.com/lostisland/faraday/blob/5b8c8a6f2e2813a4449b5e1ffb7a5b895510b077/lib/faraday/utils.rb#L71). And if the URI is already a `URI`, then we just return the URI itself. The problem doesn't happen when `url` is a string because we build a new `URI` with it in that case.

[...]

PR #1481, opened 2023-01-13
Connection#build_exclusive_url: replace simple syntax

hyuraku:
## Description
apply the advice about `Connection#build_exclusive_url`'s syntax in https://github.com/lostisland/faraday/issues/1349#issuecomment-1002170559

> On a side note, the regex to check for a trailing slash [here](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L474) could be replaced with a simple base.path.end_with?('/') which probably is faster and more readable. Also [here](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L477) there's a direct call to URI.parse instead of Utils.URI(...) circumventing the custom URL parser.

## Todos
- [ ] Tests
- [ ] Documentation

## Additional Notes
Optional section

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -471,10 +471,10 @@ module Faraday
     def build_exclusive_url(url = nil, params = nil, params_encoder = nil)
       url = nil if url.respond_to?(:empty?) && url.empty?
       base = url_prefix.dup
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
Looks good and small!

Issue #1479, opened 2023-01-05
ArgumentError due to missing ruby2_keywords on Ruby 3.2.0


anero:
## Basic Info

* Faraday Version: 1.10.2
* Ruby Version: 3.2.0

## Issue description
When adding a middleware that receives keyword arguments in the constructor, the call from [`DependencyLoader#new`](https://github.com/lostisland/faraday/blob/1.x/lib/faraday/dependency_loader.rb#L21) fails because the method is not defined using `ruby2_keywords`. I noticed that this was first added in https://github.com/lostisland/faraday/pull/1153 but then got removed in https://github.com/lostisland/faraday/pull/1211, mentioning that ruby 2.7 did't display any warnings when removing `ruby2_keywords`.
I think that with the change on Ruby 3.2.0 (related to bugs https://bugs.ruby-lang.org/issues/18625 and https://bugs.ruby-lang.org/issues/16466) this no longer works. This is also demonstrated by running the `spec/faraday/rack_builder_spec.rb` specs when targeting Ruby 3.2.0.

I can submit a small PR with this change targeting the `1.x` branch if that works for you.

## Steps to reproduce
Run `spec/faraday/rack_builder_spec.rb` specs on Ruby 3.2.0.

serggl:
facing the same issue

iMacTia:
Thank you @anero, I was reading through Ruby 3.2 release notes and figured this change might cause some issues.
At the time we decided to remove it because it wasn't causing any issues, but I agree we should now follow Ruby 3.2 requirements.

Please feel free to open a PR and address this. Ideally I'd like to fix the `main` branch first, and we can then backport the fix over to `1.x` as I assume the same issue would affect `main` as well?

It goes without saying we would need a test to demonstrate this as well. I'm surprised none of our tests failed but I guess that's because we never test adding a middleware with keyword arguments

timrogers:
Fixed in #1483.

olleolleolle:
And released! Closing this issue.

PR #1483, opened 2023-01-18
Add support for Ruby 3.2.0 in Faraday v1.x

timrogers:
When adding a middleware that receives keyword arguments in the constructor, the call from `DependencyLoader#new` fails because the method is not defined using `ruby2_keywords`.

This adds the required `ruby2_keywords` declaration to `DependencyLoader#new`, fixing the tests and getting Faraday v1.x working with Ruby 3.2.0.

Fixes https://github.com/lostisland/faraday/issues/1479.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0', '3.1']
+        ruby: ['2.4', '2.5', '2.6', '2.7', '3.0', '3.1', '3.2']
 
     steps:
     - uses: actions/checkout@v1
--- a/lib/faraday/dependency_loader.rb
+++ b/lib/faraday/dependency_loader.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'ruby2_keywords'
+
 module Faraday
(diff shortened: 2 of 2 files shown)


timrogers:
There are some failing tests here - which at least according to my machine - seem to exist before my changes. Any ideas? 🙏

iMacTia:
Apologies about that @timrogers, let me take a look

timrogers:
> Apologies about that @timrogers, let me take a look

Thank you ❤️ Apologies if my comment came across as a bit demanding - on re-reading it, it wasn't as kind as it could have been. I know that maintaining an open-source project is a thankless task!

iMacTia:
Not at all! I'm just surprised as to why this would start failing all of a sudden 🤔
I'm testing it locally simply switching ruby with RVM and it works fine, which makes me think this might be due to some dependency breaking on a recent update

iMacTia:
OK I managed to pinpoint it to `rack 3.0`, trying to understand what changed and if/what we should fix.

timrogers:
@iMacTia Thanks! This is looking good now.

timrogers:
@iMacTia Thanks for the quick review and merge. Looking forward to this getting shipped as the issue was noticed in my project, Restforce.

iMacTia:
Thank you @timrogers for fixing it 🙏 !
This is now released as `v1.10.3`

PR #1485, opened 2023-01-20
Use Uri.parse instead of Utils.URI

### LINKED ISSUE #1484: 2.7.3 breaks compatibility with Addressable as URI parser
## Basic Info

* Faraday Version: 2.7.3
* Ruby Version: 3.0.4

## Issue description
Not sure if it's still officially supported, but in the past, Faraday allowed replacing the default URI parser with a custom one (like `Addressable::URI#parse`). It was working fine before 2.7.3, but due to changes in https://github.com/lostisland/faraday/pull/1481 Addressable is no longer compatible.
The example below works fine with 2.7.2, but with 2.7.3 throws:

```ruby
/usr/local/bundle/gems/faraday-2.7.3/lib/faraday/connection.rb:476:in `build_exclusive_url': undefined method `opaque' for #<Addressable::URI:0x550 URI:http://example.com> (NoMethodError)
	from /usr/local/bundle/gems/faraday-2.7.3/lib/faraday/rack_builder.rb:202:in `build_env'
	from /usr/local/bundle/gems/faraday-2.7.3/lib/faraday/rack_builder.rb:153:in `build_response'
	from /usr/local/bundle/gems/faraday-2.7.3/lib/faraday/connection.rb:444:in `run_request'
	from /usr/local/bundle/gems/faraday-2.7.3/lib/faraday/connection.rb:200:in `get'
	from /usr/local/bundle/gems/faraday-2.7.3/lib/faraday.rb:145:in `method_missing'
	from test2.rb:11:in `<main>'
```

## Steps to reproduce
```ruby 
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'addressable', '2.8.1'
  gem 'faraday', '2.7.3'
end

Faraday::Utils.default_uri_parser = Addressable::URI

puts Faraday.get('http://example.com').status
```

iMacTia:
## Description
Utils.URI can be configured to use a different parser and return an object that "behaves like a URI". However, `opaque` is not available in other objects (e.g. Addressable::URI), so in this instance we need to use URI.

Fixes #1484

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -473,7 +473,7 @@ module Faraday
       if url && !base.path.end_with?('/')
         base.path = "#{base.path}/" # ensure trailing slash
       end
-      url = url.to_s.gsub(':', '%3A') if Utils.URI(url.to_s).opaque
+      url = url.to_s.gsub(':', '%3A') if URI.parse(url.to_s).opaque
       uri = url ? base + url : base
       if params
         uri.query = params.to_query(params_encoder || options.params_encoder)
--- a/spec/faraday/connection_spec.rb
+++ b/spec/faraday/connection_spec.rb
@@ -310,6 +310,21 @@ RSpec.describe Faraday::Connection do
         expect(uri.to_s).to eq('http://service.com/api/service%3Asearch?limit=400')
       end
     end
+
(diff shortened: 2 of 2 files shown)


olleolleolle (approved):
Good catch!

PR #1489, opened 2023-01-31
Subclass Options structs using a block

bdewater:
## Description
As recommended on https://ruby-doc.org/core-2.2.2/Struct.html#method-c-new and https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/StructInheritance

Has the nice side-effect of making the methods defined by `Struct.new` discoverable by [Tapioca](https://github.com/Shopify/tapioca), which is how I came across the issue in the first place.

## Todos
N/A

### DIFF
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -1,71 +1,70 @@
 # frozen_string_literal: true
 
 module Faraday
-  # @!attribute method
-  #   @return [Symbol] HTTP method (`:get`, `:post`)
-  #
-  # @!attribute body
-  #   @return [String] The request body that will eventually be converted to a
-  #   string.
-  #
-  # @!attribute url
-  #   @return [URI] URI instance for the current request.
-  #
-  # @!attribute request
-  #   @return [Hash] options for configuring the request.
-  #   Options for configuring the request.
-  #
-  #   - `:timeout`       - time limit for the entire request (Integer in
-  #                        seconds)
-  #   - `:open_timeout`  - time limit for just the connection phase (e.g.
-  #                        handshake) (Integer in seconds)
-  #   - `:read_timeout`  - time limit for the first response byte received from
-  #                        the server (Integer in seconds)
-  #   - `:write_timeout` - time limit for the client to send the request to the
-  #                        server (Integer in seconds)
-  #   - `:on_data`       - Proc for streaming
-  #   - `:proxy`         - Hash of proxy options
-  #       - `:uri`         - Proxy server URI
-  #       - `:user`        - Proxy server username
-  #       - `:password`    - Proxy server password
-  #
-  # @!attribute request_headers
-  #   @return [Hash] HTTP Headers to be sent to the server.
-  #
-  # @!attribute ssl
-  #   @return [Hash] options for configuring SSL requests
-  #
-  # @!attribute parallel_manager
-  #   @return [Object] sent if the connection is in parallel mode
-  #
-  # @!attribute params
-  #   @return [Hash]
-  #
-  # @!attribute response
-  #   @return [Response]
-  #
-  # @!attribute response_headers
-  #   @return [Hash] HTTP headers from the server
-  #
-  # @!attribute status
-  #   @return [Integer] HTTP response status code
-  #
-  # @!attribute reason_phrase
-  #   @return [String]
-  class Env < Options.new(:method, :request_body, :url, :request,
-                          :request_headers, :ssl, :parallel_manager, :params,
-                          :response, :response_headers, :status,
-                          :reason_phrase, :response_body)
-
-    # rubocop:disable Naming/ConstantName
-    ContentLength = 'Content-Length'
-    StatusesWithoutBody = Set.new [204, 304]
-    SuccessfulStatuses = (200..299).freeze
-    # rubocop:enable Naming/ConstantName
+  # @!parse
+  #   # @!attribute method
+  #   #   @return [Symbol] HTTP method (`:get`, `:post`)
+  #   #
+  #   # @!attribute body
+  #   #   @return [String] The request body that will eventually be converted to a
+  #   #   string.
+  #   #
+  #   # @!attribute url
+  #   #   @return [URI] URI instance for the current request.
+  #   #
+  #   # @!attribute request
+  #   #   @return [Hash] options for configuring the request.
+  #   #   Options for configuring the request.
+  #   #
+  #   #   - `:timeout`       - time limit for the entire request (Integer in
+  #   #                        seconds)
+  #   #   - `:open_timeout`  - time limit for just the connection phase (e.g.
+  #   #                        handshake) (Integer in seconds)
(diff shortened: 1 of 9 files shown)


lib/faraday/options/env.rb
```diff
@@ -203,3 +206,5 @@ def self.member_set
     end
   end
 end
+# rubocop:enable Metrics/BlockLength
+# rubocop:enable Metrics/ModuleLength
```
bdewater, line 210: Lot of linter wrangling here, LMK if using the config file is preferable.

olleolleolle:
I'm curious how well YARD can locate the classes like this.

bdewater:
@olleolleolle better it seems, I ran `bundle exec yardoc` and found one instance I had missed, which went away after changing it to the recommended form:
```
[warn]: in YARD::Handlers::Ruby::ClassHandler: Undocumentable superclass (class was added without superclass)
	in file 'lib/faraday/request/instrumentation.rb':8:

	8: class Options < Faraday::Options.new(:name, :instrumenter)
```

olleolleolle:
> @olleolleolle ~better it seems,~ I ran `bundle exec yardoc` and found one instance I had missed, which went away after changing it to the recommended form:
> 
> ```
> [warn]: in YARD::Handlers::Ruby::ClassHandler: Undocumentable superclass (class was added without superclass)
> 	in file 'lib/faraday/request/instrumentation.rb':8:
> 
> 	8: class Options < Faraday::Options.new(:name, :instrumenter)
> ```

Right, so we can log a new issue about YARD overhaul for methods that should be discoverable - it takes a little bit of hand-wrangling to make "synthetic" or manually written doc blocks for code that can not be found by YARD. Help it on its way.

bdewater:
It looks like YARD [can handle `const = Struct.new { ... }`](https://github.com/kagucho/yard/blob/343b3142df7dce725e709d9aa498d48155d17ab9/lib/yard/handlers/ruby/constant_handler.rb#L33-L38) but not subclasses - makes sense since it works on an AST. 

I guess some paths forward could look like:
- use Struct directly and make Options a module that's included
- avoid the usage of Struct altogether and re-implement what's necessary to achieve the same behaviour
- keep PR as-is but implement [a custom YARD handler](https://github.com/lsegal/yard/blob/main/docs/Handlers.md) to help YARD render as we want

lib/faraday/options/env.rb
```diff
@@ -52,12 +53,13 @@ module Faraday
...
-                          :response, :response_headers, :status,
-                          :reason_phrase, :response_body)
-
+  # rubocop:disable Metrics/BlockLength
+  Env = Options.new(:method, :request_body, :url, :request,
+                    :request_headers, :ssl, :parallel_manager, :params,
+                    :response, :response_headers, :status,
+                    :reason_phrase, :response_body) do
     # rubocop:disable Naming/ConstantName
+    # rubocop:disable Lint/ConstantDefinitionInBlock
```
iMacTia, line 62: This rubocop lint is actually correct, switching from class definition to constant assignment is not a compatible change because constants gets defined in the global scope as opposed to the local one.

See this example:
```ruby
Test = Struct.new(:name) do
  CONSTANT = 1
end

class AnotherTest < Struct.new(:name)
  CONSTANT = '1'
end

CONSTANT #=> 1
Test::CONSTANT #=> uninitialized constant Test::CONSTANT (NameError)
AnotherTest::CONSTANT #=> "1"
```

As suggested by the cop, to maintain the same behaviour we need to use `const_set`:

```ruby
FixedTest = Struct.new(:name) do
  const_set(:CONSTANT, :one)
end

FixedTest::CONSTANT #=> :one
```

lib/faraday/options/env.rb
iMacTia: Yeah I believe we already have `Metrics/BlockLength` and `Metrics/ModuleLength` in `.rubocop_todo.yml` so I'd rather update the values there instead 👍

iMacTia (changes requested):
Thank you @bdewater, as a Sorbet/Tapioca user myself I very much welcome this contribution 🎉 !

Unfortunately, a quick test showed that constants defined inside the structs are not behaving like before, so I've requested some changes.

As an additional note, it's worth mentioning that we've had a long standind desire to move away from structs and use proper classes instead, but this will have to be done in a future Faraday v3 release as it will hardly be backwards compatible 😅 

I'd also like to introduce type signatures for Sorbet and RBS to ship with the gem ❤️

lib/faraday/options/env.rb
```diff
@@ -126,18 +123,18 @@ def body=(value)
 
     # @return [Boolean] true if status is in the set of {SuccessfulStatuses}.
     def success?
-      SuccessfulStatuses.include?(status)
+      Env::SuccessfulStatuses.include?(status)
```
iMacTia, line 126: I'm a bit surprised these `Env::` are necessary, but it's not the end of the world 😄

iMacTia (approved):
Thank you 🙏 !

iMacTia:
I'm following the discussion in the `yard` repo, it would be good to find a quick-and-easy solution (they mentioned `!@parse`?) to avoid breaking documentation before merging this in, if possible

bdewater:
The directive breaks out the classes as before. Methods defined inside the block are not automatically documented anymore, but that was already the case for the accessors defined by Struct due to subclassing.

iMacTia:
This is already better than the inline doc generate before.
I assume we'll have to add the missing methods manually as we previously did with the accessors?
I'm happy to get this merged as-is and log a separate issue to get this fixed 👍

@olleolleolle you're the YARD expert in the team, are you aware of any quick-fix we can take advantage of and we might be missing 😅?

iMacTia:
FYI -- https://github.com/lostisland/faraday/issues/1490

olleolleolle:
Thanks for merging - that was the right call. @bdewater Thanks for keeping Faraday future-compatible!

lib/faraday/options/env.rb
```diff
@@ -1,71 +1,70 @@
...
+  #   # @!attribute reason_phrase
+  #   #   @return [String]
+  #   class Env < Options; end
+  Env = Options.new(:method, :request_body, :url, :request,
+                    :request_headers, :ssl, :parallel_manager, :params,
+                    :response, :response_headers, :status,
+                    :reason_phrase, :response_body) do
+    const_set(:ContentLength, 'Content-Length')
+    const_set(:StatusesWithoutBody, Set.new([204, 304]))
+    const_set(:SuccessfulStatuses, (200..299).freeze)
```
semaperepelitsa, line 63: Alternative way is to reopen the class. That way, you can define constants as usual.

```ruby
Env = Options.new(:method, :request_body, :url, :request,
                  :request_headers, :ssl, :parallel_manager, :params,
                  :response, :response_headers, :status,
                  :reason_phrase, :response_body)

class Env
  ContentLength = 'Content-Length'
  StatusesWithoutBody = Set.new([204, 304])
  SuccessfulStatuses = (200..299).freeze
```

PR #1491, opened 2023-02-07
Remove direct struct subclassing

bdewater:
## Description
This is a follow-up to https://github.com/lostisland/faraday/pull/1489 - I noticed I was still missing `Faraday::Request#body=` from Tapioca's generated definitions. 

~Since it also overrides `#[]` and `#[]=` it was impossible to set Struct members, so refactored this to a PORO that mostly (but [not exactly](https://ruby-doc.org/core-2.6/Struct.html)) looks like the struct it used to be.~ 
Edit: aliased the original member getter/setters so we can override `#[]` and `#[]=`. This retains methods inherited from Struct people might be relying on (like `#==` which is tested in Faraday via Marshal dump/load).

Also did `Faraday::Adapter::Test::Stub` while I was here for consistency.

### DIFF
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -24,12 +24,13 @@ module Faraday
   #   @return [String] body
   # @!attribute options
   #   @return [RequestOptions] options
-  #
-  # rubocop:disable Style/StructInheritance
-  class Request < Struct.new(:http_method, :path, :params, :headers, :body, :options)
-    # rubocop:enable Style/StructInheritance
-
+  Request = Struct.new(:http_method, :path, :params, :headers, :body, :options) do
     extend MiddlewareRegistry
+
+    alias_method :member_get, :[]
+    private :member_get
+    alias_method :member_set, :[]=
+    private :member_set
 
     # @param request_method [String]
     # @yield [request] for block customization, if block given
@@ -48,7 +49,7 @@ module Faraday
       if params
         params.replace hash
       else
-        super
+        member_set(:params, hash)
       end
     end
 
@@ -59,7 +60,7 @@ module Faraday
       if headers
         headers.replace hash
       else
-        super
+        member_set(:headers, hash)
       end
     end
 
--- a/lib/faraday/adapter/test.rb
+++ b/lib/faraday/adapter/test.rb
@@ -184,7 +184,7 @@ module Faraday
       end
 
       # Stub request
-      class Stub < Struct.new(:host, :path, :query, :headers, :body, :strict_mode, :block) # rubocop:disable Style/StructInheritance
+      Stub = Struct.new(:host, :path, :query, :headers, :body, :strict_mode, :block) do
         # @param env [Faraday::Env]
         def matches?(env)
           request_host = env[:url].host


olleolleolle:
@bdewater Thanks for adding the context of the change, that makes a lot of sense.

lib/faraday/request.rb
```diff
@@ -24,13 +24,14 @@ module Faraday
...
-  #
-  # rubocop:disable Style/StructInheritance
-  class Request < Struct.new(:http_method, :path, :params, :headers, :body, :options)
-    # rubocop:enable Style/StructInheritance
-
+  Request = Struct.new(:http_method, :path, :params, :headers, :body, :options) do
     extend MiddlewareRegistry
 
+    alias_method :member_get, :[]
+    private :member_get
```
yykamei, line 31: `member_get` doesn't seem to be used in this file. Is it still necessary?

lib/faraday/request.rb
```diff
@@ -24,13 +24,14 @@ module Faraday
...
+    alias_method :member_set, :[]=
+    private :member_set
```
yykamei, line 33: Maybe, I understood why `member_set` is required. Previously, `Request` inherited an anonymous class generated by `Struct`, but in this PR, it becomes just a `Struct` and cannot use `super`. Is it correct?

As far as I see, `#params=` and `#headers=` don't follow the documentation: " Replace params, preserving the existing hash type." In other words, their implementations are different from their documents. I think they don't have to be explicitly declared by overriding themselves. What about simply removing `#params=` and `#headers=` and delegating the functionalities to `Struct`? `Struct` has "#attribute=" out of the box.

lib/faraday/request.rb
bdewater: This is a leftover from an earlier iteration, it could be removed.

lib/faraday/request.rb
bdewater: You are correct for why I added `member_set` :)

I believe the docs for `#params=` and `#headers=` are correct. Normally you don't build a `Request` directly but use methods on `Faraday::Connection`. `Connection` builds `Utils::ParamsHash` and `Utils::Headers` (both subclasses of Hash) in the initializer and [passes](https://github.com/lostisland/faraday/blob/2477fe2da3477559819ccee3ec8aeedf5301d0f9/lib/faraday/connection.rb#L454-L456) these to `Faraday::Request.create`. If you merge in additional params or headers (probably using a regular Hash) like this:
```ruby
connection.get do |req|
  request.headers = { foo: "bar" }
end
```
this code ensures `headers` stays a `Utils::Headers` instance (and does things like stringifying the key) instead of becoming a regular Hash with symbol keys.

Removing these overridden setters leads to one (not very clear) test failure on this line: https://github.com/lostisland/faraday/blob/2477fe2da3477559819ccee3ec8aeedf5301d0f9/spec/faraday/connection_spec.rb#L744
```
  1) Faraday::Connection request params with url and extra params allows to override all params
     Failure/Error: expect(req.params['b']).to eq('b')
     
       expected: "b"
            got: nil
     
       (compared using ==)
```

olleolleolle: Thank you both for adding even more context to this detail!

iMacTia: Thank you @bdewater, and sorry if this struct is a mess 🙄.
Can't wait to get rid of them in v3.0 🙌!

lib/faraday/request.rb
iMacTia: Good point. I'll merge this as-is and cleanup in my other PR 👍

iMacTia (approved):

lib/faraday/request.rb
yykamei: Thank you! That's a great explanation 👍 The document seems to be good because it says "preserving the existing hash **type**".

So, `member_set ` is aliased as Struct's original `#[]=` before it is overridden.

https://github.com/lostisland/faraday/blob/2477fe2da3477559819ccee3ec8aeedf5301d0f9/lib/faraday/request.rb#L96-L98

LGTM

PR #1497, opened 2023-04-13
Fix incorrect use of logger middleware in docs

mattbrictson:
`f.request :logger` as originally suggested in the docs does not work.

```
:logger is not registered on Faraday::Request (Faraday::Error)
```

The correct usage is `f.response :logger`


olleolleolle:
Thanks, @mattbrictson!

iMacTia (approved):
Good catch! Thank you @mattbrictson

PR #1502, opened 2023-05-18
Always assume the existence of `URI::Generic#find_proxy`

yykamei:
## Description

Previously, the method `#proxy_from_env` checked the existence of `URI::Generic#find_proxy` by using `respond_to?(:find_proxy)`; this method was introduced in Ruby 2.0.0.

Since the latest Faraday requires Ruby 2.6, as specified in `faraday.gemspec`, there is no need for `#proxy_from_env` to handle the case where `URI::Generic#find_proxy` doesn't exist.

## Additional Notes

I recommend reviewing this pull request changes with "Hide whitespace."

Screenshot 2023-05-18 at 12 31 28

### DIFF
--- a/lib/faraday/connection.rb
+++ b/lib/faraday/connection.rb
@@ -514,22 +514,17 @@ module Faraday
       return if Faraday.ignore_env_proxy
 
       uri = nil
-      if URI.parse('').respond_to?(:find_proxy)
-        case url
-        when String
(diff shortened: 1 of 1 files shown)


iMacTia (approved):
Fantastic 🙌 !
Thank you for simplifying the code and for the "hide whitespaces" trick 😄 !

PR #1506, opened 2023-06-05
Fix "method redefined" warnings introduced in Faraday 2.7.5

### LINKED ISSUE #1505: Lots of "warning: method redefined; discarding old" printed during testing
## Basic Info

* Faraday Version: 2.7.5
* Ruby Version: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]

## Issue description
With warnings enabled (eg during testing), there are several lines printed with "warning: method redefined; discarding old" warnings.

## Steps to reproduce
```
$ cat Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "faraday"
$ bundle install
Using bundler 2.4.7
Using faraday-net_http 3.0.2
Using ruby2_keywords 0.0.5
Using faraday 2.7.5
Bundle complete! 1 Gemfile dependency, 4 gems now installed.
Bundled gems are installed into `./vendor/bundle`
$ bundle exec irb -W
irb(main):001:0> require "faraday"
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old user
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old password
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old request
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old ssl
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old builder_class
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/request.rb:48: warning: method redefined; discarding old params=
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/request.rb:59: warning: method redefined; discarding old headers=
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/request/instrumentation.rb:10: warning: method redefined; discarding old name
/Users/sarna/Developer/ruby/test-faraday/vendor/bundle/ruby/3.2.0/gems/faraday-2.7.5/lib/faraday/request/instrumentation.rb:15: warning: method redefined; discarding old instrumenter
=> true
```

mattbrictson:
## Description

Faraday 2.7.5 prints warnings about redefined methods. This was reported in #1505.

```
faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old user
faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old password
faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old request
faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old ssl
faraday-2.7.5/lib/faraday/options.rb:177: warning: method redefined; discarding old builder_class
faraday-2.7.5/lib/faraday/request.rb:48: warning: method redefined; discarding old params=
faraday-2.7.5/lib/faraday/request.rb:59: warning: method redefined; discarding old headers=
faraday-2.7.5/lib/faraday/request/instrumentation.rb:10: warning: method redefined; discarding old name
faraday-2.7.5/lib/faraday/request/instrumentation.rb:15: warning: method redefined; discarding old instrumenter
```

The underlying problem is that these methods are implicitly defined by `Struct`. When Faraday attempts to redefine them to alter the default Struct behavior, the warnings are printed.

Fix by using [`remove_method`](https://ruby-doc.org/3.2.2/Module.html#method-i-remove_method) to remove the default Struct implementation before Faraday's are declared.

I tested that the warnings are no longer printed on Ruby 2.6 and 2.7 (via Docker) and 3.0, 3.1, and 3.2 (on macOS via rbenv).

Fixes #1505

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,6 +2,8 @@
 
 require 'rspec/core/rake_task'
 
-RSpec::Core::RakeTask.new(:spec)
+RSpec::Core::RakeTask.new(:spec) do |task|
+  task.ruby_opts = %w[-W]
+end
 
 task default: :spec
--- a/lib/faraday/options.rb
+++ b/lib/faraday/options.rb
@@ -174,6 +174,7 @@ module Faraday
 
       memoized_attributes[key.to_sym] = block
       class_eval <<-RUBY, __FILE__, __LINE__ + 1
+        remove_method(key) if method_defined?(key, false)
         def #{key}() self[:#{key}]; end
       RUBY
     end
--- a/lib/faraday/request.rb
+++ b/lib/faraday/request.rb
@@ -42,6 +42,7 @@ module Faraday
       end
     end
 
+    remove_method :params=
     # Replace params, preserving the existing hash type.
     #
     # @param hash [Hash] new params
@@ -53,6 +54,7 @@ module Faraday
       end
     end
 
+    remove_method :headers=
     # Replace request headers, preserving the existing hash type.
     #
     # @param hash [Hash] new headers
--- a/lib/faraday/request/instrumentation.rb
+++ b/lib/faraday/request/instrumentation.rb
@@ -6,11 +6,13 @@ module Faraday
     class Instrumentation < Faraday::Middleware
       # Options class used in Request::Instrumentation class.
       Options = Faraday::Options.new(:name, :instrumenter) do
+        remove_method :name
         # @return [String]
         def name
           self[:name] ||= 'request.faraday'
         end
 
+        remove_method :instrumenter
         # @return [Class]
         def instrumenter
           self[:instrumenter] ||= ActiveSupport::Notifications


Rakefile
```diff
@@ -2,6 +2,8 @@
 
 require 'rspec/core/rake_task'
 
-RSpec::Core::RakeTask.new(:spec)
+RSpec::Core::RakeTask.new(:spec) do |task|
+  task.ruby_opts = %w[-W]
```
mattbrictson, line 6: 🗒️ This ensures that Ruby warnings are printed when running the specs.

lib/faraday/options.rb
```diff
@@ -174,6 +174,7 @@ def self.memoized(key, &block)
 
       memoized_attributes[key.to_sym] = block
       class_eval <<-RUBY, __FILE__, __LINE__ + 1
+        remove_method(key) if method_defined?(key, false)
```
mattbrictson, line 177: 🗒️ The `memoized` helper can be used to memoize existing methods, or it can be used to create entirely new methods. So it is necessary to check whether the method exists before attempting to remove it.

The specs have an example of `memoized` being used to synthesize a previously non-existent method `a`: https://github.com/lostisland/faraday/blob/d40f3dbb997bc12d8f89978f68eee3262b791fba/spec/faraday/options/options_spec.rb#L260

lib/faraday/request.rb
```diff
@@ -42,6 +42,7 @@ def self.create(request_method)
       end
     end
 
+    remove_method :params=
```
mattbrictson, line 45: 🗒️ Since `Request` is defined via `Struct.new`, it already has getter and setter methods for its struct attributes, which include `params` and `params=`. The code that follows this statement redefines the `params=` method, which would normally trigger a "method redefined" warning. To avoid that warning, I am explicitly removing the default Struct method first.

I've applied this pattern to the other method definitions that were triggering warnings.

iMacTia: Right, that makes sense now!

Before the switch from inheritance (`Request < Struct(...)`) to assignment (`Request = Struct(...)`), these methods were defined in the superclass, so these definitions were simply overriding the parent implementation.
Now they're technically redefinition, which triggers the Ruby warning.

Great catch!

iMacTia (approved):
Thank you for providing the fix AND the insightful comments 🙏 
I was a bit at a loss as to how this was introduced, but now I can clearly see why 🎉

PR #1510, opened 2023-06-15
Fix implementation of Faraday::Error helpers.

### LINKED ISSUE #1509: Faraday::ParsingError.response_status doesn't return expected value when stubbing response
## Basic Info

* Faraday Version: 2.7.6
* Ruby Version: 3.2.2

## Issue description

I'll start off by saying I'm not sure this is an error in production, but we've encountered it when attempting to mock/test our application's handling of odd responses. In particular, we're currently testing that when a parsing error occurs that the HTTP status code will be what the server returned.

## Steps to reproduce

I slightly modified the [existing test](https://github.com/lostisland/faraday/blob/e0ce4a4d7919981181ef44eec6c966a82b65e06e/spec/faraday/response/json_spec.rb#L78-L83) in `json_spec` to demonstrate the "bug":

```rb
  it 'includes the response on the ParsingError instance' do
    process('{') { |env| env[:response] = Faraday::Response.new(status: 502) }
    raise 'Parsing should have failed.'
  rescue Faraday::ParsingError => e
    expect(e.response_status).to eq(502)
    expect(e.response).to be_a(Faraday::Response)
  end
```

This test will fail:

```
  1) Faraday::Response::Json includes the response on the ParsingError instance
     Failure/Error: @response[:status] if @response
     
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/faraday/error.rb:32:in `response_status'
     # ./spec/faraday/response/json_spec.rb:82:in `rescue in block (2 levels) in <top (required)>'
     # ./spec/faraday/response/json_spec.rb:79:in `block (2 levels) in <top (required)>'
     # /Users/brendanabbott/.gem/ruby/3.2.2/gems/webmock-3.18.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # JSON::ParserError:
     #   unexpected token at '{'
     #   ./lib/faraday/response/json.rb:30:in `parse'
```

It appears that `response_status` will delegate to `response[:status]` and this will attempt to find the status using the headers of the stubbed response. In this test, the headers are `nil`, so it blows up.

I'm curious, should it be `response.status` instead, or should we stubbing our responses as:

```rb
Faraday::Response.new(status: 502, response_headers: { status: 502 })
```

iMacTia:
Fix #1509

## Description
Fix implementation of Faraday::Error helpers.
Fixes #1509 

## Todos
- [x] Tests

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -29,15 +29,21 @@ module Faraday
     end
 
     def response_status
-      @response[:status] if @response
+      return unless @response
+
+      @response.is_a?(Faraday::Response) ? @response.status : @response[:status]
     end
 
     def response_headers
-      @response[:headers] if @response
+      return unless @response
+
+      @response.is_a?(Faraday::Response) ? @response.headers : @response[:headers]
     end
 
     def response_body
-      @response[:body] if @response
+      return unless @response
+
+      @response.is_a?(Faraday::Response) ? @response.body : @response[:body]
     end
 
     protected
--- a/spec/faraday/error_spec.rb
+++ b/spec/faraday/error_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe Faraday::ClientError do
+RSpec.describe Faraday::Error do
   describe '.initialize' do
     subject { described_class.new(exception, response) }
     let(:response) { nil }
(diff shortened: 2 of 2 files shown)


olleolleolle (approved):
A vast improvement, and now the "can be nil" is at least the defined behaviour. 

Good stuff!

brendo (approved):
Thanks! Can confirm that copying this patch into our app fixes the issue 👍

larouxn (approved):
Definitely more robust. Thanks for the fix! 👏

iMacTia:
Thank you for merging this @olleolleolle, I'm going to release this as `2.7.7` 🎉

PR #1513, opened 2023-06-29
Raise Error: Add Faraday::RequestTimeoutError

### LINKED ISSUE #1511: Missing HTTP 408 Request Timeout specific Faraday ClientError?
Hey there. I noticed that there is no specific `Faraday::ClientError` for HTTP 408:

https://github.com/lostisland/faraday/blob/be98e8ef6fd97513ce7d4967b796ef44d8b16bcf/lib/faraday/error.rb#L95-L121 does not contain an error for HTTP 408 Request Timeout. We recently seen this one a couple of times when talking to the Slack API for example.

Is there a reason this is not defined? If not, would you accept a PR to add this?

tisba:
Fixes #1511

This PR adds `Faraday::RequestTimeoutError` and raises it on HTTP 408 responses when the [Raise Error Middleware](https://lostisland.github.io/faraday/middleware/raise-error) is used.

I also took the liberty to hint out how to get `eventmachine` installed with Ruby >= 3.0 to work on `docs/`.

Let me know if this works for you or if you expect something in addition. I basically just copy-pasted the new error looking at the surrounding code 😀

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,4 +18,12 @@ $ bundle exec jekyll serve
 # The site will now be reachable at http://127.0.0.1:4000/faraday/
 ```
 
+On newer Ruby versions (>= 3.0) `eventmachine` needs a little help to find OpenSSL 1.1 to get compiled (see <https://github.com/eventmachine/eventmachine/issues/936>). If you're using homebrew on macOS, you can do this:
+
+```bash
+brew install openssl@1.1
+bundle config build.eventmachine --with-openssl-dir=$(brew --prefix openssl@1.1)
+bundle install
+```
+
 [website]: https://lostisland.github.io/faraday
--- a/docs/middleware/response/raise_error.md
+++ b/docs/middleware/response/raise_error.md
@@ -32,13 +32,14 @@ Specific exceptions are raised based on the HTTP Status code, according to the l
 An HTTP status in the 400-499 range typically represents an error
 by the client. They raise error classes inheriting from `Faraday::ClientError`.
 
-* 400 => `Faraday::BadRequestError`
-* 401 => `Faraday::UnauthorizedError`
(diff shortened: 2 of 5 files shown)


olleolleolle (approved):
Thank you, @tisba!

The MDN URLs look great. The new error class follows convention.

Let's get this merged!

docs/README.md
```diff
@@ -18,4 +18,12 @@ $ bundle exec jekyll serve
...
 ```
 
+On newer Ruby versions (>= 3.0) `eventmachine` needs a little help to find OpenSSL 1.1 to get compiled (see <https://github.com/eventmachine/eventmachine/issues/936>). If you're using homebrew on macOS, you can do this:
+
+```bash
+brew install openssl@1.1
+bundle config build.eventmachine --with-openssl-dir=$(brew --prefix openssl@1.1)
+bundle install
+```
+
```
iMacTia, line 28: Thank you for pointing this out @tisba.
Faraday doesn't bundle an `eventmachine` adapter since v2.0, so people should not have this issue unless they're using https://github.com/lostisland/faraday-em_http or https://github.com/lostisland/faraday-em_synchrony.

I'd suggest moving this suggestion in those gems because `eventmachine` is not a dependency in this gem

iMacTia (commented):
Thank you for the addition @tisba, and love the MDN links as well ❤️ 
I left a small comment on the docs addition for `eventmachine` as I believe those should not be in this repo, but moved to the relevant ones.

docs/README.md
tisba: What @olleolleolle said. This is about getting the docs running locally.

iMacTia: really?! Well that's a surprise 😄!
In that case, thanks for the addition 🙏 !

PR #1514, opened 2023-06-29
Include `env[:headers]` in `Stubs::NotFound`

yykamei:
When using `strict_mode`, I often get an error with `Stubs::NotFound` and can't figure out why such an error happens at a glance. After looking into the problem carefully, I finally find out some headers are missing. This is not a good developer experience.

I think the error message of `Stubs::NotFound` should include headers to let developers know the difference between the expected and actual requests.

That's why this patch appends `env[:headers]` to the error message.


iMacTia (approved):
Great quality-of-life improvement for devs 🙌 !
Thank you @yykamei

PR #1079, opened 2019-11-22
Add log_level option to logger default formatter

amrrbakry:
## Description
Logger default formatter log severity can now be configured through the `log_level` option (related #1078)

## Todos
List any remaining work that needs to be done, i.e:
- [X] Tests
- [X] Documentation

PR #1515, opened 2023-07-05
Fix some logging inefficiencies

semaperepelitsa:
When reviewing Faraday logger I noticed two inefficiencies:

- It validates :log_level option on every log message, when it could only do it once during initialisation.
- It allocates a Proc object on every log message for no reason.

This PR fixes both issues.

### DIFF
--- a/lib/faraday/logging/formatter.rb
+++ b/lib/faraday/logging/formatter.rb
@@ -13,11 +13,11 @@ module Faraday
 
       def initialize(logger:, options:)
         @logger = logger
-        @filter = []
         @options = DEFAULT_OPTIONS.merge(options)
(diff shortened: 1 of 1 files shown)


iMacTia (approved):
Thank you @semaperepelitsa for addressing these performance issues 🙌 !

This LGTM, we just have a minor rubocop offense.
The suggested fix from Rubocop works, but it's not ideal in this particular case, so if you prefer you could try moving `@filter` initialisation after `@options` and that should also solve the offense 👍

semaperepelitsa:
@iMacTia, done!

Related PR: https://github.com/lostisland/faraday/pull/1079

cc @amrrbakry

PR #1522, opened 2023-08-06
Fix included middleware links

edwardloveall:
## Description

Hi 👋🏻 Thanks for maintaining Faraday.

The "Available Middleware" links on the [Middleware Overview](https://lostisland.github.io/faraday/#/middleware/index) page were pointing to `/middleware/___` instead of `/middleware/included/___`. I updated them to point to the correct place.

## Todos
- [ ] Tests (N/A I think)
- [x] Documentation


olleolleolle (approved):
Thanks, @edwardloveall! I tried it and it now works right!

PR #1530, opened 2023-10-20
TooManyRequests error

dpep:
## Description
Faraday enumerates many common HTTP errors....could we please add [429 / TooManyRequests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) to the list?

## Todos
- [x] Tests


olleolleolle (approved):
Nicely done! Thanks, @dpep!

PR #1531, opened 2023-10-31
Delete request with body

SemihCag:
## Description

We have the opportunity to send requests simply with both params and body.

Example
```
@request = Faraday.new(url: url, headers: @headers) do |faraday|
  faraday.request(req_type)
  faraday.response(:json)
end
```

Body
`@request.delete(@path, body = @msg)`

Params
`@request.delete(@path, params = @msg)`

## Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/lib/faraday/methods.rb
+++ b/lib/faraday/methods.rb
@@ -2,5 +2,5 @@
 
 module Faraday
   METHODS_WITH_QUERY = %w[get head delete trace].freeze
-  METHODS_WITH_BODY = %w[post put patch].freeze
+  METHODS_WITH_BODY = %w[post put patch delete].freeze
 end


iMacTia:
Hi @SemihCag, I see what you're trying to do here, but unfortunately the change you're proposing is not going to work.
`delete` cannot be in both `METHODS_WITH_QUERY` and `METHODS_WITH_BODY` at the same time, as the latter would override the method definition of the former. This would also make the change backwards-incompatible, so I'm afraid I can't accept it.

The good news however is that you can already send bodies in your `delete` requests 🎉 !
This is possible thanks to the block syntax:

```ruby
conn.delete('/') do |req|
  req.body = # ...
end
```

Please note that not all adapters support this feature, so you'll need to make sure you're using the right one as shown in this [comparative table](https://github.com/lostisland/awesome-faraday#adapters). At the moment, it seems like only the `patron` adapter does not support it.

If you have any issues sending a body with your `delete` requests, feel free to open an issue 👍

SemihCag:
Thank you for your detailed explanation. 🙂

PR #1536, opened 2023-12-14
Lint

olleolleolle:
## Description

Rubocop had found new things to lint warn about.

This fixes it.

Also, when using rc1 of ruby 3.3, racc wasn't automatically included. So, I added that to the Gemfile.

### DIFF
--- a/lib/faraday/logging/formatter.rb
+++ b/lib/faraday/logging/formatter.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'pp'
-
 module Faraday
   module Logging
     # Serves as an integration point to customize logging
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ group :development, :test do
 end
 
 group :development, :lint do
+  gem 'racc', '~> 1.7' # for RuboCop, on Ruby 3.3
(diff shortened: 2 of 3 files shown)


lib/faraday/logging/formatter.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'pp'
```
segiddins, line 3: this removal is causing `NoMethodError`s because `pretty_inspect` is now undefined on hashes after updating to faraday 2.8.0

indirect: Turns out this entire thing already happened a year ago for 2.7.0 as well, and had to be fixed by adding the require: https://github.com/lostisland/faraday/issues/1462.

segiddins: my bad!

olleolleolle: There I go for listening to tools and NOT adding context-preserving comments. Will add back the require and a comment.

olleolleolle: Now released as a patch version. Thank you so much!

https://github.com/lostisland/faraday/releases/tag/v2.8.1

PR #1539, opened 2023-12-19
Configurable JSON encoders and decoders

ne006:
Moves [lostisland/faraday_middleware/#286](https://github.com/lostisland/faraday_middleware/pull/286)

By default `Faraday::Request::Json` and `Faraday::Response::Json` use `json` to encode/decode JSON data, although there are gems like `oj` which do the job faster than Ruby's built-in `json`.

This implementation allows to pass `encoder` option to `Faraday::Request::Json` and decoder option to pass in parser_options to `Faraday::Response::Json`

```ruby
require 'oj'

Faraday.new do |f|
  f.request :json, encoder: Oj
  f.response :json, parser_options: { decoder: Oj }
end
```

### DIFF
--- a/docs/middleware/included/json.md
+++ b/docs/middleware/included/json.md
@@ -26,7 +26,7 @@ conn.post('/', { a: 1, b: 2 })
 By default, middleware utilizes Ruby's `json` to generate JSON strings.
 
 Other encoders can be used by specifying `encoder` option for the middleware:
-* a module/class which implements `encode`
+* a module/class which implements `dump`
 * a module/class-method pair to be used 
 
 ```ruby
@@ -37,7 +37,7 @@ Faraday.new(...) do |f|
 end
 
 Faraday.new(...) do |f|
-  f.request :json, encoder: [Oj, :encode]
+  f.request :json, encoder: [Oj, :dump]
 end
 ```
 
@@ -65,7 +65,7 @@ conn.get('json').body
 By default, middleware utilizes Ruby's `json` to parse JSON strings.
 
 Other decoders can be used by specifying `decoder` parser option for the middleware:
-* a module/class which implements `decode`
+* a module/class which implements `load`
 * a module/class-method pair to be used 
 
 ```ruby
@@ -76,6 +76,6 @@ Faraday.new(...) do |f|
 end
 
 Faraday.new(...) do |f|
-  f.response :json, parser_options: { decoder: [Oj, :decode] }
+  f.response :json, parser_options: { decoder: [Oj, :load] }
 end
 ```
--- a/docs/middleware/included/json.md
+++ b/docs/middleware/included/json.md
@@ -21,6 +21,26 @@ conn.post('/', { a: 1, b: 2 })
 # Body: {"a":1,"b":2}
 ```
 
+### Using custom JSON encoders
+
+By default, middleware utilizes Ruby's `json` to generate JSON strings.
+
+Other encoders can be used by specifying `encoder` option for the middleware:
+* a module/class which implements `encode`
+* a module/class-method pair to be used 
+
+```ruby
+require 'oj'
+
+Faraday.new(...) do |f|
+  f.request :json, encoder: Oj
+end
+
+Faraday.new(...) do |f|
+  f.request :json, encoder: [Oj, :encode]
+end
+```
+
 ## JSON Responses
 
 The `JSON` response middleware parses response body into a hash of key/value pairs.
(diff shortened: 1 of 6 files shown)


lib/faraday/response/json.rb
```diff
@@ -27,7 +27,22 @@ def process_response(env)
...
+        decoder, method_name =
+          if parser_options[:decoder].is_a?(Array) && parser_options[:decoder].size >= 2
+            parser_options[:decoder].slice(0, 2)
+          elsif parser_options[:decoder].respond_to?(:decode)
+            [parser_options[:decoder], :decode]
+          else
+            [::JSON, :parse]
+          end
+
+        parser_options.delete(:decoder)
```
iMacTia, line 43: I'm worried this might affect performance a bit too much since we're duplicating and mutating hashes (thinking about applications making LOTS of req/sec, this could pile up).

This whole block can be moved into the initialiser (or better, in a method that is called by the initialiser) and cache `decoder` and `method` into instance variables to use here in the parse.

lib/faraday/request/json.rb
```diff
@@ -24,7 +24,13 @@ def on_request(env)
...
 
       def encode(data)
-        ::JSON.generate(data)
+        if options[:encoder].is_a?(Array) && options[:encoder].size >= 2
+          options[:encoder][0].public_send(options[:encoder][1], data)
+        elsif options[:encoder].respond_to?(:encode)
+          options[:encoder].encode(data)
+        else
+          ::JSON.generate data
+        end
```
iMacTia, line 33: This should be good from a performance point of view considering we don't do any instantiation

lib/faraday/request/json.rb
```diff
@@ -24,7 +24,13 @@ def on_request(env)
       private
 
       def encode(data)
-        ::JSON.generate(data)
+        if options[:encoder].is_a?(Array) && options[:encoder].size >= 2
+          options[:encoder][0].public_send(options[:encoder][1], data)
+        elsif options[:encoder].respond_to?(:encode)
+          options[:encoder].encode(data)
+        else
+          ::JSON.generate data
```
iMacTia, line 32: nit: please prefer the syntax with parenthesis for consistency

docs/middleware/included/json.md
```diff
@@ -39,3 +59,23 @@ end
...
+
+```ruby
+require 'oj'
+
+Faraday.new(...) do |f|
+  f.response :json, parser_options: { decoder: Oj }
+end
+
+Faraday.new(...) do |f|
+  f.response :json, parser_options: { decoder: [Oj, :encode] }
```
iMacTia, line 79:
```suggestion
  f.response :json, parser_options: { decoder: [Oj, :decode] }
```

iMacTia (approved):
Thank you for moving this contribution 🙏 !
I've left some minor comments to be addressed, but overall this looks good, well test-covered and document 💯

lib/faraday/response/json.rb
```diff
@@ -27,7 +29,11 @@ def process_response(env)
       end
 
       def parse(body)
-        ::JSON.parse(body, @parser_options || {}) unless body.strip.empty?
+        return if body.strip.empty?
+
+        decoder, method_name = @decoder_options
+
+        decoder.public_send(method_name, body, @parser_options)
```
iMacTia, line 36: I see you removed the `|| {}` from `@parser_options`, but this can be `nil`.
It looks like `JSON.parse` is OK with that, but I tried running `Oj.load('{}', nil)` and I got an error as it expects an hash to be passed. That would make the documentation examples fail

docs/middleware/included/json.md
```diff
@@ -21,6 +21,26 @@ conn.post('/', { a: 1, b: 2 })
...
+
+```ruby
+require 'oj'
+
+Faraday.new(...) do |f|
+  f.request :json, encoder: Oj
+end
+
+Faraday.new(...) do |f|
+  f.request :json, encoder: [Oj, :encode]
```
iMacTia, line 40: I haven't been using `Oj` in a while, so I had a quick look and it seems like the correct encoding method is `dump` (using Oj v3.16).

docs/middleware/included/json.md
```diff
@@ -39,3 +59,23 @@ end
...
+
+```ruby
+require 'oj'
+
+Faraday.new(...) do |f|
+  f.response :json, parser_options: { decoder: Oj }
+end
+
+Faraday.new(...) do |f|
+  f.response :json, parser_options: { decoder: [Oj, :decode] }
```
iMacTia, line 79: Same as above, the correct method is `load`, should we avoid confusion by using the correct one?

iMacTia (changes requested):
Few more minor changes, sorry for being picky 🙏!

lib/faraday/response/json.rb
ne006: Whoops, my mistake, shaky hand :)

iMacTia (approved):
Thank you for patiently addressing all my comments, LGTM 🎉 !

victorhazbun:
Take, a digital 🍺 !

PR #1525, opened 2023-08-15
Add `base64` to runtime dependency

koic:
## Description

This PR adds `base64` to runtime dependency to suppress the following Ruby 3.3's warning:

```console
$ ruby -ve 'require "faraday"'
ruby 3.3.0dev (2023-08-14T15:48:39Z master 52837fcec2) [x86_64-darwin22]
/Users/koic/.rbenv/versions/3.3.0-dev/lib/ruby/gems/3.3.0+0/gems/faraday-2.7.10/lib/faraday/utils.rb:3:
warning: base64 which will be not part of the default gems since Ruby 3.4.0
```

cf: https://github.com/rails/rails/pull/48907

## Todos

List any remaining work that needs to be done, i.e:

- [x] Tests
- [x] Documentation

## Additional Notes

N/A

PR #1541, opened 2023-12-22
Remove runtime dependency on base64

Earlopain:
## Description
#1525 removed a warning in ruby 3.3 because of the base64 dependency. Only a single method in one place is used so there's no need to pull it in. Since it's just a wrapper around the `pack` method it's trivial to replace.

Same approach by other gems:
* https://github.com/rack/rack/pull/2110
* https://github.com/rubocop/rubocop/pull/12313

### DIFF
--- a/lib/faraday/utils.rb
+++ b/lib/faraday/utils.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'base64'
 require 'uri'
 require 'faraday/utils/headers'
 require 'faraday/utils/params_hash'
@@ -54,7 +53,7 @@ module Faraday
     end
 
     def basic_header_from(login, pass)
-      value = Base64.encode64("#{login}:#{pass}")
+      value = ["#{login}:#{pass}"].pack('m') # Base64 encoding
       value.delete!("\n")
       "Basic #{value}"
     end
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -15,7 +15,6 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.6'
 
-  spec.add_dependency 'base64'
   # faraday-net_http is the "default adapter", but being a Faraday dependency it can't
   # control which version of faraday it will be pulled from.
   # To avoid releasing a major version every time there's a new Faraday API, we should


lib/faraday/utils.rb
```diff
@@ -54,7 +53,7 @@ def default_params_encoder
     end
 
     def basic_header_from(login, pass)
-      value = Base64.encode64("#{login}:#{pass}")
+      value = ["#{login}:#{pass}"].pack('m')
```
olleolleolle, line 56: For folks reading this code, to teach this reasonable trick, it'd be nice with a code comment, or, a local private method with a clear name. Either way, a sign-post to readers that don't immediately get the connection between Base64 encoding and pack.

```suggestion
      value = ["#{login}:#{pass}"].pack('m') # Base64 encoding 
```

olleolleolle (approved):
I like this approach. Thank you  @Earlopain for bringing it up, in a PR.

(Seen other gems wrap local methods to break the gem dependency.)

Raises the question: Are there other gems in the Awesome Faraday list that sort of second-level depends on base64 being there for them?

Earlopain:
Good point with the other parts of the ecosystem. I have checked every entry on the list and its looking pretty good. No direct usage, I haven't checked transitive dependencies however. The exception being `faraday-jwt` which I know needs some base64 functionality, see my note on that entry below

-----

Anything relating to faraday inside the lostisland org [no](https://github.com/search?q=org%3Alostisland%20base64&type=code)

## Adapters
* async-http-faraday [no](https://github.com/search?q=repo%3Asocketry%2Fasync-http-faraday%20base64&type=code)
* faraday-typhoeus [no](https://github.com/search?q=repo%3Adleavitt%2Ffaraday-typhoeus%20base64&type=code)
* faraday-http [no](https://github.com/search?q=repo%3Alostisland%2Ffaraday-http%20base64&type=code)
* httpx has implemented its own workaround for the base64 dependency (which I don't think is used in the adapter itself anyways) [base64.rb](https://gitlab.com/os85/httpx/-/blob/master/lib/httpx/base64.rb?ref_type=heads)

## Middleware
* faraday-encoding [no](https://github.com/search?q=repo%3Ama2gedev%2Ffaraday-encoding%20base64&type=code)
* faraday-follow-redirects [no](https://github.com/search?q=repo%3Atisba%2Ffaraday-follow-redirects%20base64&type=code)
* faraday-parse_dates [no](https://github.com/search?q=repo%3AAlexWayfer%2Ffaraday-parse_dates%20base64&type=code)
* faraday-cookie_jar [no](https://github.com/search?q=repo%3Amiyagawa%2Ffaraday-cookie_jar%20base64&type=code)
* faraday-detailed_logger [no](https://github.com/search?q=repo%3Aenvylabs%2Ffaraday-detailed_logger%20base64&type=code)
* faraday-encoding [no](https://github.com/search?q=repo%3Ama2gedev%2Ffaraday-encoding%20base64&type=code)
* faraday-http-cache [no](https://github.com/search?q=repo%3Asourcelevel%2Ffaraday-http-cache%20base64&type=code)
* faraday-gzip [no](https://github.com/search?q=repo%3Abodrovis%2Ffaraday-gzip%20base64&type=code)
* faraday-jwt [not directly](https://github.com/search?q=repo%3Anov%2Ffaraday-jwt%20base64&type=code) but through a dependency `json-jwt`. Not currently an issue since a transitive dependency apart from faraday takes care of it. I have openened https://github.com/nov/json-jwt/issues/114
* faraday-mashify [no](https://github.com/search?q=repo%3Asue445%2Ffaraday-mashify%20base64&type=code)
* faraday-oauth1 [no](https://github.com/search?q=repo%3Agemhome%2Ffaraday-oauth1%20base64&type=code)
* faraday-follow-redirects [no](https://github.com/search?q=repo%3Atisba%2Ffaraday-follow-redirects%20base64&type=code)
* faraday-encode_xml [no](https://github.com/search?q=repo%3AAlexWayfer%2Ffaraday-encode_xml%20base64&type=code)
* faraday-decode_xml [no](https://github.com/search?q=repo%3Asoberstadt%2Ffaraday-decode_xml%20base64&type=code)
* faraday-xml [no](https://github.com/search?q=repo%3Agemhome%2Ffaraday-xml%20base64&type=code)

Speaking of the awesome faraday list, the `json_request`, `json_response`, and `url_encoded` links are currently 404. They are bundled in faraday itself so should be part of the check above.

iMacTia (approved):
Thank you for the fix and the thorough analysis @Earlopain.
This should be enough to confidently remove the dependency, and I'd add that if a downstream gem (middleware or adapter) requires `base64` then it should also list it as a dependency rather than relying on Faraday 👍

PR #1544, opened 2023-12-27
Make Ruby 3.0 the min version

iMacTia:
## Description
Make Ruby 3.0 the min version.
Fixes #1543


.github/workflows/refresh_team_page.yml
```diff
@@ -1,15 +0,0 @@
...
-
-permissions: {}
-jobs:
-  build:
-    name: Refresh Contributors Stats
-    runs-on: ubuntu-latest
-    steps:
-      - name: Call GitHub API
-        run: |
-          curl "https://api.github.com/repos/${{ github.repository }}/stats/contributors"
```
AlexWayfer, line 15: Do we really need this change here?

.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2022-08-08 14:26:32 UTC using RuboCop version 1.33.0.
+# on 2023-12-27 11:12:52 UTC using RuboCop version 1.59.0.
```
AlexWayfer, line 3: Seems not completely relevant, but, I guess, acceptable.

README.md
```diff
@@ -35,7 +35,7 @@ Need more details? See the [Faraday API Documentation][apidoc] to see how it wor
 This library aims to support and is [tested against][actions] the currently officially supported Ruby
 implementations. This means that, even without a major release, we could add or drop support for Ruby versions,
 following their [EOL](https://endoflife.date/ruby).
-Currently that means we support Ruby 2.6+
+Currently that means we support Ruby 3.0+
```
AlexWayfer, line 38: You've mentioned some "outdated docs" below. Do we need to correct it?

spec/spec_helper.rb
```diff
@@ -29,14 +29,15 @@
   minimum_coverage_by_file 26
 end
 
-# Ensure all /lib files are loaded
-# so they will be included in the test coverage report.
-Dir['./lib/**/*.rb'].sort.each { |file| require file }
-
 require 'faraday'
```
AlexWayfer, line 32: Just as a note: this should not work correctly with global installed `faraday`. We should use `require_relative` here. But it's for a separate PR.

AlexWayfer (commented):
Thanks.

spec/spec_helper.rb
iMacTia: I believe this shouldn't be an issue if you run `bundle exec rspec`, but yeah I agree this can be discussed separately 🙂

README.md
iMacTia: This was the outdated part I was referring to in the discussion thread. Ruby 2.6 has been EOL for quite some time, but it was still referred to here anyway 😅

.rubocop_todo.yml
iMacTia: Timestamp maybe not, I agree, but the RuboCop version can be useful 👍

.github/workflows/refresh_team_page.yml
iMacTia: Yeah you're right, I let my hand slip. This should go in another PR.
Will remove now 🙂

.rubocop_todo.yml
AlexWayfer: Yeah, I'm generally talking more about RuboCop itself… the PR is about minimal Ruby version, and there are RuboCop updates. But it's up to you, I'm not against it.

.github/workflows/ci.yml
```diff
@@ -24,10 +24,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Ruby 2.7
+      - name: Set up Ruby 3.0
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.7
+          ruby-version: 3.0
```
olleolleolle, line 30: Quote 3.0 to make it not become "3" in YAML stringification.

.rubocop.yml
```diff
@@ -7,7 +7,7 @@ require:
 AllCops:
   DisplayCopNames: true
   DisplayStyleGuide: true
-  TargetRubyVersion: 2.6
+  TargetRubyVersion: 3.0
```
olleolleolle, line 10: I guess THIS YAML works. Idk though, I'd make it quotes anyway.

.github/workflows/ci.yml
AlexWayfer: I guess, we can use just the latest version for RuboCop. Anyway, it has its own `TargetRubyVersion`.

.rubocop.yml
AlexWayfer: Yes, RuboCop doesn't consider `3` as "the latest available minor version 3.x", it requires a specific one, and `3 == 3.0`, so it's OK here.

.rubocop_todo.yml
iMacTia: Ah right, good point. I've limited changes to those necessary after increasing the `TargetRubyVersion` rubocop config from 2.6 to 3.0. This change was required because Rubocop needs that to be the same as the Gemspec's `required_ruby_version`. After that some new offences popped up

.github/workflows/ci.yml
iMacTia: Ah I always forget about that! Thank you @olleolleolle, and good point @AlexWayfer I'll use 3.3 here 👍

olleolleolle:
This good thing means that endless methods are available!

.rubocop.yml
iMacTia: Just tested this locally and @AlexWayfer is right. If I change this to 3.3 I get 60+ new offences, so it seems like Rubocop does indeed parse this correctly 😄

.github/workflows/ci.yml
AlexWayfer: > and good point @AlexWayfer I'll use 3.3 here 👍

I meant, you can use just `latest` here… I've checked [the syntax of `setup-ruby` action](https://github.com/ruby/setup-ruby#supported-version-syntax) — have not found `latest` word support, but you can use `ruby`, which means "MRI engine latest version".

But, what more effective, I believe, as I see in examples, you can just not specify `ruby-version:` option at all and it'll use the latest one. "Less code is a better code". :)

iMacTia: mmmh let's give it a try. I'm a bit worried it might break unexpectedly at some point by pulling a new version, but as far as it pulls stable ones it should be fine? (at least until 4.0, but I guess we have time before that 😂)

iMacTia: Got `input ruby-version needs to be specified if no .ruby-version or .tool-versions file exists`, trying with just `3` now which might give us what we want

iMacTia: Yep, this will use the latest 3.x version available, just what we wanted 🙂
https://github.com/lostisland/faraday/actions/runs/7347680494/job/20004510675?pr=1544#step:3:17

AlexWayfer: > mmmh let's give it a try. I'm a bit worried it might break unexpectedly at some point by pulling a new version, but as far as it pulls stable ones it should be fine? (at least until 4.0, but I guess we have time before that 😂)

I consider it as a safe warning that something has been changed and we should take some actions (it's a linter job, not test or build).



> Got `input ruby-version needs to be specified if no .ruby-version or .tool-versions file exists`

Got it.



> trying with just `3` now which might give us what we want

I'd still write `ruby`, which means `MRI`, and would be OK with that, even for possible 4.0 version. But, as you wish.

iMacTia: > I consider it as a safe warning that something has been changed and we should take some actions

I'd be OK with that being just a warning, but in practice it will cause PRs to fail build, which is annoying for the contributor since it's unrelated to their changes.
This should work fine until we reach Ruby 4.0 so I'd say it's a good compromise 👍 
Thank you for the whole discussion, TIL quite a lot!

olleolleolle (approved):
Super!

PR #1548, opened 2024-01-14
add bundler config to dependabot

geemus:
## Description
I was working on config for faraday-excon and when updating it's dependabot stuff I saw you all didn't have bundler configured here either, so I thought I would offer the setup in case you want it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


olleolleolle:
In case there's no Gemfile.lock, the dependabot Bundler update, does that really do anything?

olleolleolle:
It won't hurt, and it won't be many PRs anyway. Go ahead, please!

olleolleolle:
And, if we don't enjoy it, we can disable it and put a documenting comment in the YAML file and go on.

olleolleolle (approved):
Let's give it a whirl!

PR #1557, opened 2024-04-01
Make dig method case-insensitive in Faraday::Utils::Headers

### LINKED ISSUE #1556: Case Insensitivity Issue with #dig Method on Response Headers
## Basic Info

* Faraday Version: 2.9.0
* Ruby Version: 3.1.2

## Issue description
I've encountered an issue with the case-insensitive [Hash](https://github.com/lostisland/faraday/blob/main/lib/faraday/utils/headers.rb) which is used as a response headers container. While the case insensitivity works as expected when accessing headers using `#[]`, it fails when using the `#dig` method. Rack handles this properly: https://github.com/rack/rack/blob/8c73aefcc7085c71bdfe6c1ec867f126ede34124/lib/rack/headers.rb#L131-L133

## Steps to reproduce
```ruby
require 'faraday'

response = Faraday.get('http://httpbingo.org')

pp response.headers.dig('via') # "1.1 fly.io"
pp response.headers.dig('Via') # nil
```

vitali-semenyuk:
## Description
Adds `#dig` method to Faraday::Utils::Headers to keep Hash access methods consistent

Fixes #1556 

## Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/lib/faraday/options/env.rb
+++ b/lib/faraday/options/env.rb
@@ -169,7 +169,7 @@ module Faraday
     def stream_response(&block)
       size = 0
       yielded = false
-      block_result = block.call do |chunk| # rubocop:disable Performance/RedundantBlockCall
+      block_result = block.call do |chunk|
         if chunk.bytesize.positive? || size.positive?
           yielded = true
           size += chunk.bytesize


spec/faraday/utils/headers_spec.rb
```diff
@@ -56,6 +56,15 @@
     it { expect(subject.delete('content-type')).to be_nil }
   end
 
+  describe '#dig' do
+    before { subject['Content-Type'] = 'application/json' }
+
+    it { expect(subject&.dig('Content-Type')).to eq('application/json') }
+    it { expect(subject&.dig('CONTENT-TYPE')).to eq('application/json') }
+    it { expect(subject&.dig(:content_type)).to eq('application/json') }
+    it { expect(subject&.dig('invalid')).to be_nil }
```
olleolleolle, line 65: The safe-navigation operator is not needed here.

(The use of the `subject` dynamic definition's superfluous, too, but make no changes to that in this PR.)

olleolleolle (approved):
Thanks!

PR #1563, opened 2024-05-28
Support default json decoder even when nil responds to `:load`

gtmax:
## Description
In Rails (some versions at least), nil responds to `:load` (via ActiveSupport::Dependencies::Loadable mixin). Default json decoder should still work to work in this case

### DIFF
--- a/lib/faraday/response/json.rb
+++ b/lib/faraday/response/json.rb
@@ -60,7 +60,7 @@ module Faraday
         @decoder_options =
           if @decoder_options.is_a?(Array) && @decoder_options.size >= 2
             @decoder_options.slice(0, 2)
-          elsif @decoder_options.respond_to?(:load)
+          elsif @decoder_options&.respond_to?(:load)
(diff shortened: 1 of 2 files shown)


iMacTia (commented):
Wow this is nifty find, thank you for fixing it!

There are a couple Rubocop offences to fix in order to make this pass (I've suggested a change for the first, and the second is a trailing whitespace in the spec file).

I've also requested the addition of a comment to explain what's going on, to make sure future contributors are aware of this issue.

As an alternative, we could also add a guard at the beginning of the method:

```ruby
return [::JSON, :parse] unless @decode_options
```

Adding this guard would mean we don't need to use the safe navigator afterwards, it will make the change more explicit and won't require the `# rubocop:disable` comment, so please consider this alternative (the test shouldn't need to change either!)

St0nersdash (approved):

iMacTia (approved):

PR #1565, opened 2024-06-04
Add `TooManyRequestsError` (429) to error docs

tijmenb:
## Description

`TooManyRequestsError` for HTTP 429 was added in https://github.com/lostisland/faraday/pull/1530. This updates the docs. I've checked the other exceptions as well, this one was the only one missing.


iMacTia (approved):
Thank you for updating the docs 🙌 !

PR #1571, opened 2024-06-25
Configure "npm" package-ecosystem for Dependabot

yykamei:
## Description

I propose adding a new `package-ecosystem` to `dependabot.yml` to maintain the npm dependencies up to date.

After migrating to the new docs, this repository has included `package.json` and its lock file. Maybe, using Dependabot for them can be beneficial.


iMacTia (approved):
Fair point. They're not as important as the Ruby ones, but it would still be nice to keep them up-to-date 👍

PR #1572, opened 2024-06-27
Introduce Middleware DEFAULT_OPTIONS with Application and Instance Configurability

ryan-mcneil:
## Description

This PR introduces a new feature to Faraday::Middleware: Default Options. While the concept of `DEFAULT_OPTIONS` isn't fundamentally new to Faraday, this PR provides some building blocks to introduce Default Options, and the ability to override them at the application level. While it has only been implemented in the RaiseError class, it can easily be added to other Faraday::Middleware classes for which there's a need.

The suggestion was first mentioned [here](https://github.com/lostisland/faraday/pull/1526#issuecomment-2093164772) with some conversation and ideation afterwards. As our team _did_ have a need for this functionality (i.e. setting a different default for `:include_request`), I created a little monkeypatch in our repo, and figured I'd submit a [much more thought out] addition to Faraday itself.

My addition to the documentation (please let me know if you would like this elsewhere!) mostly sums up the general idea and how a user would implement it.

Addresses comment [here](https://github.com/lostisland/faraday/pull/1526#issuecomment-2093348768)

## Additional Notes

The additional "feature" that I did not include in the documentation is that this implementation allows for `DEFAULT_OPTIONS` to be set at the `FARADAY::MIDDLEWARE` level, that would be propagated to child classes. Leaving it blank for now allows for the feature to work as intended, but could open it up for additional configuration in future iterations of the gem. I'd love to hear your thoughts here.

This is my first time contributing to a widely used gem, so be gentle 😆

### DIFF
--- a/docs/middleware/index.md
+++ b/docs/middleware/index.md
@@ -139,7 +139,7 @@ If the default value provides the desired functionality, no further consideratio
 
 #### Setting Alternative Options per Application
 
-In the case where it is desirable to change the default option for all instances within an applcation, it can be done by configuring the options in a `/config/initializers` file. For example:
+In the case where it is desirable to change the default option for all instances within an application, it can be done by configuring the options in a `/config/initializers` file. For example:
 
 ```ruby
 # config/initializers/faraday_config.rb
--- a/docs/middleware/index.md
+++ b/docs/middleware/index.md
@@ -114,6 +114,47 @@ conn = Faraday.new do |f|
 end
 ```
 
+### DEFAULT_OPTIONS
+
+`DEFAULT_OPTIONS` improve the flexibility and customizability of new and existing middleware. Class-level `DEFAULT_OPTIONS` and the ability to set these defaults at the application level compliment existing functionality in which options can be passed into middleware on a per-instance basis.
+
+#### Using DEFAULT_OPTIONS
+
+Using `RaiseError` as an example, you can see that `DEFAULT_OPTIONS` have been defined at the top of the class:
+
+```ruby
+  DEFAULT_OPTIONS = { include_request: true }.freeze
+```
+
+These options will be set at the class level upon instantiation and referenced as needed within the class. Using our same example:
+
+```ruby
+  ...
+  return response unless options[:include_request]
+```
+
+If the default value provides the desired functionality, no further consideration is needed.
+
+#### Setting Alternative Options per Application
+
+In the case where it is desirable to change the default option for all instances within an applcation, it can be done by configuring the options in a `/config/initializers` file. For example:
+
+```ruby
+# config/initializers/faraday_config.rb
+
+Faraday::Response::RaiseError.default_options = { include_request: false }
+```
+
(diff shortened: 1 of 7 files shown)


docs/middleware/index.md
```diff
@@ -114,6 +114,49 @@ conn = Faraday.new do |f|
...
+  ...
+    return response unless options[:include_request]
+  ...
+```
+
+If the default value provides the desired functionality, no further consideration is needed.
+
+#### Setting Alternative Options per Application
+
+In the case where it is desirable to change the default option for all instances within an applcation, it can be done by configuring the options in a `/config/initializers` file. For example:
```
iMacTia, line 142:
```suggestion
In the case where it is desirable to change the default option for all instances within an application, it can be done by configuring the options in a `/config/initializers` file. For example:
```

lib/faraday/middleware.rb
```diff
@@ -7,9 +7,43 @@ class Middleware
...
+      # class level.
+      #
+      # @example Set the Faraday::Response::RaiseError option, `include_request` to `false`
+      # my_app/config/initializers/my_faraday_middleware.rb
+      #
+      # Faraday::Response::RaiseError.default_options = { include_request: false }
+      #
+      def default_options=(options = {})
+        validate_default_options(options)
+        @default_options = default_options.merge(options)
```
iMacTia, line 28: This method reads/writes from a class-level variable, so it's not thread-safe.
Could you please wrap its implementation into a `Monitor` object?

lib/faraday/middleware.rb
```diff
@@ -7,9 +7,43 @@ class Middleware
...
+      end
+
+      # default_options attr_reader that initializes class instance variable
+      # with the values of any Faraday::Middleware defaults, and merges with
+      # subclass defaults
+      def default_options
+        @default_options ||= DEFAULT_OPTIONS.merge(self::DEFAULT_OPTIONS)
```
iMacTia, line 35: This is also not thread safe, but worst case scenario we just re-assign the same value twice, so no biggie 😄

iMacTia (changes requested):
Hi @ryan-mcneil, thank you for the thorough PR (complete with extensive documentation, much appreciated 🙏) and apologies if it took me longer than expected to review.

This looks really great and implements pretty well what I had in mind and described in the comments you linked to.
I requested a couple changes but this looks like a really solid start 💪

lib/faraday/middleware.rb
ryan-mcneil: Thanks for the feedback! ✅

ryan-mcneil:
@iMacTia good catch on the typo 🤦 , and thanks for the feedback!

iMacTia (approved):
Looking great, congrats on your first Faraday contribution 🙌 
This was very well executed 💯

iMacTia:
Release note: I'll be away most of the weekend so I'll cut a release on Monday if that's OK as I would hate to be unavailable in case anything goes wrong

ryan-mcneil:
Shoutout to @ericboehs for the assist!

Thanks @iMacTia! Looking forward to it!

iMacTia:
This has just been released in v2.10.0 🚀

PR #1573, opened 2024-07-05
Add logger as dependency

wynksaiddestroy:
To ensure compatibility with Ruby 3.4 logger needs to be added as an explicit dependency.
A deprecation warning is already shown. See the output of the RSpec step for example: 
https://github.com/lostisland/faraday/actions/runs/9661172692/job/26648356492

### DIFF
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
   # This way, we can release minor versions of the adapter with "breaking" changes for older versions of Faraday
   # and then bump the version requirement on the next compatible version of faraday.
   spec.add_dependency 'faraday-net_http', '>= 2.0', '< 3.2'
+  spec.add_dependency 'logger'
 
   # Includes `examples` and `spec` to allow external adapter gems to run Faraday unit and integration tests
   spec.files = Dir['CHANGELOG.md', '{examples,lib,spec}/**/*', 'LICENSE.md', 'Rakefile', 'README.md']


iMacTia (approved):
Woah I did not expect to see `logger` taken out!
Is Ruby gonna ship without a logger out of the box 😂?

Well, jokes aside, thank you for reporting and proposing the PR 🙂

iMacTia:
Note on release: unless you need this released in a timely manner, I'd wait for #1572 to be merged and then release them together as v2.10

wynksaiddestroy:
> Note on release: unless you need this released in a timely manner, I'd wait for #1572 to be merged and then release them together as v2.10

As long as it released before Ruby 3.4.0, there's no rush. Thanks for the quick merge.

iMacTia:
I *think* we're good for Ruby 3.4, in the sense that we should only get the warning with that version but the code will still work. The job you linked do, in fact, run `ruby 3.4.0dev` and as the warning suggests `logger` will be removed in 3.5.0.

That said, I still want to release this as soon as possible because i) the warning is still annoying 😄 and ii) I don't want to forget about it until Ruby 3.5.0 actually arrives 🙃

iMacTia:
Just released v2.10.0 with this fix 🙌 !
Thanks again for contributing 🙇

wynksaiddestroy:
> Just released v2.10.0 with this fix 🙌 ! Thanks again for contributing 🙇

Thank **you** for the quick merge and the fast release ❤️

PR #1575, opened 2024-07-22
CI: Drop EOL Ruby 3.0 from build matrix

olleolleolle:
## Description

Ruby versions go out of support after a few years.

This change stops using Ruby 3.0 to run our test suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


iMacTia:
Do we actually have any reported issues with 3.0?
It's true that Ruby 3.0 EOL'ed 3 months ago, but I think it might be worth keeping in in the CI pipeline while it's basically effortless to maintain it?

olleolleolle:
No need to drop this, yet.

PR #1580, opened 2024-08-13
Opt-in for MFA requirement explicitly

tagliala:
As a popular gem, `faraday` implicitly requires that all privileged operations by any of the owners require OTP.

However, by explicitly setting `rubygems_mfa_required` metadata, the gem will show "NEW VERSIONS REQUIRE MFA" and
"VERSION PUBLISHED WITH MFA" in the sidebar at
https://rubygems.org/gems/faraday

Ref:
- https://blog.rubygems.org/2022/08/15/requiring-mfa-on-popular-gems.html
- https://guides.rubygems.org/mfa-requirement-opt-in/

---

image

### DIFF
--- a/faraday.gemspec
+++ b/faraday.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
     'changelog_uri' =>
       "https://github.com/lostisland/faraday/releases/tag/v#{spec.version}",
     'source_code_uri' => 'https://github.com/lostisland/faraday',
-    'bug_tracker_uri' => 'https://github.com/lostisland/faraday/issues'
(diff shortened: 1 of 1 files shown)


iMacTia (approved):
Thank you @tagliala, I wasn't aware of this option.
It sounds like we're already MFA-compliant (I vaguely remember having to deal with this at some point), so this change shouldn't affect our release process 👍

tagliala:
Welcome

> It sounds like we're already MFA-compliant

Yes, it should already be this way because of the number of downloads. This is just for presentation (and metadata check for automated tools) purposes

Issue #1581, opened 2024-08-16
Add ciphers to SSL options


womblep:
I have hit a server where I needed to pin the ciphers to be able to connect. The server is most certainly badly configured but it is not something I can get changed.

I was able to work around this by monkey-patching Faraday and the Faraday Net::HTTP adapter (Net::HTTP supports ciphers) but I would like to submit patches to make it easier to do.

The Faraday change is just adding a "ciphers" member to the SSLOptions

Linked to https://github.com/lostisland/faraday-net_http/issues/44

iMacTia:
As a general rule, we only add new `SSLOptions` if these are general HTTP concepts or supported by most of the Faraday adapters / Ruby HTTP clients.

In this case, we know `Net::HTTP` supports it, and I found evidence of [Typhoeus also supporting it](https://github.com/typhoeus/typhoeus/issues/464).

@womblep would you have time to go through the list of [Faraday adapters](https://github.com/lostisland/awesome-faraday?tab=readme-ov-file#adapters) and check how many of the others also support it?

iMacTia:
async-http was a bit tricky, but I think I found support via the `SSLEndpoint` class: https://github.com/search?q=repo%3Asocketry%2Fasync-io%20cipher&type=code

womblep:
**Excon** - supports
# @option params [String] :ciphers Only use the specified SSL/TLS cipher suites; use OpenSSL cipher spec format e.g. 'HIGH:!aNULL:!3DES' or 'AES256-SHA:DES-CBC3-SHA'
* It needs to be added to [OPTS_KEY ](https://github.com/excon/faraday-excon/blob/main/lib/faraday/adapter/excon.rb#L84)

**HTTPClient** - supports
    # A String of OpenSSL's cipher configuration.  Default value is
    # ALL:!ADH:!LOW:!EXP:!MD5:+SSLv2:@STRENGTH
    # See ciphers(1) man in OpenSSL for more detail.
    attr_config :ciphers
* It needs to be added to [configure_ssl(client, ssl)](https://github.com/lostisland/faraday-httpclient/blob/main/lib/faraday/adapter/httpclient.rb#L95) 

**[Net::HTTP::Persistent](https://github.com/lostisland/faraday-net_http_persistent)** - supports
https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent.rb#L571
[...]

PR #1582, opened 2024-08-16
Add ciphers attribute to SSLOptions

womblep:
Fixes issue #1581 
Allows a ciphers attribute to be used to pass a cipher string in OpenSSL format to adapters
This controls which ciphers are chosen to connect to the server.


iMacTia (approved):
As discussed in the linked issue, this is an SSL feature we CAN support in most adapters, so happy to add it to the list of options 👍
Thank you for the research work!

PR #1590, opened 2024-09-10
Make RaiseError middleware configurable to not raise error on certain status codes (e.g. 404)

### LINKED ISSUE #1587: Feature Suggestion: Make RaiseError middleware configurable to not raise error on certain status codes (e.g. 404)
The `RaiseError` would IMO be significantly be more useful if it allows for somewhat flexible configuration of which status codes may not be treated errors. The most prominent example is a 404 status code, which might be entirely expected, such as in cases where you need to implement upsert behavior in an API that doesn't offer `PUT` semantics:

``` ruby
existing_object = client.get('/foo/1').body

if existing_object
  client.patch('/foo', data)
else
  client.post('/foo', data)
end
```

If I generally want to raise errors for 4xx/5xx status codes and thus enable the `RaiseError` middleware, in order to implement such a case, I'd have to clumsily wrap it:

``` ruby
existing_object = begin
  client.get('/foo/1').body
rescue Faraday::ResourceNotFound
end

# ...
```

I could imagine an API like `f.response :raise_error, allowed_statuses: [404]` that consequently just doesn't raise errors for the status codes in that list.

I'm happy to provide a PR with the feature implemented. But before I invest time, I just wanted to know whether you'd be willing to accept such a change and if yes, if you have any strong opinions on what the option should be called (`allowed_statuses`?  `whitelisted_statuses`? `dont_raise_on`? etc.).

clemens:
## Description

This adds the possibility to use the `RaiseError` middleware while preventing certain status from raising errors.

A common example are cases where raising errors on 4xx/5xx status codes is generally desirable and consequently we want to use the `RaiseError` middleware and then having to work around a missing upsert (`PUT`) behavior by first trying to fetch a resource and depending on whether or not the resource is found, we subsequently `POST` to create or `PATCH` to update.

Before:

``` ruby
client = Faraday.new('https://example.com') do |b|
  b.response :raise_error
end

existing_object = begin
  client.get('/foo/1').body
rescue Faraday::ResourceNotFound
  # We expect this resource to potentially be missing, but still have to
  # rescue from it when using the RaiseError middleware
end

if existing_object
  client.patch('/foo', data)
else
  client.post('/foo', data)
end
```

After:

``` ruby
client = Faraday.new('https://example.com') do |b|
  b.response :raise_error, allowed_statuses: [404]
end

existing_object = client.get('/foo/1').body

if existing_object
  client.patch('/foo', data)
else
  client.post('/foo', data)
end
```

Closes #1587.

## Todos

None.

## Additional Notes

In addition to adding the feature, I've refactored the repetitive `case`/`when` statement. I'm happy to take it out if that change is undesirable.

### DIFF
--- a/docs/middleware/included/raising-errors.md
+++ b/docs/middleware/included/raising-errors.md
@@ -63,23 +63,28 @@ and raised as `Faraday::NilStatusError`, which inherits from `Faraday::ServerErr
 
 The behavior of this middleware can be customized with the following options:
 
-| Option              | Default | Description |
-|---------------------|---------|-------------|
-| **include_request** | true    | When true, exceptions are initialized with request information including `method`, `url`, `url_path`, `params`, `headers`, and `body`. |
+| Option               | Default | Description |
+|----------------------|---------|-------------|
+| **include_request**  | true    | When true, exceptions are initialized with request information including `method`, `url`, `url_path`, `params`, `headers`, and `body`. |
+| **allowed_statuses** | []      | An array of status codes that should not raise an error. |
 
 ### Example Usage
 
 ```ruby
 conn = Faraday.new(url: 'http://httpbingo.org') do |faraday|
-  faraday.response :raise_error, include_request: true
+  faraday.response :raise_error, include_request: true, allowed_statuses: [404]
 end
 
 begin
-  conn.get('/wrong-url') # => Assume this raises a 404 response
-rescue Faraday::ResourceNotFound => e
-  e.response[:status]              #=> 404
-  e.response[:headers]             #=> { ... }
-  e.response[:body]                #=> "..."
-  e.response[:request][:url_path]  #=> "/wrong-url"
+  conn.get('/wrong-url')           # => Assume this raises a 404 response
+  conn.get('/protected-url')       # => Assume this raises a 401 response
+rescue Faraday::UnauthorizedError => e
+  e.response[:status]              # => 401
+  e.response[:headers]             # => { ... }
+  e.response[:body]                # => "..."
+  e.response[:request][:url_path]  # => "/protected-url"
 end
 ```
+
+In this example, a `Faraday::UnauthorizedError` exception is raised for the `/protected-url` request, while the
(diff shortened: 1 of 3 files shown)


iMacTia (approved):
This looks great, thank you for working on this!

I just want to be sure though you understand the behaviour will be slightly different from what you highlighted in the "After" section of the PR description:

```ruby
# In the case of a 404 status, this won't raise an error anymore,
# which is what we want 👍
existing_object = client.get('/foo/1').body

# However the "failed" response will still be processed as a successful one,
# meaning `existing_object` will always evaluate to `true` here 👇
# (e.g. the body of the response could still be an empty string)
# 
# You'd need to use a more explicitly check here, e.g. on the response status
# or the actual presence of the response body
if existing_object
  client.patch('/foo', data)
else
  client.post('/foo', data)
end
```

This is perfectly fine for me and it does make sense, but appreciate it might cause a bit of surprise

clemens:
Thanks for the positive feedback!

What you're saying is true: The code in my "After" example doesn't exactly work yet. I do think it makes more sense to adapt the example rather than trying to build in some "special" behavior into the middleware – would you agree? Then I'd just change the PR description accordingly.

iMacTia:
Yes, absolutely 👍 
No need to amend the PR description, it looks like we're on the same page anyway and the changes to the documentation don't present the same example.

I just wanted to make sure there was no misunderstanding 😄
Merging 🚀 !

iMacTia:
This is now merged into `main`, do you think you might test the changes in your project from there @clemens?
Nothing against cutting a new release, just trying to play it safe with some testing if possible 🙂

clemens:
> This is now merged into `main`, do you think you might test the changes in your project from there @clemens? Nothing against cutting a new release, just trying to play it safe with some testing if possible 🙂

I've tested it today (with multiple 404 cases in a larger script) and everything worked as expected: I had other 4xx and 5xx errors being raised, but 404 just let the script proceed => great! :)

PR #1600, opened 2024-11-20
Fix relative path on middleware doc

nogtk:
## Description
I found the link to each included middleware from middleware/index.md was broken, it's defined as a relative path so we don't need `middleware` directory.

https://github.com/lostisland/faraday/blob/93ef9e0ea905675358e2ae3edadebe1e13df95ef/docs/middleware/index.md?plain=1#L162-L168

## Todos
- [ ] Tests
- [ ] Documentation

This PR contains only the change of document, I think we can skip this todos but please let me know if I should take action for them.

## Additional Notes
nothing special

### DIFF
--- a/docs/middleware/index.md
+++ b/docs/middleware/index.md
@@ -160,12 +160,12 @@ After app initialization, all instances of the middleware will have the newly co
 ### Available Middleware
 
 The following pages provide detailed configuration for the middleware that ships with Faraday:
-* [Authentication](middleware/included/authentication.md)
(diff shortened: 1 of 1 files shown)


iMacTia:
I appreciate you looking into this, but as it was already pointed out, we need to write hyperlinks like this in order for the docs website to be rendered correctly.

I'd love for a solution that makes both work, but unfortunately I was unable to find one 🥲

Maybe someone more knowledgeable will have better luck!

Closing for now, but hopefully we'll get this to work one day 🤞

PR #1606, opened 2024-12-03
Formatting the log using parameter progname for the logger

### LINKED ISSUE #1605: Formatting the log using parameter progname for the logger
## Basic Info

* Faraday Version: 2.12.1
* Ruby Version: ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]

## Issue description
Now the logger uses the `progname` parameter as the message type. For example, here:
```
public_send(log_level, 'response') { "Status #{env.status}" }
```
https://docs.ruby-lang.org/en/master/Logger.html#method-i-info

And now you can't use this parameter from the outside. For example, like this:
```
conn = Faraday.new(url: 'http:/hello') do |faraday|
  faraday.response :logger, MyLogger.new($stdout, progname: 'my_best_program')
end
```
The log will still remain the same:
```
I, [2024-12-03T18:34:44.550323 #2735645]  INFO -- : request: GET http:/hello
I, [2024-12-03T18:34:44.550353 #2735645]  INFO -- : request: User-Agent: "Faraday v2.12.1"
I, [2024-12-03T18:34:44.550427 #2735645]  INFO -- : response: Status 200
I, [2024-12-03T18:34:44.550435 #2735645]  INFO -- : response: Content-Type: "text/html"
```

It seems to me it is better to transfer the message type from `progname` to the message body. And then this parameter will be used in the log:
```
I, [2024-12-03T18:35:01.582513 #2736044]  INFO -- my_best_program: request: GET http:/hello
I, [2024-12-03T18:35:01.582543 #2736044]  INFO -- my_best_program: request: User-Agent: "Faraday v2.12.1"
I, [2024-12-03T18:35:01.582619 #2736044]  INFO -- my_best_program: response: Status 200
I, [2024-12-03T18:35:01.582628 #2736044]  INFO -- my_best_program: response: Content-Type: "text/html"
```

This is necessary when several logs are recorded in one place (for example, in one file), and then you can easily separate some logger entries from others by parameter `progname`.

kodram:
Fixes #1605

### DIFF
--- a/spec/faraday/response/logger_spec.rb
+++ b/spec/faraday/response/logger_spec.rb
@@ -55,6 +55,26 @@ RSpec.describe Faraday::Response::Logger do
     end
   end
 
+  context 'when logger with program name' do
+    let(:logger) { Logger.new(string_io, progname: 'my_best_program') }
+
+    it 'logs with program name' do
+      conn.get '/hello'
+
+      expect(string_io.string).to match('-- my_best_program: request:')
+      expect(string_io.string).to match('-- my_best_program: response:')
+    end
+  end
+
+  context 'when logger without program name' do
+    it 'logs without program name' do
+      conn.get '/hello'
+
+      expect(string_io.string).to match('-- : request:')
+      expect(string_io.string).to match('-- : response:')
+    end
+  end
+
   context 'with default formatter' do
     let(:formatter) { instance_double(Faraday::Logging::Formatter, request: true, response: true, filter: []) }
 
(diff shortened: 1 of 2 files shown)


spec/faraday/response/logger_spec.rb
```diff
@@ -55,6 +55,26 @@
...
+      expect(string_io.string).to match('-- my_best_program: response:')
+    end
+  end
+
+  context 'when logger without program name' do
+    it 'logs without program name' do
+      conn.get '/hello'
+
+      expect(string_io.string).to match('-- : request:')
+      expect(string_io.string).to match('-- : response:')
```
iMacTia, line 74: Not a huge fan of the `:` at the beginning, but I guess it can't be helped. It doesn't seem like we're printing it so I assume this is default `Logger` behaviour

iMacTia (approved):
Thank you for the change!
I agree we were abusing the `progname` parameter until now, the change is very much welcome 🙌

PR #1618, opened 2025-04-16
Logger middleware default options

yenshirak:
## Description
Currently, it isn't possible to set the default options with of the Logger middleware with `Faraday::Response::Logger.default_options=` because `DEFAULT_OPTIONS` isn't defined and `options` isn't passed to `super`.

### DIFF
--- a/lib/faraday/response/logger.rb
+++ b/lib/faraday/response/logger.rb
@@ -10,11 +10,13 @@ module Faraday
     # lifecycle to a given Logger object. By default, this logs to STDOUT. See
     # Faraday::Logging::Formatter to see specifically what is logged.
     class Logger < Middleware
+      DEFAULT_OPTIONS = { formatter: Logging::Formatter }.merge(Logging::Formatter::DEFAULT_OPTIONS).freeze
+
       def initialize(app, logger = nil, options = {})
-        super(app)
+        super(app, options)
         logger ||= ::Logger.new($stdout)
-        formatter_class = options.delete(:formatter) || Logging::Formatter
-        @formatter = formatter_class.new(logger: logger, options: options)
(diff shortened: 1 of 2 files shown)


lib/faraday/response/logger.rb
```diff
@@ -10,11 +10,13 @@ class Response
     # lifecycle to a given Logger object. By default, this logs to STDOUT. See
     # Faraday::Logging::Formatter to see specifically what is logged.
     class Logger < Middleware
+      DEFAULT_OPTIONS = { formatter: Logging::Formatter }.merge(Logging::Formatter::DEFAULT_OPTIONS).freeze
```
iMacTia, line 13: `Logging::Formatter::DEFAULT_OPTIONS` are already used in the formatter class ([here](https://github.com/lostisland/faraday/blob/main/lib/faraday/logging/formatter.rb#L16)), so we shouldn't need to merge them here as well.

iMacTia (changes requested):
Hi @yenshirak and thank you for your contribution!
Apologies for the slow review (mostly because of the Easter Holidays 😅).

This change looks great. I only left a small comment with a change request.

lib/faraday/response/logger.rb
yenshirak: @iMacTia The problem is that the default options are validated and only the keys that are present in `DEFAULT_OPTIONS` can be used: https://github.com/lostisland/faraday/blob/main/lib/faraday/middleware.rb#L45

iMacTia: Ah good point! That is very much messed up 😄
I wish we had gone with some namespaced option like `formatter_options`, but changing that now would mean making this a breaking change.

iMacTia (approved):

PR #1621, opened 2025-04-28
Add Connection#middlewares to get registered middlewares

yykamei:
## Description

This method returns the list of registered middlewares, which can aid in debugging and verifying the correct order of middleware configuration.

While the internal builder uses the term `handlers`, Faraday's public documentation consistently refers to `middleware` or `middlewares`. Naming this public method `middlewares` aligns better with the terminology familiar to Faraday users.

## Todos
- [x] Tests
- ~[ ] Documentation~
  - I think this feature is not need to be written in the docs.


.rubocop_todo.yml
yykamei, line 1: Adding a new method to `Connection` triggered the `Metrics/ClassLength` RuboCop violation. I've updated `.rubocop_todo.yml` to temporarily ignore it.

iMacTia:
This addition is really small so I don't feel strongly against it, but I also can't entirely see what value it actually brings and would appreciate some further argumentation in its favour before considering to add it.

My main concerns are:
1. The same can easily be achieved with `conn.builder.handlers`, so the new method is just a slightly shorter alias;
2. The returned array could be misleading: although the method is called `middlewares` and running this on the console shows a list of what look like `Faraday::Middleware` classes, that's actually a trick because they're instances of `Faraday::RackBuilder::Handler` and appear like so because they override `#to_s`
3. The main argument in favour of this new method is that it might be more user-friendly to use, but without additional documentation people won't know about its existence. Having to dig through the source code to find it doesn't make it much more useful than the existing `conn.builder.handlers`

yykamei:
Thank you for your thoughtful feedback and for taking the time to discuss this pull request.

After carefully considering your comments, particularly your first point, I've decided to close this pull request.

You raised very valid points, and I've come to agree with your assessment. Ultimately, as you pointed out, this method is essentially just a shorter alias for `conn.builder.handlers`. For me, using `conn.builder.handlers` resolves my problems so far, and I don't have strong opinion to encourage other developers to use it after receiving your feedback.

Again, thank you for your time to read this pull request 😄

PR #1623, opened 2025-05-11
Only load what is required from `cgi`

Earlopain:
## Description

In Ruby 3.5 most of the `cgi` gem will be removed. Only the various escape/unescape methods will be retained by default.

On older versions, `require "cgi/util"` is needed because the unescape* methods don't work otherwise.

## Additional Notes

https://bugs.ruby-lang.org/issues/21258

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-require 'cgi'
+require 'cgi/escape'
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):

iMacTia (approved):
This is passing on all `3.0+` Ruby versions, including `3.5.0.dev`, thank you 🙇 !

PR #1624, opened 2025-05-14
Convert strings to UTF-8

c960657:
## Description
If the HTTP response body contains 8bit characters, and the adapter does not set the encoding of `body` to UTF-8 (like here: https://github.com/lostisland/faraday-net_http/pull/13), the logger middleware fails with an error such as this:

> log writing failed. "\xC3" from ASCII-8BIT to UTF-8

With this PR, the body is converted to UTF-8, replacing any 8-bit characters with the [replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character) (U+FFFD, �).

### DIFF
--- a/lib/faraday/logging/formatter.rb
+++ b/lib/faraday/logging/formatter.rb
@@ -63,7 +63,7 @@ module Faraday
 
       def dump_body(body)
         if body.respond_to?(:to_str)
-          body.to_str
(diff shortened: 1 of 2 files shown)


iMacTia (approved):
Thank you for fixing this @c960657 and apologies if it took me so long to get to it 🙇 
This fix only affects the logger and is not touching the actual response body, so it should be perfectly backwards-compatible 🙌

PR #1625, opened 2025-06-15
Add migrating from rest-client docs section.

simi:
`rest-client` is going to be archived and it is recommend to migrate to other library. Faraday seems the best and most straight-forward option. I have added simple guide comparing basic actions using both libraries for simpler upgrade experience. Happy to extend this later with more useful cases.


olleolleolle (approved):
Thank you for this helpful conversion article. ❤️

PR #1635, opened 2025-08-03
Fixes documentation

dharamgollapudi:
Although custom adapter is also a middleware, in this context, it is better to call it out as custom adapter than middleware. Fixes it by updating index.md

## Description
A few sentences describing the overall goals of the pull request's commits.
Link to related issues if any. (As `Fixes #XXX`)

## Todos
- [ ] Tests
- [ ] Documentation

## Additional Notes
Optional section


iMacTia (approved):
Good catch, most likely a copy-paste error from the middleware page. Thank you!

PR #1638, opened 2025-08-14
Use newer Unprocessable Content naming for 422

tylerhunt:
## Description

Rename `Faraday::UnprocessableEntityError` to `Faraday::UnprocessableContentError` but provide compatibility with the older error constant through the use of an included module.

This is based on the status code name specified by [IETF RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110). Rack 3.1 was updated to account for this change (rack/rack#2137), and Rails has added support for it (rails/rails#53383), as well.

### DIFF
--- a/lib/faraday/error.rb
+++ b/lib/faraday/error.rb
@@ -154,8 +154,13 @@ module Faraday
   class ConflictError < ClientError
   end
 
+  # Used to provide compatibility with legacy error name.
+  module UnprocessableEntityError
+  end
+
   # Raised by Faraday::Response::RaiseError in case of a 422 response.
-  class UnprocessableEntityError < ClientError
+  class UnprocessableContentError < ClientError
+    include UnprocessableEntityError
   end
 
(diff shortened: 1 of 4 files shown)


lib/faraday/error.rb
```diff
@@ -154,8 +154,13 @@ class RequestTimeoutError < ClientError
   class ConflictError < ClientError
   end
 
+  # Used to provide compatibility with legacy error name.
+  module UnprocessableEntityError
```
iMacTia, line 158: I'd rather use a constant assignment to keep this as a class

iMacTia (changes requested):
Thank you @tylerhunt and apologies for taking so long to get back to you 🙇 

I appreciate the change, but I'd suggest using a straight constant assignment to provide backwards compatibility as opposed to turning `UmprocessableEntity` into a `module` and including it in the new class.

Appreciate this PR has been open for a while, so I'll proceed with this change myself.

PR #1657, opened 2025-12-09
Explicit top-level namespace reference

c960657:
## Description
Use top-level constant operator to reference root namespace.

Fixes the issue raised by @stanley90 in https://github.com/lostisland/faraday/pull/1624#issuecomment-3627584422.  I cannot reproduce it, but I assume it is an issue triggered by Zeitwerk or autoload.


olleolleolle (approved):
Thanks, @c960657 !

PR #1660, opened 2026-02-07
Modernize RuboCop configuration and fix offenses

larouxn:
## Description

Modernizes and generally simplifies our RuboCop usage via the following:
- migrate from `require` to `plugins`
- remove explicit new cops in favour of `NewCops: enable`
- remove version boundaries from `rubocop-packaging` and `rubocop-performance` gems
- add and enable `rubocop-rake` gem
- auto-correct (safe only) a bunch of offenses
- manually correct "useless private scoping of constant" offences
- regen `.rubocop_todo.yml` to add remaining few complex offences

Screenshot From 2026-02-07 18-29-52
As a future improvement we could consider adding the `rubocop-rspec` gem.

image

## Todos

None if CI looks good. Passed locally.

## Additional Notes

Local testing and RuboCop, for reference.

Screenshot From 2026-02-07 18-38-54
Screenshot From 2026-02-07 18-39-04


olleolleolle:
This is not me blocking a merge, but this chang means we are taking on following the changes in RuboCop henceforth.

lib/faraday/rack_builder.rb
```diff
@@ -15,6 +15,11 @@ class RackBuilder
     # Used to detect missing arguments
     NO_ARGUMENT = Object.new
 
+    LOCK_ERR = "can't modify middleware stack after making a request"
```
olleolleolle: Minor:

Perhaps these constants were _intended_ to be private constants, and their placement in the private section was just a... foiled attempt at that.

Perhaps making them private_constant would be useless, and that since they've been exposed as visible API for a long while, it's de facto our current public API.

So, with those thoughts, I recommend this move.

olleolleolle (approved):
I guess we _can_ take this step.

PR #1675, opened 2026-05-19
feat(test): add Stubs#clear to remove all stubs

### LINKED ISSUE #1570: Would you be open to me contributing a `#clear` method in `Faraday::Adapter::Test::Stubs`?
Hey folks! I've recently stumbled upon a challenge similar to what's documented in https://github.com/lostisland/faraday/issues/1041. That is, my application uses singleton classes with memoized methods to "cache" a `Faraday` instance. This works great in production but my tests will suffer from being unable to stub out different behaviors of the same URL/action in subsequent test examples.

I'm comfortable enough with the test adapter and stub code, and believe I could implement a method to clear the stubs reasonably well. Before I go off and spend time on that, could the maintainers of the project weigh in on whether such a thing would be in-line with the maintainers' goals and design philosophy?

Thanks in advance!

SAY-5:
## Description
Adds a `clear` method to `Faraday::Adapter::Test::Stubs` that removes all stubs, including consumed ones, so the test adapter can be reset between examples without the `Faraday.default_connection = nil` hack. Docs updated to suggest it. Closes #1570, as @iMacTia agreed in the issue ("I'd be open to introducing a `clear` method on `Faraday::Adapter::Test::Stubs`").

## Todos
- [x] Tests
- [x] Documentation

## Additional Notes
None.

### DIFF
--- a/docs/adapters/test-adapter.md
+++ b/docs/adapters/test-adapter.md
@@ -100,9 +100,16 @@ verify the order or count of any stub.
 stubs.verify_stubbed_calls
 ```
 
-After the test case is completed (possibly in an `after` hook), you should clear
-the default connection to prevent it from being cached between different tests.
-This allows for each test to have its own set of stubs
+After the test case is completed (possibly in an `after` hook), you should reset
+the stubs so each test starts with its own set. When you hold a reference to the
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Thank you, @SAY-5, this clarifies!

PR #1687, opened 2026-08-12
Support JSON 3

byroot:
In the upcoming `json 3.x`, `JSON.parse` and `JSON.load` take keyword arguments rather than a positional option Hash.

So the options must be splatted.

Thankfully, when splatting on a method that doesn't take keyword arguments, Ruby pass the arguments as positional:

```ruby
def load(source, options = {})
  p options
end

kwargs = {a: 1, b: 2}
load('null', **kwargs) # => {a: 1, b: 2}
```

So the change is compatible with both `json 2.x` and `json 3.x`

### DIFF
--- a/lib/faraday/response/json.rb
+++ b/lib/faraday/response/json.rb
@@ -33,7 +33,7 @@ module Faraday
 
         decoder, method_name = @decoder_options
 
-        decoder.public_send(method_name, body, @parser_options || {})
+        decoder.public_send(method_name, body, **(@parser_options || {}))
       end
 
       def parse_response?(env)


olleolleolle (approved):
Thanks, sounds ace.

iMacTia:
This is great, thank you for contributing!
There's only one thing I'm not entirely sure about: the `decoder` is configurable and only defaults to `JSON`, so people could swap it with any other library (the most popular likely being `Oj`).

I don't think we have tests that cover this, as it would require the Oj library to be added (which would be fine actually, in the Gemfile), so I was thinking we should at least make sure that one accepts the same change in call structure before shipping a release and risking to break every app using a custom decoder on a minor release.

byroot:
If you wish to make changes to this, I suggest instead of taking a method name and arguments, just take either a Proc, or an object that respond to the standard `.load` method.

sue445:
Hi there.

`json` gem v3.0.0 was recently released.
https://github.com/ruby/json/releases/tag/v3.0.0

However, because this patch has not been released yet, the following error occurs in `Faraday::Response::Json`.

```
 Faraday::ParsingError:
   wrong number of arguments (given 2, expected 1)
 # ./vendor/bundle/ruby/3.0.0/gems/json-3.0.2/lib/json/common.rb:296:in `parse'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/response/json.rb:36:in `public_send'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/response/json.rb:36:in `parse'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/response/json.rb:26:in `process_response'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/response/json.rb:19:in `on_complete'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/middleware.rb:57:in `block in call'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/response.rb:46:in `on_complete'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/middleware.rb:56:in `call'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/middleware.rb:56:in `call'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/rack_builder.rb:158:in `build_response'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/connection.rb:452:in `run_request'
 # ./vendor/bundle/ruby/3.0.0/gems/faraday-2.14.3/lib/faraday/connection.rb:200:in `get'
 # ./spec/faraday/mashify/integration_spec.rb:16:in `block (2 levels) in <top (required)>'
 # ./vendor/bundle/ruby/3.0.0/gems/webmock-3.26.4/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'
 # ------------------
 # --- Caused by: ---
 # ArgumentError:
 #   wrong number of arguments (given 2, expected 1)
 #   ./vendor/bundle/ruby/3.0.0/gems/json-3.0.2/lib/json/common.rb:296:in `parse'
```

Do you have any plans to release a new version that includes this patch?

dblock:
This fix isn't in a release yet (latest is v2.14.3). Worked around it in the meantime with a custom middleware: dblock/strava-ruby-client#115. Looking forward to a new Faraday release with this included!

(Comment posted by an AI on behalf of @dblock.)

crevete:
@olleolleolle When the next release will be cut please?

sue445:
@crevete It looks like v2.14.4 has already been released.
https://github.com/lostisland/faraday/releases/tag/v2.14.4

@olleolleolle Thanks for the release!

dblock:
Verified in dblock/strava-ruby-client#115: Faraday v2.14.4 fixes the `json` 3 compatibility issue for us. We now require Faraday >= 2.14.4, use the built-in `:json` middleware again, and our test suite passes against both JSON 2 and JSON 3.

Thanks for the fix and release!

Posted by an AI on behalf of @dblock.

PR #1689, opened 2026-08-27
Preserve caller-owned JSON parser options

OskarEichler:
## Summary

Duplicate parser options before extracting the decoder.

## Reproduction and verification

Construct JSON response middleware twice from the same parser_options hash containing :decoder: baseline deletes the decoder and the second instance falls back to JSON. A frozen options hash raises FrozenError. Both work after the fix.

- Individual patch: 9 focused Faraday checks passed (json); all 639 existing RSpec examples pass; changed-file RuboCop and Ruby syntax pass.
- Combined installed-release-based branch: 639 existing examples, 1,124 focused checks, full RuboCop (75 files), YARD checks and gem packaging pass.
- External faraday-net_http adapter suite: 386 existing examples, zero failures with the combined fixes.
- Ruby 4.0.6 through rbenv. No production or live external HTTP operations.

## Limitations

No test/spec files were added or changed because the originating repository explicitly forbids this. Reproductions ran outside the repository. This differs from Faraday's requested regression-test contribution convention and is disclosed for maintainer review. Other Ruby versions/platforms were not run locally. This PR includes only the focused source change; the other fixes and the consumer's separately verified merged JSON decoder patch #1687 are not added here.

## Breaking-change notes

No API changes. Initialization stops mutating caller options and snapshots their top-level values.

### DIFF
--- a/lib/faraday/response/json.rb
+++ b/lib/faraday/response/json.rb
@@ -8,7 +8,7 @@ module Faraday
     class Json < Middleware
       def initialize(app = nil, parser_options: nil, content_type: /\bjson$/, preserve_raw: false)
         super(app)
-        @parser_options = parser_options
+        @parser_options = parser_options&.dup
         @content_types = Array(content_type)
         @preserve_raw = preserve_raw
 


olleolleolle (approved):
This change looks good.

The note about the repo not accepting tests read a little strong.

Issue #1168, opened 2020-07-17
Feature Request: explicit method defs for HTTP verbs


dduugg:
It seems the top-level HTTP verb methods (e.g. `Faraday.get`) are handled via `missing_method` [proxying](https://github.com/lostisland/faraday/blob/5acab3623994e6dbcc77f203fe0b5e1f29684874/lib/faraday.rb#L112-L120) to the `default_connection` ivar. In addition to being a tad inefficient (admittedly a minor concern compared to typical network request latency), it causes issues with various tools:

- The methods don't autocomplete in `pry`, `solargraph`, etc.
- The documentation isn't picked up by https://www.rubydoc.info/gems/faraday/Faraday
- Usage fails sorbet typechecking (though I attempted to fix this in https://github.com/sorbet/sorbet-typed/pull/268/commits/83132c1b19c50ad7516ecc806243afbd3445b98d, but it's susceptible to drift)

Would you be amenable to replacing this implementation with explicit method definitions?

iMacTia:
Hi @dduugg, rather than explicitly defining the methods, I'd explore a documentation-friendly solution.
We use meta-programming definition in the connection as well: https://github.com/lostisland/faraday/blob/master/lib/faraday/connection.rb#L195

Does that work nicely with the points you mentioned above?
If so, we could replicate a similar documentation snippet for the `Faraday` module as well.

dduugg:
Hi @iMacTia, thanks for the reply. I don't think the problem is solvable with documentation. Could I trouble you to elaborate on the resistance to explicit definitions? I think that would better frame the discussion. For instance, a documentation approach is likely just as verbose (counting various `@!method` & `@!scope` directives, in addition to the `method_missing` code) than the seven one-line explicit forwarding defs I'm suggesting on `Faraday`. But I don't know if that's the rationale here (code is read much more than it's written, so I think optimizing for the code writer is in error anyway).

iMacTia:
There are mainly 2 points that put me on the fence with this change.
* I personally don't think using `delegates` is an advantage only for the code writer. I've been using Ruby for years now and like others I came to appreciate the succinct syntax of this language. When I read some code and I find `delegates` I immediately know what that means and it can potentially condense tens of lines of code into just a couple of them.
[...]

PR #1708, opened 2026-09-02
Define HTTP verb methods on Faraday so tools can see them

mokevnin:
This implements the approach discussed in #1168, following what @iMacTia said there:

> But if we can fix `pry`'s discovery by adding dynamic definitions or delegations to the main `Faraday` module, then that's something I'd be OK with.

### What

`Faraday.get`, `.post` and friends only exist at runtime today — they reach `default_connection` via the private `method_missing` proxy. So they're absent from `Faraday.methods` and `ls Faraday` in pry, they don't turn up in generated docs or editor autocomplete, and static analysis tools flag them as undefined.

Since the verb set is already fixed in `METHODS_WITH_QUERY` and `METHODS_WITH_BODY`, the methods are now generated from those constants using the same `class_eval` loop `Faraday::Connection` already uses for its own verbs — no hand-written list, and the per-shape signature (`url, params, headers` / `url, body, headers`, plus the block) is preserved, so `Faraday.method(:get).parameters` is meaningful.

```ruby
Faraday.methods(false).grep(/get|post/)  # => [:get, :post]
Faraday.method(:get).parameters          # => [[:opt, :url], [:opt, :params], [:opt, :headers], [:block, :block]]
```

### No behavior change

The generated methods call the exact same `default_connection` methods the proxy called. `method_missing` and `respond_to_missing?` are untouched and still handle everything else they forwarded, `Faraday.options` included.

### Tested

`bundle exec rake` (646 examples, 0 failures), `bundle exec rubocop` (clean) and `bundle exec yard-junk --path lib` (clean) on Ruby 4.0. The added spec asserts each verb is really defined on the singleton class (not resolved through `method_missing`) and that url, second argument, headers and the block all arrive at the connection.

Happy to add YARD `@!method` directives on top if you'd like the rubydoc side covered too — left out here to keep the diff small.

### DIFF
--- a/lib/faraday.rb
+++ b/lib/faraday.rb
@@ -137,6 +137,38 @@ module Faraday
       @default_connection_options = ConnectionOptions.from(options)
     end
 
+    # HTTP verb methods, forwarded to {.default_connection}.
+    #
+    # They take the same arguments as their {Faraday::Connection} counterparts,
+    # see {Faraday::Connection#get} and {Faraday::Connection#post}.
+    # Defining them here rather than relying on {method_missing} keeps them
+    # visible to `Faraday.methods`, `pry`'s `ls`, documentation tools and
+    # editor autocomplete.
+
+    # @!visibility private
+    METHODS_WITH_QUERY.each do |method|
+      class_eval <<-RUBY, __FILE__, __LINE__ + 1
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
I find the pry `ls` argument compelling.

PR #1709, opened 2026-09-04
Document Options struct attributes for YARD (Fixes #1490)

### LINKED ISSUE #1490: Fix YARD doc for Options
## Issue description
In #1489, we changed the definition of `Options` (`Struct`s) from `Env < Options.new` to `Env = Options.new { }`.
This makes our `Options` structs compatible with Sorbet, but it breaks their YARD auto-generate docs.
See the PR thread for more details.

tachyurgy:
Fixes #1490.

`Faraday::Options` subclasses are built dynamically with `Options.new(...)`, so YARD has no
static `attr_accessor` calls to pick up and the struct members end up undocumented. The
`@!parse` blocks in these three files documented the class but not its attributes.

This adds `@!attribute` tags inside the existing `@!parse` blocks for `ConnectionOptions`,
`RequestOptions` and `ProxyOptions`, with a `@return` type and a one-line description per
member. The tag names and order match the `Options.new(...)` member lists exactly.

Documentation only. No runtime code is touched.

**Verified locally (Ruby 3.3.7):**

| | upstream/main | this branch |
|---|---|---|
| YARD `Attributes` | 54 | **68** |
| Overall documented | 65.03% | **66.47%** |

- `bundle exec rspec` — 646 examples, 0 failures
- `bundle exec rubocop` — 75 files inspected, no offenses

Happy to adjust any of the type signatures or wording; a few were inferred from usage rather
than from an explicit declaration, so maintainer correction is welcome.

### DIFF
--- a/lib/faraday/options/connection_options.rb
+++ b/lib/faraday/options/connection_options.rb
@@ -4,6 +4,33 @@ module Faraday
   # @!parse
   #   # ConnectionOptions contains the configurable properties for a Faraday
   #   # connection object.
+  #   #
+  #   # @!attribute request
+  #   #   @return [RequestOptions] default request options for the connection
+  #   #
+  #   # @!attribute proxy
+  #   #   @return [ProxyOptions] proxy options for the connection
+  #   #
+  #   # @!attribute ssl
+  #   #   @return [SSLOptions] SSL options for the connection
+  #   #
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
This looks good, thanks!

PR #1717, opened 2026-09-19
Fix FrozenError when a frozen URI is passed as the base URL

### LINKED ISSUE #1349: Faraday tries to modify frozen URI (FrozenError)

hikmetba-bit:
`Utils.URI` returns a `URI` object as-is (no `dup`) whenever it already responds to `:host`:

```ruby
def URI(url)
  if url.respond_to?(:host)
    url
  elsif url.respond_to?(:to_str)
    default_uri_parser.call(url)
  else
    raise ArgumentError, 'bad argument (expected URI object or URI string)'
  end
end
```

`url_prefix=` then mutates that same object (`self.path_prefix = uri.path` → `uri.path=`, plus `uri.query=` and `uri.user=`/`uri.password=` later in the method). If the caller passed a frozen `URI`, this raises `FrozenError`. Even for a non-frozen `URI`, it silently mutates an object the caller still holds a reference to (see the `alexcwatt` comment on the issue — this bit them independently of freezing).

```ruby
endpoint = URI('http://example.org/api/v1').freeze
Faraday.new(endpoint)
# FrozenError: can't modify frozen URI::HTTP: #<URI::HTTP http://example.org/api/v1>
```

## Fix

Stringify `url` before handing it to `Utils.URI`, so the `elsif url.respond_to?(:to_str)` branch — which always builds a fresh `URI` via the configured parser — is taken unconditionally, regardless of whether the caller passed a `String` or a `URI`. This is the fix @iMacTia suggested in the issue thread ("I'd go with one of the simple solutions (maybe the `#to_s` one, with a complimentary comment explaining why we're doing that)"), scoped narrowly to avoid touching the separately-discussed `path_prefix=` inconsistencies that are being deferred to a 3.0 rewrite.

A prior attempt at this fix (#1685) was closed with no review comments; the bug is unchanged on `main`.

## Testing

- Added 5 regression tests to `spec/faraday/connection_spec.rb`: constructing a connection from a frozen `URI` doesn't raise, resolves the expected host/path, and leaves the original object frozen; constructing from a non-frozen `URI` doesn't mutate the caller's object.
- Mutation-tested: reverted just the one-line source fix and reran — 4 of the 5 new tests failed with the exact reported `FrozenError`, then restored the fix and the full suite passed.
- Full `connection_spec.rb` suite: 143 examples, only one pre-existing failure (`proxy support does not accept uppercase env`), confirmed present and identical on unmodified `main` too — unrelated to this change (a Windows-environment proxy-detection quirk, not something this PR touches).
- `rubocop` on both changed files: no offenses.

Fixes #1349

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/faraday/connection.rb
```diff
@@ -362,7 +362,12 @@ def proxy=(new_value)
     #
     #   conn.get("nigiri?page=2") # accesses https://httpbingo.org/api/nigiri
     def url_prefix=(url, encoder = nil)
-      uri = @url_prefix = Utils.URI(url)
+      # Utils.URI returns a URI object as-is without duping it, so mutating
+      # it below (path_prefix=, query=, user=/password=) would mutate the
+      # caller's own URI -- and raise FrozenError if they passed a frozen
+      # one. Stringifying first guarantees Utils.URI always builds a fresh,
+      # unfrozen URI instead.
```
olleolleolle, line 369: Drop the comment - the tests prove the story. 😀

